PR #28, opened 2018-07-25
Unittests

ManuelAlvarezC:
Unittests covering most of the basic transformers behavior.

Some are skipped due to bugs whose tickets are already being worked on.


ManuelAlvarezC:
Resolve #3

PR #40, opened 2018-08-28
Issue 39: Fixing KeyError: nan bug in CatTransformer

### LINKED ISSUE #39: KeyError: nan problem with CatTransformer get_val function
https://github.com/HDI-Project/RDT/blob/7bc38da2c9cd737fdd2ff3207d3cd3482701bb6d/rdt/transformers/CatTransformer.py#L86

This dictionary will represent all nans as type <class 'float'>. However, sometimes the original value in the column will be of type <class 'numpy.float64'>. If this is the case, then when looking up the key of type <class 'numpy.float64'>, the dictionary created on line 86 will give KeyError: nan.

To recreate:
```python
s = pd.Series([np.nan, 1, 5])
s2 = s.fillna(-1)
d = s2.groupby(s2).count().rename({-1: np.nan}).to_dict()
d[s[0]]
```

amontanez24:
Fixes #39. Sometimes nan values are loaded as type <class 'numpy.float64'> and other times as type <class 'float'>. The CatTransformer should handle all nan types the same since they are all mapped to the same category in the get_probability_map method. This PR fixes this by simply mapping all nan types to -1 and using that as the key in the probability map instead.

### DIFF
--- a/rdt/transformers/CatTransformer.py
+++ b/rdt/transformers/CatTransformer.py
@@ -20,8 +20,11 @@ class CatTransformer(BaseTransformer):
         """ Returns a tuple (transformed_table, new_table_meta) """
         out = pd.DataFrame()
         col_name = col_meta['name']
-        self.get_probability_map(col)
-        out[col_name] = col.apply(self.get_val)
+        # Make sure all nans are handled the same by replacing
+        # with -1
+        column = col.fillna(-1)
+        self.get_probability_map(column)
+        out[col_name] = column.apply(self.get_val)
         # Handle missing
 
         if missing:
@@ -77,13 +80,8 @@ class CatTransformer(BaseTransformer):
 
     def get_probability_map(self, col):
         """ Maps each unique value to probability of seeing it """
-        # According to pandas docs, nan values are excluded on groupby:
-        # http://pandas.pydata.org/pandas-docs/stable/missing_data.html#na-values-in-groupby
-        # So we are replacing them with -1 and replace them back with np.nan after groupby
-        # Further discussion: https://github.com/pandas-dev/pandas/issues/3729
 
-        column = col.fillna(-1)
-        self.probability_map = column.groupby(column).count().rename({-1: np.nan}).to_dict()
+        self.probability_map = col.groupby(col).count().to_dict()
         # next set probability ranges on interval [0,1]
         cur = 0
         num_vals = len(col)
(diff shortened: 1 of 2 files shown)


tests/rdt/transformers/test_CatTransformer.py
```diff
@@ -178,3 +179,20 @@ def test_get_probability_map(self):
...
+            "name": "breakfast",
+            "type": "categorical"
+        }
+        transformer = CatTransformer()
+        # Run
+        output = transformer.fit_transform(data, col_meta)
+
+        # Check
+        # the  nan value in the data should be in probability map
+        assert output is not None
```
ManuelAlvarezC: This is not checking that the nan value is in the probability map.

The check for that should be something like 

```python
assert -1 in transformer.category_map
```
But in the concrete case, you are using integers as categorical data, and if that's a valid use case, then you can't guaratee that you won't be having a category with value -1.

rdt/transformers/CatTransformer.py
```diff
@@ -20,8 +20,11 @@ def fit_transform(self, col, col_meta, missing=True):
         """ Returns a tuple (transformed_table, new_table_meta) """
         out = pd.DataFrame()
         col_name = col_meta['name']
-        self.get_probability_map(col)
-        out[col_name] = col.apply(self.get_val)
+        # Make sure all nans are handled the same by replacing
+        # with -1
+        column = col.fillna(-1)
```
ManuelAlvarezC: I have checked what caused the issue with the nan keys, and the main problem is that:

```python
>>> np.nan == np.nan
False
>>> float('nan') == float('nan')
False
```
But also, `-1` could be a valid categorical value, so I'll recommend fill the nan values with `None` wich also represents nullity, but have a more consistent behaviour:

```python
>>> None == None
True
```
I would also recommend make the `fillna` call inside the `get_probability_map` method.

rdt/transformers/CatTransformer.py
```diff
@@ -20,9 +20,8 @@ def fit_transform(self, col, col_meta, missing=True):
         """ Returns a tuple (transformed_table, new_table_meta) """
         out = pd.DataFrame()
         col_name = col_meta['name']
-        # Make sure all nans are handled the same by replacing
-        # with -1
-        column = col.fillna(-1)
+        # Make sure all nans are handled the same by replacing with None
+        column = col.replace({np.nan: None})
```
ManuelAlvarezC: This line shouldn't be here. With the replacement inside `get_probability_map` should be enough.

amontanez24: It does not work without it

Issue #61, opened 2018-10-05
Enforce python standards.


ManuelAlvarezC:
Currently, there a some minor detailis that broke the python standards such as:

- Redundant names, or using uppercase.
- Docstrings not following PEP257.
- Requirements out of `setup.py`
- No module contents declaration on  `rdt.__init__`

csala:
This is already covered in the current code base

PR #74, opened 2019-01-09
Issue 71: Add option to anonimize data

### LINKED ISSUE #71: Add option to anonimize data
Sometimes transfomers are used with sensitive data we don't want the `reverse_transform` to take the values from nor keep them after extracting its distribution. 

In order to do so, we should anonimize data before creating `CatTransformer.probability_map`, so its mapped to new values taken from `faker` but preserving its original distribution.

To set this option, there would be two flags on `rdt.transformers.CatTransformer.fit`:

- `anonimize`: bool 
- `category`: str 

If the flag `pii` is set, then the values should be mapped with values generated from `faker`, before generating the `probability_map`. The supported values of `category` should be one of the attributes of `faker` object, that list includes, but is not limited to:
- first_name
- last_name
- name
- ssn
- phone_number
- email

ManuelAlvarezC:
- Add two arguments on `rdt.transfomers.category.CatTransformer.__init__` to control the anonimization of data.
- Add methods `get_generator` and `anonimize_column` that implement the actual logic of anonimization.
- Changed the behavior, so, if `anonimize=True`, the `fit` method is disabled ( as the `probability_map` will be generated with the anonimized data, thus making it impossible to process the original data), and only `fit_transform` can be used.
- Fixed the module structure as #61 

Resolves #71 #75


setup.py
```diff
@@ -13,6 +13,7 @@
     'numpy==1.15.4',
     'pandas==0.23.4',
     'scipy==1.1.0',
+    'Faker==1.0.1',
```
csala: Please use open dependency specification: `>=`
Also do that for the other dependencies. There is no need to hard fix all the versions.

PR #98, opened 2019-09-20
Issue 96 remake sphinx documentation

### LINKED ISSUE #96: Remake Sphinx Documentation
Sphinx documentation should be reconfigured to use the README as the main source of content, like it has been done in other HDI-Project projects.

JDTheRipperPC:
Resolves #96 
Updated sphinx documentation style.


Makefile
```diff
@@ -172,7 +173,7 @@ bumpversion-major: ## Bump the version the next major skipping the release
 	bumpversion --no-tag major
 
 CURRENT_BRANCH := $(shell git rev-parse --abbrev-ref HEAD)
```
csala: Can we also add ` 2>/dev/null` to this line?

docs/Makefile
```diff
@@ -4,7 +4,7 @@
 # You can set these variables from the command line.
 SPHINXOPTS    =
 SPHINXBUILD   = python -msphinx
-SPHINXPROJ    = rdt
+SPHINXPROJ    = tgan
```
csala: `rdt` :-P

Makefile
JDTheRipperPC: We don't need to redirect to /dev/null, this statement will return the current branch that will be compared later if the current branch is the master branch. We don't need to add it.

PR #137, opened 2020-11-13
OneHotEncodingTransformer support for lists and lists of lists

fealho:
Resolves sdv-dev/CTGAN#87

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -236,9 +236,6 @@ class OneHotEncodingTransformer(BaseTransformer):
 
         Returns:
             pandas.Series or numpy.ndarray
-
-                if isinstance(data, list):
-            data = np.array(data)
         """
         if isinstance(data, list):
             data = np.array(data)
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -245,7 +245,7 @@ class OneHotEncodingTransformer(BaseTransformer):
 
         if len(data.shape) > 2:
             raise ValueError("Unexpected format.")
-        elif len(data.shape) == 2:
+        if len(data.shape) == 2:
             if data.shape[1] != 1:
                 raise ValueError("Unexpected format.")
 
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -223,15 +223,40 @@ class OneHotEncodingTransformer(BaseTransformer):
     dummy_na = None
     dummies = None
 
+    @staticmethod
+    def _prepare_data(data):
+        """Transform data to appropriate format.
+
+        If data is a valid list or a list of lists, transforms it into an np.array,
+        otherwise returns it.
+
+        Args:
+            data (pandas.Series, numpy.ndarray, list or list of lists):
+                Data to prepare.
+
+        Returns:
+            pandas.Series or numpy.ndarray
+        """
+        if isinstance(data, list):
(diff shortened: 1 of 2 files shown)


rdt/transformers/categorical.py
```diff
@@ -223,28 +223,54 @@ class OneHotEncodingTransformer(BaseTransformer):
...
+        Args:
+            data (pandas.Series, numpy.ndarray, list or list of lists):
+                Data to prepare.
+
+        Returns:
+            pandas.Series or numpy.ndarray
+        """
+        if isinstance(data, list):
+            data = np.array(data)
+            if len(data.shape) == 2:
```
csala, line 242: Maybe we should put this block one level higher, outside of the `isinstance(data, list)` block?

This way we would also support 2D numpy arrays as input.

rdt/transformers/categorical.py
```diff
@@ -223,28 +223,54 @@ class OneHotEncodingTransformer(BaseTransformer):
...
+                if data.shape[1] != 1:
+                    raise ValueError("Unexpected format.")
+                data = data[:, 0]
```
csala, line 245: Can we add a couple of blank lines to separate the blocks after unindenting?

```python
if len(data.shape) == 2:
    if data.shape[1] != 1:
        raise ValueError("Unexpected format.")

    data = data[:, 0]

elif len(data.shape) > 2:
    raise ValueError("Unexpected format.")

return data
```

csala (changes requested):
Looks good! Just a couple of comments.

rdt/transformers/categorical.py
fealho: Sorry, I'm a little confused here, you mean just delete the isinstance?

rdt/transformers/categorical.py
fealho: Yes, this does look much more clean.

rdt/transformers/categorical.py
csala: No, I mean removing the indentation of the `if len(data.shape)` lines one level, so they are always run independently on whether the input was a list or a numpy array.

This way, if the input was already a numpy array, its shape is validated too.

fealho: Oh, mean this?

```python3
if isinstance(data, list):
    data = np.array(data)

if isinstance(data, np.ndarray):
    if len(data.shape) == 2:
        if data.shape[1] != 1:
            raise ValueError("Unexpected format.")
    
        data = data[:, 0]

    elif len(data.shape) > 2:
        raise ValueError("Unexpected format.")
    
return data
```

Maybe at this point we might as well just support anything np.ndarray does? Like this:

```python3
data = np.array(data)
if len(data.shape) == 2:
    if data.shape[1] != 1:
        raise ValueError("Unexpected format.")

    data = data[:, 0]

elif len(data.shape) > 2:
    raise ValueError("Unexpected format.")

return data
```

rdt/transformers/categorical.py
```diff
@@ -236,15 +236,21 @@ def _prepare_data(data):
 
         Returns:
             pandas.Series or numpy.ndarray
+
+                if isinstance(data, list):
```
csala, line 240: I suspect these lines have made their way here by error :-)

fealho: Oof, my bad.

csala (approved):

PR #144, opened 2020-12-02
Makes Copulas an optional dependency

fealho:
See title.

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -1,15 +1,55 @@
 """Transformers for numerical data."""
-
 import copy
 
 import numpy as np
 import pandas as pd
 import scipy
-from copulas import EPSILON, univariate
 
 from rdt.transformers.base import BaseTransformer
 from rdt.transformers.null import NullTransformer
 
+def _get_distributions():
+    from copulas import univariate
+    return {
+        'univariate': univariate.Univariate,
+        'parametric': (
+            univariate.Univariate, {
+                'parametric': univariate.ParametricType.PARAMETRIC,
+            },
+        ),
+        'bounded': (
+            univariate.Univariate,
+            {
+                'bounded': univariate.BoundedType.BOUNDED,
+            },
+        ),
+        'semi_bounded': (
+            univariate.Univariate,
+            {
+                'bounded': univariate.BoundedType.SEMI_BOUNDED,
+            },
+        ),
+        'parametric_bounded': (
+            univariate.Univariate,
+            {
+                'parametric': univariate.ParametricType.PARAMETRIC,
+                'bounded': univariate.BoundedType.BOUNDED,
+            },
+        ),
+        'parametric_semi_bounded': (
+            univariate.Univariate,
+            {
+                'parametric': univariate.ParametricType.PARAMETRIC,
+                'bounded': univariate.BoundedType.SEMI_BOUNDED,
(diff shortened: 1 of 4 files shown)


rdt/transformers/numerical.py
```diff
@@ -241,7 +244,7 @@ def fit(self, data):
 
     def _copula_transform(self, data):
         cdf = self._univariate.cdf(data)
-        return scipy.stats.norm.ppf(cdf.clip(0 + EPSILON, 1 - EPSILON))
+        return scipy.stats.norm.ppf(cdf.clip(0 + 1e-5, 1 - 1e-5))
```
csala, line 247: Let's define EPSILON like it's done in Copulas: https://github.com/sdv-dev/Copulas/blob/3210fc141c741185b781686cb69e00a96972d960/copulas/__init__.py#L16

```python
EPSILON = np.finfo(np.float32).eps
```

We can put this at module level.

rdt/transformers/numerical.py
```diff
@@ -1,16 +1,58 @@
...
 import numpy as np
 import pandas as pd
 import scipy
-from copulas import EPSILON, univariate
 
 from rdt.transformers.base import BaseTransformer
 from rdt.transformers.null import NullTransformer
 
 
+def _get_distributions():
```
csala, line 12: Since this is specific to the Transformer class only, I'd put it as a `@staticmethod` inside it.

rdt/transformers/numerical.py
```diff
@@ -1,16 +1,58 @@
...
+    from copulas import univariate  # pylint: disable=import-outside-toplevel
```
csala, line 13: Let's wrap this in a `try/except` block and provide a user friendly message with instructions about what to do.

Similar to this: https://github.com/sdv-dev/SDV/blob/8f3c787dc87ff2b85fe1ac392e0decf18ce134ad/sdv/tabular/ctgan.py#L93-L102

In this case, let's recommend installing `rdt[copulas]`

setup.py
```diff
@@ -28,6 +27,7 @@
     'pytest-cov>=2.6.0',
     'jupyter>=1.0.0,<2',
     'rundoc>=0.4.3,<0.5',
+    'copulas>=0.3.3,<0.4',
```
csala, line 30: Let's put this as a separate `copulas` extra, similar to the `ctgan` extra on `sdv`: https://github.com/sdv-dev/SDV/blob/8f3c787dc87ff2b85fe1ac392e0decf18ce134ad/setup.py#L26-L28

It should be also added to all the other `extras_require`, like here: https://github.com/sdv-dev/SDV/blob/8f3c787dc87ff2b85fe1ac392e0decf18ce134ad/setup.py#L94-L98

csala (changes requested):
Looks good except for a few minor things.

setup.py
```diff
@@ -81,6 +86,7 @@
     ],
     description='Reversible Data Transforms',
     extras_require={
+        'copulas': copulas_requires,
```
csala, line 89: I think you should add the `copulas_requires` to `test` and `dev` too:
```
'test': tests_require + copulas_requires,
'dev': development_requires + tests_require + copulas_requires,
```

csala (approved):
Looks good except for one comment about the `extras_require`.
Once that is fixed it can be merged.

PR #157, opened 2021-03-12
Issue 156 hypertransformer reverse transform memory increase

### LINKED ISSUE #156: HyperTransformer: Memory usage increase when reverse_transform is called
### Environment Details

* RDT version: 0.4.0
* Python version: Python 3.8
* Operating System: PopOS! 20.04

### Error Description

When trying to `reverse_transform`  categorical data using `rdt.transformers.OneHotEncoder` having more than 500 categories the memory usage is being increased drastically.

### Steps to reproduce
```python
In [1]: import pandas as pd
In [2]: data = pd.DataFrame({
   ...:     'categorical': [str(x) for x in range(600)] * 10
   ...: })

In [3]: from rdt.hyper_transformer import HyperTransformer 
In [4]: from rdt.transformers import OneHotEncodingTransformer
In [5]: ht = HyperTransformer(transformers={'categorical': OneHotEncodingTransformer()})
In [6]: ht.fit(data)
In [7]: transform = ht.transform(data)
In [8]: reverse = ht.reverse_transform(transform)
```
Screenshot from 2021-03-11 18-30-47

On the image can be observed the memory consumption which increases when `reverse_transform` is run.

This issue was originally raised on https://github.com/sdv-dev/SDV/issues/304

### Cause

The main cause of this memory increase is that using `pandas.DataFrame.pop` duplicates the data frame view and increases the memory used by the application instead of reducing it. When we try to `pop` more than 500 columns we end up creating 500+ views of the data frame which ends up increasing the memory drastically.

pvk-developer:
* Replace  `data.pop` with `get`.
* Store the columns that we want to `pop` and `drop` them afterwards.
* `_get_columns` now returns a list of columns instead of their values.
* Update tests.
* Resolve #156

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -173,20 +173,25 @@ class HyperTransformer:
         if self.copy:
             data = data.copy()
 
+        drop_columns = []
         for column_name, transformer in self._transformers.items():
             if column_name in data:
-                column = data.pop(column_name)
+                column = data[column_name]
                 transformed = transformer.transform(column)
 
                 shape = transformed.shape
 
                 if len(shape) == 2:
+                    drop_columns.append(column_name)
                     for index in range(shape[1]):
                         new_column = '{}#{}'.format(column_name, index)
                         data[new_column] = transformed[:, index]
 
                 else:
                     data[column_name] = transformed
+
+        if drop_columns:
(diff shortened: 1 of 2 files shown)


csala (approved):
LGTM!

PR #165, opened 2021-05-26
Fix CategoricalTransformer with NaN values to be pickled

### LINKED ISSUE #164: CategoricalTransformer with NaN values cannot be pickled
The CategoricalTransformer becomes unusable after being pickled and unpickled if it had `NaN` values in the data which it was fit on.

The problem seems to come from this line: https://github.com/sdv-dev/RDT/blob/cf62cc33c732f0bdec2aef9a88111fca11ae1be4/rdt/transformers/categorical.py#L103

When the transformer is fitted, one of the `intervals` takes `np.nan` as the key, so later on `np.nan` can be properly transformed.
However, if the transformer is pickled and unpickled, the `nan` value that ends up being used as the `key` for the `intervals` dictionary is a different `float('nan')` instances, which prevents the transformer from being able to `transform` a `np.nan` value.

Here is a snippet reproducing the issue:

```python
>>> import pickle
>>> import numpy as np
>>> from rdt.transformers import CategoricalTransformer
>>> 
>>> nans = np.array([np.nan])
>>> 
>>> cat = CategoricalTransformer()
>>> cat.fit(nans)
>>> cat.transform(nans)
array([0.5])
>>> 
>>> with open('test.pkl', 'wb') as f:
...     pickle.dump(cat, f)
... 
>>> with open('test.pkl', 'rb') as f:
...     cat2 = pickle.load(f)
... 
>>> cat2.transform(nans)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/nvme0n1p2/xals/.virtualenvs/SDV/lib/python3.8/site-packages/rdt/transformers/categorical.py", line 125, in transform
    return data.fillna(np.nan).apply(self._get_value).to_numpy()
  File "/mnt/nvme0n1p2/xals/.virtualenvs/SDV/lib/python3.8/site-packages/pandas/core/series.py", line 4212, in apply
    mapped = lib.map_infer(values, f, convert=convert_dtype)
  File "pandas/_libs/lib.pyx", line 2403, in pandas._libs.lib.map_infer
  File "/mnt/nvme0n1p2/xals/.virtualenvs/SDV/lib/python3.8/site-packages/rdt/transformers/categorical.py", line 103, in _get_value
    mean, std = self.intervals[category][2:]
KeyError: nan
```

To fix this, we can add a `__setstate__` method to the `CategoricalTransformer` to replace any `null` key by the actual `np.nan` instance:

```python
def __setstate__(self, state):
    intervals = state.get('intervals')
    if intervals:
        for key in list(intervals):
            if pd.isnull(key):
                intervals[np.nan] = intervals.pop(key)

    self.__dict__ = state
```

pvk-developer:
* Add integration test to reproduce the issue.
* Add  `__setstate__` as suggested by @csala that solves this issue by replacign the `nan` values to the current `np.nan` instance.

Resolves #164

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -36,6 +36,16 @@ class CategoricalTransformer(BaseTransformer):
     intervals = None
     dtype = None
 
+    def __setstate__(self, state):
+        """Replace any ``null`` key by the actual ``np.nan`` instance."""
+        intervals = state.get('intervals')
+        if intervals:
+            for key in list(intervals):
+                if pd.isnull(key):
+                    intervals[np.nan] = intervals.pop(key)
+
+        self.__dict__ = state
+
     def __init__(self, fuzzy=False, clip=False):
         self.fuzzy = fuzzy
         self.clip = clip
--- a/tests/integration/transformers/test_categorical.py
+++ b/tests/integration/transformers/test_categorical.py
@@ -1,3 +1,6 @@
+import pickle
+from io import BytesIO
+
 import numpy as np
 import pandas as pd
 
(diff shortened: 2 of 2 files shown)


csala (approved):
Looks good @pvk-developer !

PR #167, opened 2021-06-16
Add `rounding` argument to `NumericalTransformer`

### LINKED ISSUE #166: Add `rounding` argument to `NumericalTransformer`
### Problem Description

The `NumericalTransformer` could have a `rounding` argument which would accept multiple input values:
- `None or False`: Do not round anything
- `'auto'`: Figure out the rounding decimals from the data
- `<int>`: Round to the indicated amount of decimals (negatives accepted, which round to the indicated unit)

### Expected behavior

```python3
>>> nt = NumericalTransformer(rounding=None)   # Do not round
>>> nt.fit(np.array([123.456]))
>>> nt.reverse_transform(nt.transform(np.array([123.456789])))
array([123.456789])
>>> nt = NumericalTransformer(rounding='auto')    # Learn the rounding level (3 decimals) from fit data
>>> nt.fit(np.array([123.456]))
>>> nt.reverse_transform(nt.transform(np.array([123.456789])))
array([123.457])
>>> nt = NumericalTransformer(rounding=2)    # Round to 2 decimals
>>> nt.fit(np.array([123.456]))
>>> nt.reverse_transform(nt.fit_transform(np.array([123.456789])))
array([123.46])
>>> nt = NumericalTransformer(rounding=-2)    # Round to hundreds
>>> nt.fit(np.array([123.456]))
>>> nt.reverse_transform(nt.fit_transform(np.array([123.456789])))
array([100.0])
>>> nt = NumericalTransformer(rounding='auto')    # Learn the rounding level (hundreds) from the input
>>> nt.fit(np.array([100]))
>>> nt.reverse_transform(nt.fit_transform(np.array([123.456789])))
array([100])
```

amontanez24:
Resolves #166

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -11,7 +11,6 @@ from rdt.transformers.null import NullTransformer
 
 EPSILON = np.finfo(np.float32).eps
 MAX_DECIMALS = sys.float_info.dig - 1
-MIN_DECIMALS = -sys.float_info.max_10_exp
 
 
 class NumericalTransformer(BaseTransformer):
@@ -64,8 +63,10 @@ class NumericalTransformer(BaseTransformer):
             for decimal in range(MAX_DECIMALS):
                 if (np.allclose(clean_data, clean_data.round(decimal), rtol=0, atol=1e-15)):
                     return decimal
+
         else:
-            for decimal in range(MIN_DECIMALS, 1):
+            start = int(np.log10(max(data)))
+            for decimal in range(-start, 0):
                 if (clean_data == clean_data.round(decimal)).all():
                     return decimal
 
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -123,15 +123,15 @@ class NumericalTransformer(BaseTransformer):
         if self.nan is not None:
             data = self.null_transformer.reverse_transform(data)
 
+        if self.rounding_digits is not None:
+            data = data.round(self.rounding_digits)
+
         if np.dtype(self._dtype).kind == 'i':
             if pd.notnull(data).all():
                 return data.round().astype(self._dtype)
 
             data[pd.notnull(data)] = np.round(data[pd.notnull(data)]).astype(self._dtype)
             return data
-
-        if self.rounding_digits is not None:
-            data = data.round(self.rounding_digits)
 
         return data.astype(self._dtype)
 
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -128,7 +128,7 @@ class NumericalTransformer(BaseTransformer):
             data[pd.notnull(data)] = np.round(data[pd.notnull(data)]).astype(self._dtype)
             return data
 
-        if self.rounding:
+        if self.rounding_digits:
             data = data.round(self.rounding_digits)
 
         return data.astype(self._dtype)
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -1,5 +1,6 @@
 """Transformers for numerical data."""
 import copy
+import sys
 
 import numpy as np
 import pandas as pd
@@ -9,6 +10,8 @@ from rdt.transformers.base import BaseTransformer
 from rdt.transformers.null import NullTransformer
 
 EPSILON = np.finfo(np.float32).eps
+MAX_DECIMALS = sys.float_info.dig + 1
+MIN_DECIMALS = -sys.float_info.max_10_exp - 1
 
 
 class NumericalTransformer(BaseTransformer):
@@ -35,15 +38,34 @@ class NumericalTransformer(BaseTransformer):
             If ``True``, always create the new column whether there are null values or not.
             If ``False``, do not create the new column.
             Defaults to ``None``.
+        rounding (int, str or None):
+            Define rounding scheme for data. If set to an int, values will be rounded
+            to that number of decimal places. If ``None``, values will not be rounded.
+            If set to ``'auto'``, the transformer will round to the largest number of
+            decimal places seen in the fitted data.
     """
 
     null_transformer = None
+    rounding_digits = None
 
-    def __init__(self, dtype=None, nan='mean', null_column=None):
+    def __init__(self, dtype=None, nan='mean', null_column=None, rounding='auto'):
         self.nan = nan
         self.null_column = null_column
         self.dtype = dtype
         self._dtype = dtype
+        self.rounding = rounding
+
+    def _learn_rounding_digits(self, data):
+        if (data % 1 != 0).any():
+            for decimal in range(MAX_DECIMALS):
+                if (np.allclose(data, data.round(decimal), rtol=0, atol=1e-15)):
+                    return decimal
+        else:
+            for decimal in range(0, MIN_DECIMALS, -1):
+                if (data == data.round(decimal)).all():
+                    return decimal
+
+        return None
 
     def fit(self, data):
         """Fit the transformer to the data.
@@ -58,6 +80,8 @@ class NumericalTransformer(BaseTransformer):
         self._dtype = self.dtype or data.dtype
         self.null_transformer = NullTransformer(self.nan, self.null_column, copy=True)
         self.null_transformer.fit(data)
+        if self.rounding == 'auto':
+            self.rounding_digits = self._learn_rounding_digits()
 
     def transform(self, data):
         """Transform numerical data.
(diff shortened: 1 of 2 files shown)


rdt/transformers/numerical.py
```diff
@@ -35,15 +37,41 @@ class NumericalTransformer(BaseTransformer):
...
         self.null_column = null_column
         self.dtype = dtype
         self._dtype = dtype
+        self.rounding = rounding
+
+    @staticmethod
+    def _learn_rounding_digits(data):
+        clean_data = data.replace([pd.NA, None], np.nan)
+        if (clean_data % 1 != 0).any():
+            if not (clean_data == clean_data.round(MAX_DECIMALS)).all():
```
amontanez24, line 61: @csala  I think it is ok to compare like this instead of using `np.allclose` since we aren't rounding past 15 decimals places anymore, but I'm not sure.

csala: Yeah, I think so too. Let's keep it like that and make sure that the integration tests validate this thoroughly with random values, so if there are cases in which it fails we will eventually catch it there.

csala (changes requested):

rdt/transformers/numerical.py
```diff
@@ -35,15 +37,41 @@ class NumericalTransformer(BaseTransformer):
...
         self.nan = nan
         self.null_column = null_column
         self.dtype = dtype
         self._dtype = dtype
+        self.rounding = rounding
+
+    @staticmethod
+    def _learn_rounding_digits(data):
+        clean_data = data.replace([pd.NA, None], np.nan)
+        if (clean_data % 1 != 0).any():
```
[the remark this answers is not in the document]
amontanez24: This is to check if there are decimals. That way we only do the rounding check one way or the other

tests/transformers/test_numerical.py
```diff
@@ -33,6 +33,223 @@ def test_fit(self):
...
+    def test_fit_rounding_int(self):
+        """Test fit rounding parameter with int
+
+        If the rounding parameter is set to ``None``, the ``fit`` method
+        should not set its ``rounding`` or ``rounding_digits`` instance
+        variables.
+        """
+        # Setup
+        data = np.array([1.5, None, 2.5])
+        expected_digits = np.random.randint(10)
```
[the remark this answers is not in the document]
amontanez24: What is special about 0? Also, if the user provides an int >15, do we overwrite that?

rdt/transformers/numerical.py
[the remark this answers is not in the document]
amontanez24: This checks if there are any decimals so we know which direction to check

tests/transformers/test_numerical.py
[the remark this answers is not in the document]
amontanez24: If they provide >15, should anything be done? It wouldn't throw an error or anything so I'm inclined to just let users provide whatever they want.

tests/transformers/test_numerical.py
```diff
@@ -33,6 +33,223 @@ def test_fit(self):
...
+
+        # Run
+        transformer = NumericalTransformer(dtype=np.float, nan='nan', rounding=None)
+        transformer.fit(fitted_data)
+        result = transformer.reverse_transform(data)
+
+        # Assert
+        assert (result == data).all()
+
+    def test_reverse_transform_rounding_auto_decimals(self):
```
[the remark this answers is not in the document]
amontanez24: I don't think `>15` is actually a special case. It only matters when we are trying to determine the digits to round to since we can't accurately be more granular than that. But I think users should be able to provide that if they want, and it will behave the same as any other positive number except it might not always be accurate

tests/transformers/test_numerical.py
```diff
@@ -33,6 +33,194 @@ def test_fit(self):
...
+        # Setup
+        data = np.random.random(10)
+
+        # Run
+        transformer = NumericalTransformer(dtype=np.float, nan=None)
+        transformer.rounding_digits = None
+        result = transformer.reverse_transform(data)
+
+        # Assert
+        assert np.array_equal(result, data)
```
csala, line 150: For this it is better to use `np.testing.assert_array_equal`, which would give you information about which are the differences if they are not equal.

tests/transformers/test_numerical.py
```diff
@@ -33,6 +33,194 @@ def test_fit(self):
...
+        exponents = [np.random.randint(10, 20) for i in range(10)]
+        big_numbers = [10**exponents[i] for i in range(10)]
+        data = np.array(big_numbers)
+
+        # Run
+        transformer = NumericalTransformer(dtype=np.float, nan='nan', rounding='auto')
+        transformer.fit(data)
+
+        # Asserts
+        expected = -min(exponents) if min(exponents) > 0 else None
```
csala, line 111: Why is there an `if` here? Is `min(exponents)` not expected to be always between 10 and 20?

tests/transformers/test_numerical.py
```diff
@@ -33,6 +33,194 @@ def test_fit(self):
         assert transformer.null_transformer.fill_value == expect_fill_value
         assert transformer._dtype == expect_dtype
 
+    def test_fit_rounding_none(self):
+        """Test fit rounding parameter with ``None``
```
csala, line 37: Can the docstrings be expanded a bit to explain inputs, outputs and side effects?

For example, in this case:
```
Input:
- An array with floats rounded to 1 decimal.
Side Effect:
- rounding and rounding_digits continue to be `None`
```

tests/transformers/test_numerical.py
```diff
@@ -33,6 +33,194 @@ def test_fit(self):
...
+        transformer.fit(data)
+
+        # Asserts
+        expected = -min(exponents) if min(exponents) > 0 else None
+        assert transformer.rounding_digits == expected
+
+    def test_fit_rounding_auto_max_decimals(self):
+        """Test fit rounding parameter with ``'auto'``
+
+        If the ``rounding`` parameter is set to ``'auto'``,
```
csala, line 117: I think this docstring does not really match or explain the test.
From the docstring one could expect that the rounding_digits should be 15, but it is `None` instead. Is it intended? If so, why?

tests/transformers/test_numerical.py
```diff
@@ -33,6 +33,194 @@ def test_fit(self):
...
+        # Run
+        transformer = NumericalTransformer(dtype=np.float, nan=None)
+        transformer.rounding_digits = 2
+        result = transformer.reverse_transform(data)
+
+        # Assert
+        expected_data = np.array([1.11, 2.22, 3.33, 4.44, 5.56])
+        assert np.array_equal(result, expected_data)
+
+    def test_reverse_transform_rounding_negative_int(self):
```
csala, line 176: We should distinguish between integer and float values in two separated tests, since they should produce different outputs.
Also notice that reverse transform should always get floats as input!

tests/transformers/test_numerical.py
```diff
@@ -33,6 +33,194 @@ def test_fit(self):
...
+        attribute.
+
+        Input:
+        - Array with with larger numbers
+
+        Output:
+        - Same array rounded to the provided number of 0s
+        """
+        # Setup
+        data = np.array([2000, 120, 3100, 40100])
```
csala, line 189: As mentioned above, this input should be float values instead.

tests/transformers/test_numerical.py
```diff
@@ -33,6 +33,194 @@ def test_fit(self):
...
+
+        # Run
+        transformer = NumericalTransformer(dtype=np.float, nan=None)
```
csala, line 192: Since this is an integer test case, the `dtype` should be `int`, to guarantee that the output is an integer array.
A separated test case should be created to also round values to negative number of decimals but keep it a `float` array

tests/transformers/test_numerical.py
```diff
@@ -33,6 +33,194 @@ def test_fit(self):
...
+        # Setup
+        data = np.array([2000.554, 120.2, 3101, 4010])
+
+        # Run
+        transformer = NumericalTransformer(dtype=np.float, nan=None)
+        transformer.rounding_digits = 0
+        result = transformer.reverse_transform(data)
+
+        # Assert
+        expected_data = np.array([2001, 120, 3101, 4010])
```
csala, line 221: Should this not be a `float` array?

csala (changes requested):
Added a few comments to improve the tests

tests/transformers/test_numerical.py
amontanez24: I suppose. I did this because `[2001, 120, 3101, 4010]` and `[2001.0, 120.0, 3101.0, 4010.0]` will pass as equivalent

tests/transformers/test_numerical.py
amontanez24: The name here might be misleading. By `int` I meant that the argument is an `int`, not the input data

tests/transformers/test_numerical.py
amontanez24: This might have been something I forgot to change. Good catch!

csala (approved):
This looks good to go now!

PR #168, opened 2021-06-17
Add min and max arguments to NumericalTransformer

### LINKED ISSUE #106: Add min and max arguments to NumericalTransformer
Add `min_value` and `max_value` arguments to the `NumericalTransformer`.

Values for each one of the can be:
- `None`: means the data can go up to infinite (or -infinite)
- `'auto'`: means learn it from the data and stick to it
- `<any other value>`: Use the given value as the min/max

amontanez24:
Resolves #106

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -42,17 +42,31 @@ class NumericalTransformer(BaseTransformer):
             to that number of decimal places. If ``None``, values will not be rounded.
             If set to ``'auto'``, the transformer will round to the maximum number of
             decimal places detected in the fitted data.
+        min_value (int, str or None):
+            Indicate whether or not to set a minimum value for the data. If an integer is given,
+            reverse transformed data will be greater than or equal to it. If the string ``'auto'``
+            is given, the minimum will be the minimum value seen in the fitted data. If ``None``
+            is given, there won't be a minimum.
+        max_value (int, str or None):
+            Indicate whether or not to set a maximum value for the data. If an integer is given,
+            reverse transformed data will be less than or equal to it. If the string ``'auto'``
+            is given, the maximum will be the maximum value seen in the fitted data. If ``None``
+            is given, there won't be a maximum.
     """
 
     null_transformer = None
     rounding_digits = None
 
-    def __init__(self, dtype=None, nan='mean', null_column=None, rounding='auto'):
+
+    def __init__(self, dtype=None, nan='mean', null_column=None, rounding='auto',
+                 min_value=None, max_value=None):
         self.nan = nan
         self.null_column = null_column
         self.dtype = dtype
         self._dtype = dtype
         self.rounding = rounding
+        self.min_value = min_value
+        self.max_value = max_value
 
     @staticmethod
     def _learn_rounding_digits(data):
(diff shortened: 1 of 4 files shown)


rdt/transformers/numerical.py
```diff
@@ -58,6 +71,10 @@ def fit(self, data):
         self._dtype = self.dtype or data.dtype
         self.null_transformer = NullTransformer(self.nan, self.null_column, copy=True)
         self.null_transformer.fit(data)
+        if self.min_value == 'auto':
+            self.min_value = min(data)
```
csala, line 75: Maybe we should store the min and max values in a different attribute (`_min_value` and `_max_value`) to avoid overwriting the user input, so a single instance can be fitted multiple times on different data with consistent behavior.

rdt/transformers/numerical.py
```diff
@@ -90,6 +107,12 @@ def reverse_transform(self, data):
         if self.nan is not None:
             data = self.null_transformer.reverse_transform(data)
 
+        if self.max_value is not None:
```
csala, line 110: I think using `clip` here would make it simpler and faster.

```
data = data.clip(self._min_value, self._max_value)
```

Also, this should probably need to be done before the `null_transformer.reverse_transform`, otherwise null values will make it crash

tests/transformers/test_numerical.py
```diff
@@ -33,6 +33,114 @@ def test_fit(self):
...
+        # Run
+        transformer = NumericalTransformer(dtype=np.float, nan='nan',
+                                           min_value='auto', max_value='auto')
+        transformer.fit(data)
+
+        # Asserts
+        assert transformer.min_value == -5000
+        assert transformer.max_value == 4000
+
+    def test_reverse_transform_min_no_max(self):
```
csala, line 92: Maybe we should add tests with null values? The null values need to be passed by making the input a 2d matrix with a second column that has values from 0 to 1 with at least 1 value above 0.5

csala (changes requested):
Added a few comments to improve performance and testing

rdt/transformers/numerical.py
amontanez24: @csala After running the timer tests for this way of clipping vs `clip` I found that the time returned for `clip` on an array of 1000 values was `0.06289540000034322` and the time returned for this custom clipping was `0.061169600000084756`. This shows the custom clipping is slightly faster, but I don't know if it's fast enough to justify the extra complication

rdt/transformers/numerical.py
amontanez24: I added those variables but I get the following after running `pylint rdt --rcfile=setup.cfg`: 

`rdt\transformers\numerical.py:14:0: R0902: Too many instance attributes (9/7) (too-many-instance-attributes)`

rdt/transformers/numerical.py
```diff
@@ -58,6 +73,15 @@ def fit(self, data):
         self._dtype = self.dtype or data.dtype
         self.null_transformer = NullTransformer(self.nan, self.null_column, copy=True)
         self.null_transformer.fit(data)
+        if self.min_value == 'auto':
+            self._min_value = min(data)
```
pvk-developer, line 77: I have the following situation and maybe we would like to write a test for it:
```python
data = np.array([1.5, None, 2.5])
# Run
transformer = NumericalTransformer(dtype=np.float, nan='nan', min_value='auto', max_value=None)
transformer.fit(data)
# Error
..........
TypeError: '<' not supported between instances of 'NoneType' and 'float'
```

csala: Good catch @pvk-developer 
I suppose we can from the null values from `data` before searching for the min and max

rdt/transformers/numerical.py
csala: Yes, being such a small difference I would go for the simpler `clip`

csala (changes requested):

csala (approved):
I pushed a change myself to move the `_min_value` and `_max_value` initialization to the root of the class (along with `_dtype`), and also renamed `rounding_digits` to `_rounding_digits` to keep behavior consistent.
I think this is now ready to go (once the tests pass)

pvk-developer (approved):
LGTM!

PR #177, opened 2021-07-13
Create script for profiling

### LINKED ISSUE #176: Create script for profiling
### Desired Outcome

1. Create file called `profiling.py`
2. the script should have a function `profile_transformer` that takes in the following:
    1. the transformer instance to run (ie. `NumericalTransformer`)
    2. the dataset generator to use
    3. the number of rows for transforming
    4. the number of rows for fitting
3. The file should have a `__main__` and be runnable from the command line

amontanez24:
resolves #176


tests/performance/profiling.py
```diff
@@ -0,0 +1,77 @@
...
+    peak_memory = Value('i', 0)
+    profiling_process = Process(target=_set_memory_for_method, args=(method, dataset, peak_memory))
+    profiling_process.start()
+    profiling_process.join()
+    return peak_memory.value
+
+
+def profile_transformer(transformer, dataset_generator, transform_size, fit_size=None):
+    """
+    Function to profile a transformer.
```
csala, line 35: This should be on the line above. I also suspect that `lint` may not be checking this folder properly, since I think this should be reported by `pylint`

tests/performance/profiling.py
```diff
@@ -0,0 +1,77 @@
...
+    Args:
+        transformer (Transformer):
+            Transformer instance.
+        dataset_generator (DatsetGenerator):
+            DatasetGenerator instance.
+        transform_size (int):
+            Number of rows to generate for ``transform`` and ``reverse_transform``.
+        fit_size (int or None):
+            Number of rows to generate for ``fit``. If None, use ``transform_size``.
+    """
```
csala, line 51: I would also add a `Return` explaining what the output is

tests/performance/profiling.py
```diff
@@ -0,0 +1,77 @@
...
+    transformer_instance = transformer()
```
csala, line 52: I think this line should not be required, since `transformed` is expected to already be an instance of the class.

tests/performance/profiling.py
```diff
@@ -0,0 +1,77 @@
...
+    fit_size = fit_size or transform_size
+    fit_dataset = dataset_generator.generate(fit_size)
+    fit_time = _profile_time(transformer_instance.fit, fit_dataset)
```
csala, line 55: Is running this twice really needed? In other words, is there any significant alteration of the results when profiling both things at the same time?
If so, maybe we should fit a deepcopy of the instance on the first execution, so we do not re-fit the same instance. Otherwise, we may be altering the output of the second call (for example, the size of the instance when the fit process  starts may be bigger after having being fitted once)

tests/performance/profiling.py
```diff
@@ -0,0 +1,77 @@
...
+    fit_memory = _profile_memory(transformer_instance.fit, fit_dataset)
+
+    transform_dataset = dataset_generator.generate(transform_size)
+    transform_time = _profile_time(transformer_instance.transform, transform_dataset)
```
csala, line 59: same as above wrt to running twice, but here the deepcopy may not be necessary, since transform and reverse_transform should not alter the internal state of the transformer.

tests/performance/profiling.py
```diff
@@ -0,0 +1,77 @@
...
+    transform_memory = _profile_memory(transformer_instance.transform, transform_dataset)
+
+    reverse_dataset = transformer_instance.transform(transform_dataset)
+    reverse_time = _profile_time(transformer_instance.reverse_transform, reverse_dataset)
+    reverse_memory = _profile_memory(transformer_instance.reverse_transform, reverse_dataset)
+
+    return pd.DataFrame({
```
csala, line 66: Since this is only one row, maybe we should return a Series instead of a DataFrame?

tests/unit/test_profiling.py
```diff
@@ -0,0 +1,65 @@
+from multiprocessing import Process
```
csala, line 1: Maybe we could place this within the `tests/performance` folder, as `tests/performance/tests/test_profiling.py`, since this is not part of the `rdt` package (notice that the `unit` folder follows the same hierarchy as `rdt`, so `tests/unit/test_profiling.py` should be dedicated to testing `rdt/profiling.py`

tests/performance/profiling.py
amontanez24: what do you mean by run twice?

tests/performance/profiling.py
```diff
@@ -0,0 +1,77 @@
+import timeit
+import tracemalloc
+from multiprocessing import Process, Value
+
+import pandas as pd
+
+
+def _profile_time(method, dataset, iterations=100):
+    total_time = 0
+    for _ in range(iterations):
```
csala, line 10: I now realized that we are iterating here!
So yes, it seems it makes sense to profile time and memory separately, unless we would want to also average the memory consumed over the iterations.

On the other hand, if we decide to do a `deepcopy` maybe it should be happening here, before capturing the start time, rather than outside before calling this function. Also, since the `deepcopy` is needed only for `fit`, it should be made optional with an argument.

tests/performance/profiling.py
amontanez24: The linter definitely complains about this file when I do certain things. Idk why that wasn't reported

tests/performance/profiling.py
```diff
@@ -0,0 +1,88 @@
...
+        'Fit Time': fit_time,
+        'Fit Memory': fit_memory,
+        'Transform Time': transform_time,
+        'Transform Memory': transform_memory,
+        'Reverse Transform Time': reverse_time,
+        'Reverse Transform Memory': reverse_memory
+    })
+
+
+if __name__ == '__main__':
```
csala, line 87: We can remove this, at least while it does nothing.

tests/performance/test_performance.py
```diff
@@ -0,0 +1,17 @@
+def test_performance():
```
csala, line 1: I think we can remove this file and the example.json below, since the current PR is not intended to cover test cases yet

tests/performance/tests/example_test_config.json
```diff
@@ -13,4 +13,4 @@
             }
         }
     ]
-}
\ No newline at end of file
+}
```
csala, line 16: This entire file can be removed

csala (approved):
I think the PR looks good and can be merged.
I would just remove a couple of things that do not seem related to the PR topic and which can always be added later on when the corresponding PR is made.

PR #185, opened 2021-07-15
Audit the LabelEncodingTransformer

### LINKED ISSUE #184: Audit the LabelEncodingTransformer
### Fit: 
Most of the time is in getting the `self.values_to_categories` map. When digging deeper, I found that about 90% of the time is from running `data.unique()` and 9% is from converting it to a `pd.Series`. Replacing that line with 
```
self.values_to_categories = dict(enumerate(data.unique()))
```
cuts the time by more than half.

### Transform:
The `transform` function is only one line that can't really be improved much. Using `apply` instead of `map` seems to yield similar results.

### Reverse Transform
About half of the time seems to come from the `map` call, and the other half comes from the `astype(int)` call. The `astype(int)` call seems to be unnecessary though, so this could reduce the time.

amontanez24:
resolves #184

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -298,7 +298,7 @@ class LabelEncodingTransformer(BaseTransformer):
             data (pandas.Series or numpy.ndarray):
                 Data to fit the transformer to.
         """
-        self.values_to_categories = pd.Series(data.unique()).to_dict()
+        self.values_to_categories = dict(enumerate(data.unique()))
         self.categories_to_values = {
             category: value
             for value, category in self.values_to_categories.items()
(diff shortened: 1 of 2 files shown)


rdt/transformers/categorical.py
```diff
@@ -329,4 +329,4 @@ def reverse_transform(self, data):
         if isinstance(data, np.ndarray) and (data.ndim == 2):
             data = data[:, 0]
 
-        return pd.Series(data).astype(int).map(self.values_to_categories)
+        return pd.Series(data).map(self.values_to_categories)
```
csala, line 332: Is this change right, considering that input could be float values too?
On the other side, maybe the `astype(int)` should rather be `round().astype(int)` instead?

amontanez24: Can the input be floats? `transform` should only yield `int` values right? Also replacing with `round().astype(int)` shouldn't help since it's the `astype` part that is slow. However maybe we can just `round` since if you have `d={1.0: 'a'}` and then do `d[1]` you will still get the value

csala: Yes, reverse transform should always work properly with float inputs, since those are the values outputted by the underlying models.
On the other side, I agree that the `astype` may not be needed, so we can possibly do `round` only.

csala (approved):
This looks good!

PR #186, opened 2021-07-15
Optimize one hot encoder

### LINKED ISSUE #178: Auditing OneHotEncodingTransformer
We audit time and memory performance of `OneHotEncodingTransformer`

### fitting
most of the time is consumed by 

https://github.com/sdv-dev/RDT/blob/efddacffd8497ee0cee53349d6373cca7cb9c12d/rdt/transformers/categorical.py#L233

```
CPU times: user 3.3 ms, sys: 159 µs, total: 3.46 ms
Wall time: 6.41 ms
peak memory: 150.72 MiB, increment: 0.00 MiB
```

### transform
similarly most of the time is consumed by

https://github.com/sdv-dev/RDT/blob/efddacffd8497ee0cee53349d6373cca7cb9c12d/rdt/transformers/categorical.py#L246

```
CPU times: user 3.18 ms, sys: 897 µs, total: 4.07 ms
Wall time: 4.87 ms
peak memory: 185.29 MiB, increment: 0.00 MiB
```

### reverse transform
mapping the index back to the original map is what takes most of the time (but not that much) 

https://github.com/sdv-dev/RDT/blob/efddacffd8497ee0cee53349d6373cca7cb9c12d/rdt/transformers/categorical.py#L268
```
CPU times: user 1.89 ms, sys: 0 ns, total: 1.89 ms
Wall time: 5.48 ms
peak memory: 185.05 MiB, increment: 0.00 MiB
```

sarahmish:
Resolve #178


tests/unit/transformers/test_categorical.py
```diff
@@ -310,6 +310,24 @@ def test_transform_all_zeros(self):
         with np.testing.assert_raises(ValueError):
             ohet.transform(['b'])
 
+    def test_transform_large(self):
```
csala, line 313: This seems to be an integration test (black-box testing) rather than a unit test (white-box testing).
I think we should go a bit deeper on the testing and do separated tests for each one of the new methods, and then a specific test on the `transform` to check whether the right transform is being used internally.

rdt/transformers/categorical.py
```diff
@@ -243,11 +268,17 @@ def transform(self, data):
...
         """
         data = self._prepare_data(data)
-        dummies = pd.get_dummies(data, dummy_na=self.dummy_na)
-        array = dummies.reindex(columns=self.dummies, fill_value=0).values.astype(int)
-        for i, row in enumerate(array):
-            if np.all(row == 0) and self.error_on_unknown:
-                raise ValueError(f'The value {data[i]} was not seen during the fit stage.')
+        size = len(data)
+
+        if size < 2800:
```
csala, line 273: I'm curious about this boundary: Is it always the same independently on the amount of unique categories involved?

rdt/transformers/categorical.py
```diff
@@ -243,11 +268,17 @@ def transform(self, data):
...
+            array = self._transform_small(data, size)
+        else:
+            array = self._transform_large(data)
+
+        unknown = array.sum(axis=1) == 0
+        if self.error_on_unknown and unknown.any():
+            raise ValueError(f'Attempted to transform {list(data[unknown])} \
```
csala, line 280: I would use multi-line concatenation using parenthesis rather than `\`

csala (changes requested):
I think the code seems to be right on spot. The only concerns would be:
1. The boundary between large and small, which may need to be dynamic
2. The unit tests, which could be a bit more thorough

csala (approved):
I think this looks good. Just one comment about an attribute that may be unnecessary

rdt/transformers/categorical.py
```diff
@@ -229,8 +253,19 @@ def fit(self, data):
...
         data = self._prepare_data(data)
-        self.dummy_na = pd.isnull(data).any()
-        self.dummies = list(pd.get_dummies(data, dummy_na=self.dummy_na).columns)
+
+        null = pd.isnull(data)
+        self.dummy_na = null.any()
+        self.dummies = list(pd.unique(data[~null]))
+        self.num_dummies = len(self.dummies)
+        self.indexer = list(range(self.num_dummies))
+        self.decoder = self.dummies.copy()
```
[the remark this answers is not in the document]
sarahmish: ideally that should be the case but the main difference is `self.dummies` is always free from `np.nan` while `self.decoder` contains `np.nan` when `self.dummy_na` is set to true. The reason for this is that `pd.Categorical` does not allow null values to be a category, so I resolve this independently by having two attributes.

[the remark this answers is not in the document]
csala: Of course! That make total sense.

PR #188, opened 2021-07-16
Optimize the `NumericalTransformer`

### LINKED ISSUE #181: Audit the `NumericalTransformer`
Audit for the time/memory performance of the `NumericalTransformer`.

In the `fit` stage, besides fitting the `NullTransformer`, the most time is consumed by:
https://github.com/sdv-dev/RDT/blob/05b4036bddca04ab76967dc9d2de39317d2d7ab1/rdt/transformers/numerical.py#L105

In the `transform` stage, most of the time is consumed by:
https://github.com/sdv-dev/RDT/blob/05b4036bddca04ab76967dc9d2de39317d2d7ab1/rdt/transformers/numerical.py#L131

In the `reverse_transform` stage, most of the time is consumed by:
https://github.com/sdv-dev/RDT/blob/05b4036bddca04ab76967dc9d2de39317d2d7ab1/rdt/transformers/numerical.py#L164

fealho:
Resolves #181.

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -122,7 +122,7 @@ class NumericalTransformer(BaseTransformer):
         Args:
             data (pandas.Series or numpy.ndarray):
                 Data to transform.
-                
+
         Returns:
             numpy.ndarray
         """
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -68,11 +68,12 @@ class NumericalTransformer(BaseTransformer):
         self.rounding = rounding
         self.min_value = min_value
         self.max_value = max_value
+        self.null_transformer = NullTransformer(self.nan, self.null_column, copy=True)
 
     @staticmethod
     def _learn_rounding_digits(data):
         # check if data has any decimals
-        roundable_data = data[~np.isinf(data)]
+        roundable_data = data[~(np.isinf(data) | ~pd.notnull(data))]
         roundable_data = roundable_data.astype(float)
         if (roundable_data % 1 != 0).any():
             if not (roundable_data == roundable_data.round(MAX_DECIMALS)).all():
@@ -93,7 +94,6 @@ class NumericalTransformer(BaseTransformer):
 
     def fit(self, data):
         """Fit the transformer to the data.
-
         Args:
             data (pandas.Series or numpy.ndarray):
                 Data to fit.
@@ -102,28 +102,23 @@ class NumericalTransformer(BaseTransformer):
             data = pd.Series(data)
 
         self._dtype = self.dtype or data.dtype
-        clean_data = data.dropna()
-        self._min_value = min(clean_data) if self.min_value == 'auto' else self.min_value
-        self._max_value = max(clean_data) if self.max_value == 'auto' else self.max_value
+        self._min_value = data.min() if self.min_value == 'auto' else self.min_value
+        self._max_value = data.max() if self.max_value == 'auto' else self.max_value
 
         if self.rounding == 'auto':
-            self._rounding_digits = self._learn_rounding_digits(clean_data)
+            self._rounding_digits = self._learn_rounding_digits(data)
         elif isinstance(self.rounding, int):
             self._rounding_digits = self.rounding
 
-        self.null_transformer = NullTransformer(self.nan, self.null_column, copy=True)
         self.null_transformer.fit(data)
 
     def transform(self, data):
         """Transform numerical data.
-
         Integer values are replaced by their float equivalent. Non null float values
         are left unmodified.
-
         Args:
             data (pandas.Series or numpy.ndarray):
                 Data to transform.
-
         Returns:
             numpy.ndarray
         """
@@ -134,11 +129,9 @@ class NumericalTransformer(BaseTransformer):
 
     def reverse_transform(self, data):
         """Convert data back into the original format.
-
         Args:
             data (numpy.ndarray):
                 Data to transform.
-
         Returns:
             numpy.ndarray
         """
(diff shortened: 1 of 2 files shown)


rdt/transformers/numerical.py
```diff
@@ -68,11 +68,12 @@ def __init__(self, dtype=None, nan='mean', null_column=None, rounding=None,
         self.rounding = rounding
         self.min_value = min_value
         self.max_value = max_value
+        self.null_transformer = NullTransformer(self.nan, self.null_column, copy=True)
```
csala, line 71: I do not think this change should be done, since we will want to re-create the NullTransformer instance if the `fit` method is called more than once.

rdt/transformers/numerical.py
```diff
@@ -68,11 +68,12 @@ def __init__(self, dtype=None, nan='mean', null_column=None, rounding=None,
...
 
     @staticmethod
     def _learn_rounding_digits(data):
         # check if data has any decimals
-        roundable_data = data[~np.isinf(data)]
+        roundable_data = data[~(np.isinf(data) | ~pd.notnull(data))]
```
csala, line 76: The `~pd.notnull` can be changed to `pd.isnull`

rdt/transformers/numerical.py
```diff
@@ -123,7 +122,7 @@ def transform(self, data):
         Args:
             data (pandas.Series or numpy.ndarray):
                 Data to transform.
-
+
```
csala, line 125: These spaces should be removed.

rdt/transformers/numerical.py
```diff
@@ -134,7 +133,6 @@ def transform(self, data):
 
     def reverse_transform(self, data):
         """Convert data back into the original format.
-
```
csala, line 137: This line should be kept.

rdt/transformers/numerical.py
```diff
@@ -158,7 +156,7 @@ def reverse_transform(self, data):
             if pd.notnull(data).all():
                 return data.round().astype(self._dtype)
 
-            data[pd.notnull(data)] = np.round(data[pd.notnull(data)]).astype(self._dtype)
+            data = np.round(data)
```
csala, line 159: We can do `return np.round(data)` directly

csala (changes requested):

tests/unit/transformers/test_numerical.py
```diff
@@ -176,7 +247,7 @@ def test_fit_rounding_auto_max_inf(self):
         transformer.fit(data)
 
         # Asserts
-        assert transformer._rounding_digits is -3
+        assert transformer._rounding_digits == -3
```
amontanez24, line 250: I'm surprised this didn't throw a lint error

rdt/transformers/numerical.py
```diff
@@ -102,12 +102,11 @@ def fit(self, data):
             data = pd.Series(data)
 
         self._dtype = self.dtype or data.dtype
-        clean_data = data.dropna()
```
amontanez24, line 105: Why did we get rid of this? If you do `data.min()` on data that has a `np.nan` value, it will return `np.nan` as the min. Same with max

rdt/transformers/numerical.py
```diff
@@ -151,14 +150,10 @@ def reverse_transform(self, data):
...
-            data = data.round(self._rounding_digits)
+        if self._rounding_digits is not None or np.dtype(self._dtype).kind == 'i':
+            data = data.round(self._rounding_digits or 0)
 
-        if np.dtype(self._dtype).kind == 'i':
-            if pd.notnull(data).all():
-                return data.round().astype(self._dtype)
-
-            data[pd.notnull(data)] = np.round(data[pd.notnull(data)]).astype(self._dtype)
+        if pd.isnull(data).any():
```
amontanez24, line 156: why are we returning data here?

amontanez24 (commented):
I mainly have one concern about not dropping nans before getting the min and max

tests/unit/transformers/test_numerical.py
csala: Yeah, me too. Because it actually complains now!

rdt/transformers/numerical.py
csala: No, `data.min()` and `data.max()` seem to return the right values. Bear in mind that at this point `data` is a `pd.Series`, so it works around the NaNs.

```
In [2]: pd.Series([1, 2, None]).min()
Out[2]: 1.0
```

rdt/transformers/numerical.py
csala: Because if there are nulls we cannot do the `astype` below, otherwise an `astype(np.int)` will silently replace the NaNs with the minimum available integer value.
I will add another check though, to ensure that we only enter this `if` if the dtype is actually an integer.

rdt/transformers/numerical.py
amontanez24: that explains my confusion

rdt/transformers/numerical.py
```diff
@@ -151,14 +150,11 @@ def reverse_transform(self, data):
         if self.nan is not None:
             data = self.null_transformer.reverse_transform(data)
 
-        if self._rounding_digits is not None:
-            data = data.round(self._rounding_digits)
+        is_integer = np.dtype(self._dtype).kind == 'i'
+        if self._rounding_digits is not None or is_integer:
+            data = data.round(self._rounding_digits or 0)
```
amontanez24, line 155: if `_rounding_digits` is not None, why do we need to do `or 0`?

csala: Because if `is_integer` is `True` we enter this branch even if `self._rounding_digits` is `None`. In that case, we want to round to `0` rather than `None`.

amontanez24 (approved):
Alright, this looks good to me now!

PR #193, opened 2021-07-20
Audit the NullTransformer

### LINKED ISSUE #192: Audit the NullTransformer
### Fit
When the fill value is provided, 99% of the time of fit is spent on the 
`self.nulls = data.isnull().any()`
 line. It appears to be slightly faster to do 
`self.nulls = data.isnull().values.any()`

When the fill value is set to mean or mode, most of the time is spent on the lines 
`self._fill_value = data.mean() if pd.notnull(data).any() else 0` or
`self._fill_value = data.mode(dropna=True)[0] if pd.notnull(data).any() else 0` 

We already have to get the values that are null later so I suggest just getting the null values once at the top, and then using that array to figure out self.nulls and another boolean that tells us if any values are not null. This reduces the time by about half.

### Transform
After analyzing the different scenarios, the slowest line in the transform function is 
`data[isnull] = self._fill_value`
 which executes if copy is False. I don't really see a faster way to implement this. The next most taxing line is return `pd.concat([data, isnull.astype('int')], axis=1).values` 
which happens if `self._null_column` is True. Again, I don't see a better way to do this since the columns need to be combined.

One problem with the code however is that the docstring suggests that data can be a numpy array, but the first line in `transform` is `isnull = data.isnull()`. Numpy arrays don't have an `isnull()` method so this will crash. We should probably convert to a pandas series.

### Reverse Transform
If `null_column` is `True` then about 43% of the time is spent on 
`data = pd.Series(data[:, 0])`
If it's `False`, then about 40% of the time is spent on 
`pd.Series(data)`
So in both cases a good chunk is spent just converting to a Series. The calculation of `isnull` is about the same time for both branches. I don't think there is a more efficient way to do this.

The next most time consuming line is `data.iloc[isnull] = np.nan`. Again, I don't know of any way to do this more efficiently.

amontanez24:
resolves #192

### DIFF
--- a/rdt/transformers/null.py
+++ b/rdt/transformers/null.py
@@ -47,14 +47,16 @@ class NullTransformer(BaseTransformer):
             data (pandas.Series or numpy.ndarray):
                 Data to transform.
         """
+        null_values = data.isnull().values
+        self.nulls = null_values.any()
+        contains_not_null = ~null_values.any()
         if self.fill_value == 'mean':
-            self._fill_value = data.mean() if pd.notnull(data).any() else 0
+            self._fill_value = data.mean() if contains_not_null else 0
         elif self.fill_value == 'mode':
-            self._fill_value = data.mode(dropna=True)[0] if pd.notnull(data).any() else 0
+            self._fill_value = data.mode(dropna=True)[0] if contains_not_null else 0
         else:
             self._fill_value = self.fill_value
 
-        self.nulls = data.isnull().any()
         if self.null_column is None:
             self._null_column = self.nulls
         else:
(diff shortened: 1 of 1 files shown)


rdt/transformers/null.py
```diff
@@ -47,14 +47,16 @@ def fit(self, data):
             data (pandas.Series or numpy.ndarray):
                 Data to transform.
         """
+        null_values = data.isnull().values
+        self.nulls = null_values.any()
+        contains_not_null = (~null_values).any()
```
csala, line 52: I'm curious about whether `not null_values.all()` makes any difference.

csala (approved):
A comment which is more a question than anything else.
Other than that, this looks good to go

rdt/transformers/null.py
amontanez24: I'll try it out

amontanez24: looks _slightly_ faster. I can change that

PR #195, opened 2021-07-20
Add performance test file

### LINKED ISSUE #194: Add performance-based testing
We should add a way to test RDT transformers based on performance and resource usage by running the `profiling` function recently added and comparing the obtained results with some expected maximum time and memory usage.

The tests should be implemented as a single `test` module with a single test file that is parameterized with a collection of test cases found in configuration files written in JSON format.

Here would be the requirements of this:

1. The test file should be in the module:
`tests/performance/test_performance.py`.
2. The test cases should be inside a `tests/performance/test_cases` folder.
3. Test cases for each transformer can be put inside a folder dedicated to each one of them (e.g. `tests/performance/test_cases/NumericalTransformer/<some_test_case>.json`
4. The test function should be parametrized with the config files found inside the `test_cases` folder.
5. It should use the `profile_transformer` function on all the transformers and datasets defined by the config files

amontanez24:
resolves #194

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -80,3 +80,19 @@ jobs:
         pip install tox tox-gh-actions
     - name: Test with tox
       run: tox -e minimum
+
+  performance:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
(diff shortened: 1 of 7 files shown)


csala (approved):
Looks good!

PR #199, opened 2021-07-23
Add boolean generators

katxiao:
Add data generators for the boolean transformer + tests. Also add performance test cases using the generators.


csala (approved):
This looks good @katxiao !

PR #200, opened 2021-07-23
Automate performance test case creation and add numerical tests

csala:
Add functions to automate the creation of performance test cases.

Usage example:

```python3
In [1]: from tests.performance.test_performance import make_test_case_configs

In [2]: import rdt

In [3]: from tests.performance.datasets import numerical

In [4]: dataset_generators = [
   ...: numerical.RandomIntegerGenerator,
   ...: numerical.RandomIntegerNaNsGenerator,
   ...: numerical.NormalGenerator,
   ...: numerical.NormalNaNsGenerator,
   ...: ]

In [5]: transformers = [
   ...: rdt.transformers.NumericalTransformer,
   ...: (rdt.transformers.NumericalTransformer, {'rounding': 'auto', 'min_value': 'auto', 'max_value': 'auto'}, 'auto')
   ...: ]

In [6]: sizes = [(1000, 1000), (10_000, 10_000)]

In [7]: make_test_case_configs(transformers, dataset_generators, sizes, iterations=10)
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/default_RandomIntegerGenerator_1000_1000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/default_RandomIntegerGenerator_10000_10000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/default_RandomIntegerNaNsGenerator_1000_1000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/default_RandomIntegerNaNsGenerator_10000_10000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/default_NormalGenerator_1000_1000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/default_NormalGenerator_10000_10000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/default_NormalNaNsGenerator_1000_1000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/default_NormalNaNsGenerator_10000_10000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/auto_RandomIntegerGenerator_1000_1000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/auto_RandomIntegerGenerator_10000_10000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/auto_RandomIntegerNaNsGenerator_1000_1000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/auto_RandomIntegerNaNsGenerator_10000_10000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/auto_NormalGenerator_1000_1000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/auto_NormalGenerator_10000_10000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/auto_NormalNaNsGenerator_1000_1000.json
Test case created: tests/performance/test_cases/numerical/NumericalTransformer/auto_NormalNaNsGenerator_10000_10000.json
```

Also add the test cases for the NumericalTransformer using all the numerical generators and 1k, 10k and 100k values.

### DIFF
--- a/tests/performance/test_performance.py
+++ b/tests/performance/test_performance.py
@@ -1,9 +1,13 @@
 """Test whether the performance of the Transformers is the expected one."""
 
+import hashlib
 import importlib
 import json
+import os
 import pathlib
 
+import numpy as np
+import pandas as pd
 import pytest
 
 from tests.performance.profiling import profile_transformer
@@ -22,6 +26,11 @@ def get_instance(obj, **kwargs):
         instance = getattr(importlib.import_module(package), name)(**kwargs)
 
     return instance
+
+
+def get_fqn(obj):
+    """Get the fully qualified name of the given object."""
+    return f'{obj.__module__}.{obj.__name__}'
 
 
 TEST_CASES_PATH = pathlib.Path(__file__).parent / 'test_cases'
@@ -54,11 +63,15 @@ def test_performance(config_path):
     with open(config_path, 'r') as config_file:
         config = json.load(config_file)
 
-    transformer = get_instance(config['transformer'], **config['kwargs'])
-    dataset_gen = get_instance(config['dataset'])
+    transformer_instance = get_instance(config['transformer'], **config['kwargs'])
+    dataset_generator = get_instance(config['dataset'])
 
-    out = profile_transformer(transformer, dataset_gen, config['transform_size'],
-                              config['fit_size'])
+    out = profile_transformer(
+        transformer=transformer_instance,
+        dataset_generator=dataset_generator,
+        fit_size=config['fit_size'],
+        transform_size=config['transform_size'],
+    )
 
     assert out['Fit Time'] < config['expected']['fit']['time']
     assert out['Fit Memory'] < config['expected']['fit']['memory']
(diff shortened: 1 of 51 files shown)


tests/performance/test_performance.py
```diff
@@ -54,15 +63,177 @@ def test_performance(config_path):
...
+
+    os.makedirs(output_path.parent, exist_ok=True)
+    with open(output_path, 'w') as output_file:
+        json.dump(test_case, output_file, indent=4)
+
+    cwd = os.getcwd()
+    if str(output_path).startswith(cwd):
+        output_path = str(output_path)[len(cwd) + 1:]
+
+    print(f'Test case created: {output_path}')
```
amontanez24, line 206: When do we want to use print statements vs loggers?

tests/performance/test_performance.py
```diff
@@ -54,15 +63,177 @@ def test_performance(config_path):
...
     assert out['Transform Time'] < config['expected']['transform']['time']
     assert out['Transform Memory'] < config['expected']['transform']['memory']
     assert out['Reverse Transform Time'] < config['expected']['reverse_transform']['time']
     assert out['Reverse Transform Memory'] < config['expected']['reverse_transform']['memory']
+
+
+def _round_to_magnitude(value):
+    for digits in range(-15, 15):
+        rounded = np.round(value, digits)
+        if rounded != 0:
```
amontanez24, line 87: I wouldn't expect this to happen so this should be safe, but is it at all possible for the value to be to small or actually 0?

amontanez24 (commented):
This looks good! I didn't really find any issues. My only question is if we should move all of the test config creator functionality into its own file. Like a script file people could run if they want to do this

tests/performance/test_performance.py
```diff
@@ -54,15 +63,177 @@ def test_performance(config_path):
...
+    return (means * 10).apply(_round_to_magnitude)
+
+
+def make_test_case_config(transformer_class, transformer_kwargs, dataset_generator,
+                          fit_size, transform_size, iterations=1, output_path=None,
+                          config_name=None):
+    """Create a Test Case JSON file for the indicated transformer and dataset.
+
+    If output path is not given, the test case is created with the filename
+    ``{dataset_generator}_{fit_size}_{transform_size}.json`` inside the folder
```
sarahmish, line 137: I think we should update this to `{config}_{dataset_generator}_{fit_size}_{transform_size}.json`

tests/performance/test_performance.py
```diff
@@ -54,15 +63,177 @@ def test_performance(config_path):
...
     assert out['Fit Time'] < config['expected']['fit']['time']
     assert out['Fit Memory'] < config['expected']['fit']['memory']
     assert out['Transform Time'] < config['expected']['transform']['time']
     assert out['Transform Memory'] < config['expected']['transform']['memory']
     assert out['Reverse Transform Time'] < config['expected']['reverse_transform']['time']
     assert out['Reverse Transform Memory'] < config['expected']['reverse_transform']['memory']
+
+
+def _round_to_magnitude(value):
+    for digits in range(-15, 15):
```
sarahmish, line 85: I'm wondering if this range can be problematic, if the metric is large in value then I'm questioning this example:
```
np.round(123, -2) # 100
np.round(123, -1) # 120
```
which case is the desired case?

sarahmish (approved):
Looks good to me!

tests/performance/test_performance.py
amontanez24: I believe he's trying to round it to the largest possible tens place, so the former would be preferred.

tests/performance/test_performance.py
```diff
@@ -54,15 +63,177 @@ def test_performance(config_path):
...
+    Returns:
+        pd.Series:
+            Candidate values for each metric.
+    """
+    results = [
+        profile_transformer(transformer, dataset_generator, transform_size, fit_size)
+        for _ in range(iterations)
+    ]
+    means = pd.DataFrame(results).mean(axis=0)
+    return (means * 10).apply(_round_to_magnitude)
```
amontanez24, line 128: Now that I think about it, we may be able to reduce the multiplier to something more like 5.

tests/performance/test_performance.py
sarahmish: That makes sense, I forgot about the multiplication by 10 part.

tests/performance/test_performance.py
csala: In general I tend to use `print` statements for quick command line interfaces or simple user-facing functions like this one which are likely to be executed without any logging setup. For any function that is likely to be integrated withing other code or command line interfaces which are complex enough to offer verbosity or output file controls, I use logging.

tests/performance/test_performance.py
csala: I would not expect it to happen either, but if it happens I would assume that getting a `ValueError` would be the proper behavior, since if the `value` is 0 we cannot really set any reasonable upper bound following the current logic.
I'll update the function to raise the proper exception message if that's the case, and if it ever becomes a blocker in the future we can always revisit it.

tests/performance/test_performance.py
csala: Good catch

tests/performance/test_performance.py
csala: Ok, I agree. And even more: I'll actually make it an argument (default = 5), so we have a way to decide decide how aggressive the tests are.

tests/performance/tests/datasets/test_utils.py
```diff
@@ -23,7 +23,7 @@ def test_add_nulls_float():
     with_nans = utils.add_nans(array)
 
     assert len(with_nans) == 100
-    assert 1 < np.isnan(with_nans).sum() < 100
+    assert 1 <= np.isnan(with_nans).sum() < 100
```
amontanez24, line 26: good catch

amontanez24 (approved):
LGTM!

PR #201, opened 2021-07-24
Add categorical generators

sarahmish:
1. add categorical generators and tests for LabelEncoder and Categorical transformers.
2. small tweak for LabelEncoder to allow input of type `np.ndarray`


tests/performance/datasets/categorical.py
```diff
@@ -0,0 +1,136 @@
...
+import numpy as np
+
+from tests.performance.datasets.base import BaseDatasetGenerator
+from tests.performance.datasets.utils import add_nans
+
+
+class RandomIntegerGenerator(BaseDatasetGenerator):
+    """Generator that creates an array of random integers."""
+
+    TYPE = 'numerical'
```
csala, line 12: Would this (and the type for the others) not be `categorical` instead?

tests/performance/datasets/categorical.py
```diff
@@ -0,0 +1,136 @@
+"""Dataset Generators for categorical transformers."""
```
csala, line 1: Should we add a dataset that produces a mixture of types?
We should also think about adding other types that can be used as categories, like `datetimes`, `floats` or `bools`.
Maybe it is not necessary to add explicitly datasets for each possible type, but at least have one dataset that produces a mixture of all of them, as well as NaNs and Nones.

csala (changes requested):

csala (approved):
I think this looks great @sarahmish !
Any thoughts @amontanez24 ?

amontanez24 (approved):
I think this is good to merge!

PR #202, opened 2021-07-26
Generate boolean transformer test cases

katxiao:


csala (changes requested):
If it does not take too long to run, I would also add tests for 100K rows.
Also I'm curious about the name used for the non-default tests set: why is the prefix `auto`? Maybe we should take something more meaningful wrt the arguments being used, like `nan_null`

katxiao:
> If it does not take too long to run, I would also add tests for 100K rows.
> Also I'm curious about the name used for the non-default tests set: why is the prefix `auto`? Maybe we should take something more meaningful wrt the arguments being used, like `nan_null`

I see, I thought we were using `auto` to describe the default arguments, but I can update it.

csala (approved):
good to go!

PR #213, opened 2021-08-13
Add readme for generating performance tests

katxiao:
Add a README for how to generate performance tests.  I think the example from the original [PR](https://github.com/sdv-dev/RDT/pull/200) is pretty useful, and I reference it to generate new tests.


csala (approved):
Thanks @katxiao ! Good addition

PR #215, opened 2021-08-19
Matching meta.yml and setup.py

fealho:
Since `psutil` was added to `setup.py`, it also needs to be added to `setup.py`.


amontanez24 (approved):
:shipit:

PR #224, opened 2021-09-20
Updates the baseclass

fealho:


rdt/transformers/base.py
```diff
@@ -9,20 +9,118 @@ class BaseTransformer:
...
+    def fit(self, data, columns):
         """Fit the transformer to the data.
 
         Args:
             data (pandas.Series or numpy.array):
                 Data to transform.
+            columns (list):
+                List of column names from the data.
+        """
+        self._columns = columns
```
csala, line 89: Here we should also craft a unique identifier of the columns by concatenating the names of the columns with a separator and storing it as `self._column_name`.

For example, if `columns == ['year', 'month', 'day']`, `self._column_name == 'year#month#day'`. We just need to make sure that the separator we use does not end up generating a column that already exists in the input table.

rdt/transformers/base.py
```diff
@@ -42,8 +140,22 @@ def fit_transform(self, data):
             numpy.array:
                 Transformed data.
         """
-        self.fit(data)
-        return self.transform(data)
+        columns_data = data[self._columns]
+        return self._fit_transform(columns_data)
```
csala, line 144: I think that `fit_transform` could stay the same as it was (so it just calls `fit` and `transform`, and `_fit_transform` is not needed), but the `columns` argument needs to be added and passed down to `fit`

rdt/transformers/base.py
```diff
@@ -42,8 +140,22 @@ def fit_transform(self, data):
...
+
+    def _fit_transform(self, columns_data):
```
csala, line 146: As mentioned above, this is possibly not needed.

rdt/transformers/base.py
```diff
@@ -52,6 +164,28 @@ def reverse_transform(self, data):
...
+            pandas.Series:
+                Reverted data.
+        """
+        output_types = self.get_output_types()
+        output_columns = list(output_types.keys())
+        columns_data = data[output_columns]
+        reversed_columns = self._reverse_transform(columns_data)
+        data[self._columns] = reversed_columns
+
+        columns_to_drop = set(output_columns) - set(self._columns)
```
csala, line 177: This could possibly be captured inside the `fit` method, after calling `self._fit`, since at that time the output names are already known.

csala (changes requested):

rdt/transformers/base.py
amontanez24: If there is only one column, should the separator still be appended?

rdt/transformers/base.py
```diff
@@ -52,6 +164,28 @@ def reverse_transform(self, data):
...
 
+        Returns:
+            pandas.Series:
+                Reverted data.
+        """
+        output_types = self.get_output_types()
+        output_columns = list(output_types.keys())
+        columns_data = data[output_columns]
+        reversed_columns = self._reverse_transform(columns_data)
+        data[self._columns] = reversed_columns
```
amontanez24, line 175: I'm not sure what our preference stylistically is, but these lines can probably be reduced by saying something like
```
columns_data = data[list(self.get_output_types())]
data[self._columns] = self._reverse_transform(columns_data)
```

amontanez24 (commented):
This is looking good so far. Just minor comments

rdt/transformers/base.py
```diff
@@ -9,20 +9,118 @@ class BaseTransformer:
...
+        return self.INPUT_TYPE
+
+    def get_output_types(self):
+        """Return the output types supported by the transformer.
+
+        Returns:
+            dict:
+                Mapping from the transformed column names to supported data types.
+        """
+        return self.OUTPUT_TYPES
```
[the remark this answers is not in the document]
fealho: I implemented it, but I think this has a few implications:

1.  First, this means we have to append something at the end of the column name always, even if only one column was generated, instead of not appending anything.
2. Currently, since we apply the null_transformer at the end of most transformers, this logic won't be used, since we have to check if the data has `None`'s, which means we have to overwrite this function.

rdt/transformers/base.py
fealho: 1) If we go with @csala suggestion to regarding the `get_output_types()`, I think we will always have to append something, yes.

2) Those are the lines @csala wrote in our meeting Friday, which is why I left them like that. I could simplify things a bit though.

rdt/transformers/base.py
fealho: I don't understand the reason for this change? It makes sense for the logic of figuring out which columns to drop to be in the reverse_transform, since that's where we drop the columns. Since we only drop the columns in the reverse_transform, it seems unnecessary to create an extra attribute in a different function, since computing the `columns_to_drop` is trivial.

rdt/transformers/base.py
```diff
@@ -9,20 +9,121 @@ class BaseTransformer:
...
 
-    def fit(self, data):
+    INPUT_TYPE = None
+    OUTPUT_TYPES = None
+    DETERMINISTIC_TRANSFORM = None
+    DETERMINISTIC_REVERSE = None
+    COMPOSITION_IS_IDENTITY = None
+    NEXT_TRANSFORMERS = None
+
+    def get_input_type(self):
```
amontanez24, line 19: Can we make this a static or class method? This will be accessed by the `HyperTransformer` to figure out which transformers to use for a data type

fealho: Sounds reasonable.

rdt/transformers/base.py
fealho: If we go with @csala suggestion regarding the `get_output_types()`, I think we will always have to append something, yes.

rdt/transformers/base.py
```diff
@@ -9,20 +9,122 @@ class BaseTransformer:
...
+        if self.NEXT_TRANSFORMERS:
+            return {f'{self._column_prefix}.{k}': v for k, v in self.NEXT_TRANSFORMERS.items()}
+
+        return None
+
+    def fit(self, data, columns):
         """Fit the transformer to the data.
 
         Args:
             data (pandas.Series or numpy.array):
```
fealho, line 84: Do we still want to support `np.array`? It makes more sense to me for the `fit` to only accept `pd.DataFrames`, seeing as we are passing the entire table.  @csala @amontanez24

rdt/transformers/base.py
```diff
@@ -9,20 +9,122 @@ class BaseTransformer:
...
+            self._column_prefix.append('#')
+
+        columns_data = data[columns]
+        self._fit(columns_data)
+
+    def _fit(self, columns_data):
+        """Fit the transformer to the data.
+
+        Args:
+            columns_data (pandas.Series or numpy.array):
```
fealho, line 101: Change this if `fit` only takes `dataframes`.

csala (changes requested):

rdt/transformers/base.py
csala: Good suggestion @amontanez24 Yes, I agree that some lines can be merged here!

rdt/transformers/base.py
fealho: Implemented it. My only concern is that with each new features the code is becoming more complex, so I'll have to do a good job of documenting.

rdt/transformers/base.py
```diff
@@ -1,28 +1,172 @@
...
+
+        self._column_prefix = '#'.join(columns)
+        self._output_columns = list(self.get_output_types().keys())
+
+        # make sure none of the generated `output_columns` exists in the data
+        while any(output_column in data for output_column in self._output_columns):
+            self._column_prefix += '#'
+            self._output_columns = list(self.get_output_types().keys())
+
+        self._columns_to_drop = set(self._output_columns) - set(columns)
```
csala, line 118: Thinking more about this, I now realize that `self._columns_to_drop` is not really needed, because we already know that `columns` and `self._output_columns` will always be disjoint sets.
In other words, we just need to drop `self._columns` in `transform` and `self._output_columns` in `reverse_transform`

rdt/transformers/base.py
```diff
@@ -31,25 +175,59 @@ def transform(self, data):
...
+        if len(self._output_columns) == 1:
+            output_columns = self._output_columns[0]
+
+        # what about columns being partially in the data?
+        if any(column not in data.columns for column in self._columns):
+            return data
+
+        columns_data = data[output_columns]
+        data[columns] = self._reverse_transform(columns_data)
+        data.drop(list(self._columns_to_drop), axis=1)
```
csala, line 222: Here we should be dropping `self._output_columns`: `data.drop(self._output_columns, axis=1)`

rdt/transformers/base.py
```diff
@@ -1,28 +1,172 @@
...
+
+        # what about columns being partially in the data?
+        if any(column not in data.columns for column in self._columns):
+            return data
+
+        columns_data = data[columns]
+        transformed_data = self._transform(columns_data)
+        data[output_columns] = transformed_data  # need to make sure columns properly ordered
+
+        return data
```
csala, line 163: We should `drop(self._columns)` before returning.

rdt/transformers/base.py
```diff
@@ -1,28 +1,172 @@
...
+        return output
+
+    def get_output_types(self):
+        """Return the output types supported by the transformer.
+
+        Returns:
+            dict:
+                Mapping from the transformed column names to supported data types.
+        """
+        return self._add_prefix(self._OUTPUT_TYPES, self._column_prefix)
```
csala, line 49: `_OUTPUT_TYPES` -> `OUTPUT_TYPES`

rdt/transformers/base.py
```diff
@@ -1,28 +1,172 @@
...
+        return self.COMPOSITION_IS_IDENTITY
+
+    def get_next_transformers(self):
+        """Return the suggested next transformer to be used for each column.
+
+        Returns:
+            dict:
+                Mapping from transformed column names to the transformers to apply to each column.
+        """
+        return self._add_prefix(self._NEXT_TRANSFORMERS, self._column_prefix)
```
csala, line 85: `NEXT_TRANSFORMERS`

rdt/transformers/base.py
```diff
@@ -1,28 +1,172 @@
...
+        """Fit the transformer to the `columns` of the `data`.
+
+        Args:
+            data (pandas.DataFrame):
+                The entire table.
+            columns (list or str):
+                List or tuple of column names from the data to transform.
+                If only one column is provided, it can be passed as a string instead.
+        """
+        if columns is None:
```
csala, line 97: Thinking more about this, making `columns` optional does not seem necessary except for the backwards compatibility with the HyperTransformer, which is going to be changed immediately, so we can just make `columns` mandatory.

rdt/transformers/base.py
```diff
@@ -31,25 +175,59 @@ def transform(self, data):
...
         Args:
-            data (pandas.Series or numpy.array):
+            data (pandas.DataFrame):
+                The entire table.
+
+        Returns:
+            pandas.DataFrame:
+                The entire table, containing the reverted data.
+        """
+        if isinstance(data, pd.Series):
```
csala, line 207: As mentioned above, this can probably be dropped.

csala (changes requested):

csala (approved):
I think this looks good now, except for the Github Actions failures.
I just opened a PR towards this branch that adapts the affected tests and fixes the PyLint configuration to avoid unnecessary warnings. Once that is merged, this can go in IMO.

rdt/transformers/base.py
```diff
@@ -1,55 +1,229 @@
...
+        columns_data = data[columns]
+        transformed_data = self._transform(columns_data)
+
+        output_columns = self._convert_if_series(self._output_columns, transformed_data)
+        data[output_columns] = transformed_data
+        data.drop(self._columns, axis=1, inplace=True)
+
+        return data
+
+    def _convert_if_series(self, columns, data):
```
amontanez24, line 157: can you move this before `transform` since `transform` uses it?

amontanez24 (approved):
One small comment but besides that looks good to go!

rdt/transformers/base.py
fealho: I prefer to write the helpers after the function that calls them. Do you know if writing them the other way around is the general guideline? Because then I'd have to change the order of all three helpers as well.

amontanez24: As far as I am concerned, everywhere in SDV lists the helpers before the methods that use them.

fealho: Huh, I hadn't noticed. I guess I can invert them then 👍

PR #227, opened 2021-09-20
Update numerical transformer

fealho:


rdt/transformers/numerical.py
```diff
@@ -91,16 +124,13 @@ def _learn_rounding_digits(data):
...
-    def fit(self, data):
+    def _fit(self, data):
         """Fit the transformer to the data.
 
         Args:
-            data (pandas.Series or numpy.ndarray):
+            data (pandas.DataFrame or pandas.Series):
                 Data to fit.
         """
-        if isinstance(data, np.ndarray):
```
csala, line 101: Why is this removed? Is it simply not necessary anymore, or does keeping it make the code crash?

csala (approved):
I think this looks good to go.
I would only mention the same think I added on the other PR, about using longer column names on the tests, just in case.

rdt/transformers/numerical.py
fealho: I think it was just not necessary.

rdt/transformers/numerical.py
```diff
@@ -70,11 +75,39 @@ def __init__(self, dtype=None, nan='mean', null_column=None, rounding=None,
...
+
+        return self.COMPOSITION_IS_IDENTITY
+
     @staticmethod
     def _learn_rounding_digits(data):
         # check if data has any decimals
+        data = np.array(data)
         roundable_data = data[~(np.isinf(data) | pd.isnull(data))]
-        if (roundable_data % 1 != 0).any():
+        if ((roundable_data % 1) != 0).any():
```
amontanez24, line 110: Was this a bug?

setup.cfg
```diff
@@ -60,4 +60,4 @@ max-attributes = 11
 ignore-comments = yes
 ignore-docstrings = yes
 ignore-imports = yes
-disable = R0903, R0913, R0914, C0209, W0223, W0221, W0237
+disable = R0801 R0903, R0913, R0914, C0209, W0223, W0221, W0237
```
amontanez24, line 63: What is this for?

amontanez24 (approved):
Couple questions, but besides that looks good

setup.cfg
csala: This disables the warning about duplicate code that PyLint raises because the  `get_output_types` functions are very similar across multiple transformers.
This is temporary and will be removed in a near-future update when the NullTransformer is decoupled as an independent thing.
@fealho The syntax is wrong, though, because the warnings need to be separated by comma (look at the others) and when the comma is missing PyLint does not take it. I'm correcting the syntax myself now

rdt/transformers/numerical.py
csala: It seems to actually be necessary, since this is what makes the performance tests fail: https://github.com/sdv-dev/RDT/runs/3719384829?check_suite_focus=true#step:5:6430

rdt/transformers/numerical.py
fealho: Nope, I just think it's more readable this way.

PR #231, opened 2021-09-20
Update categorical transformers

fealho:


tests/integration/test_hyper_transformer.py
```diff
@@ -228,7 +228,7 @@ def test_dtype_category():
 
     rever = ht.reverse_transform(trans)
 
-    pd.testing.assert_frame_equal(df, rever)
```
amontanez24, line 231: What is this change for? Also, I know it's not your code but personally I prefer to write out the full names unless it's something super long. You can leave that though

amontanez24 (approved):
This lgtm!

tests/integration/test_hyper_transformer.py
fealho: Oh, I just forgot to revert back. I agree that some of the tests need to be standardized.

rdt/transformers/categorical.py
```diff
@@ -33,6 +33,12 @@ class CategoricalTransformer(BaseTransformer):
             Defaults to ``False``.
     """
 
+    INPUT_TYPE = 'categorical'
+    OUTPUT_TYPES = {'value': 'float'}
+    DETERMINISTIC_TRANSFORM = False
```
csala, line 38: This should be made dynamically, since this depends on whether `fuzzy==True` or not.

rdt/transformers/categorical.py
```diff
@@ -145,9 +151,9 @@ def _get_value(self, category):
 
     def _transform_by_row(self, data):
         """Transform the data row by row."""
-        return data.fillna(np.nan).apply(self._get_value).to_numpy()
+        return data.fillna(np.nan).apply(self._get_value)
```
csala, line 154: I would try to avoid changing the output type, since the other transform option (`by_category`) does not change it either.
Do things fail if this is not removed?

rdt/hyper_transformer.py
```diff
@@ -260,7 +260,7 @@ def reverse_transform(self, data):
                     # temporarily support both old and new style transformers
                     rename = dict(zip(columns, transformer.output_columns))
                     reversed_data = transformer.reverse_transform(data.rename(columns=rename))
-                    reversed_data = reversed_data[transformer.columns[0]].to_numpy()
+                    reversed_data = reversed_data[transformer.columns[0]]
```
csala, line 263: Is this change needed? Not a big issue, since this will be deprecated really soon, but needing this change may mean other things are wrong.

csala (changes requested):
Overall it looks good, but I left a couple of comments that may be relevant.

rdt/transformers/categorical.py
fealho: Yep, they crash.

rdt/hyper_transformer.py
fealho: Yeah, it crashes without it.

rdt/transformers/categorical.py
csala: I dug a bit more into it. I do not think this change and the one below are necessary if we simply take care of converting the transformed data to `numpy` right before indexing it, in line 197 of the HyperTransformer.

You can see an example of this being done in this other branch: https://github.com/sdv-dev/RDT/blob/7d8c8cfb317fa99c91a8e5170441ff96172f7f43/rdt/hyper_transformer.py#L204

I would rather apply this change to fix the errors and continue to produce a consistent output format (numpy) in all the internal transformer methods.

tests/integration/test_hyper_transformer.py
```diff
@@ -228,7 +228,7 @@ def test_dtype_category():
```
csala, line 228: It seems like the `test_dtype_category` above this line fails because the `dtype` is lost during the transform/reverse cycle.
A similar inconsistency was reported for the Boolean Transformer, so I'm opening an issue to take care of this in the `BaseTransformer`.

@fealho can you flag this as an [expected failure using pytest](https://docs.pytest.org/en/latest/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail) in the meantime?

csala (approved):
This seems ready to go! :ship:

Issue #235, opened 2021-09-21
Move datasets to its own module


katxiao:
### Problem Description

The dataset generators are currently nested under performance tests. We should move it to its own module so that it can be reused for other automated tests.

csala:
Closed via #236

PR #236, opened 2021-09-21
Move dataset generators

katxiao:
Move dataset generators to its own module, so that they can be reused for other automated tests.

- Update dataset generators to inherit from a dataset generator of the relevant data type
- Update the dataset imports (in performance tests) to point to new location

Resolve issue #235


tasks.py
```diff
@@ -10,7 +10,7 @@
 @task
 def pytest(c):
     c.run(
-        'pytest ./tests/unit ./tests/integration ./tests/performance/tests '
+        'pytest ./tests/unit ./tests/integration ./tests/performance/tests ./tests/datasets/tests '
```
amontanez24, line 13: The tests here are just tests for the generators themselves correct?

tests/datasets/__init__.py
```diff
@@ -1,10 +1,11 @@
 """Dataset Generators to test the RDT Transformers."""
 
-from tests.performance.datasets import boolean, datetime, numerical
-from tests.performance.datasets.base import BaseDatasetGenerator
+from tests.datasets import boolean, categorical, datetime, numerical
```
amontanez24, line 3: good catch

amontanez24 (approved):
LGTM!

tasks.py
csala: That's right

csala (approved):
LGTM!

PR #237, opened 2021-09-21
Add automated transformer integration test

### LINKED ISSUE #223: Automated integration test for transformers
### Problem Description

To validate transformers end-to-end, there are certain behaviors that we expect from all transformers. We can automate these checks, so that they run for every transformer by default.

### Expected behavior

Automatically run end-to-end tests for each transformer, using the generated datasets that already exist for performance testing. These integration tests should validate that

1. The Transformer is correct on its own
    - The transformed data is the expected data type(s)
    - The reverse transformed data is the expected data type
    - The reverse transformed data is the same as the original data (for the transformers for which composition is identity)
    - The Transformer can be used by the HyperTransformer
   
2. If we specify the transformer in the hypertransformer, the hypertransformer can go from the specified input type to a numerical output, and back.

katxiao:
Add automated transformer integration test

Resolves #223

### DIFF
--- a/tests/datasets/base.py
+++ b/tests/datasets/base.py
@@ -20,3 +20,20 @@ class BaseDatasetGenerator(ABC):
             numpy.ndarray of size ``num_rows``
         """
         raise NotImplementedError()
+
+    @classmethod
+    def get_subclasses(cls):
+        """Recursively find subclasses of this Baseline.
+
+        Returns:
+            list:
+                List of all subclasses of this class.
+        """
+        subclasses = []
+        for subclass in cls.__subclasses__():
+            if ABC not in subclass.__bases__:
+                subclasses.append(subclass)
+
+            subclasses += subclass.get_subclasses()
+
+        return subclasses
--- a/tests/integration/test_transformers.py
+++ b/tests/integration/test_transformers.py
@@ -1,0 +1,92 @@
+import numpy as np
+import pandas as pd
+import pytest
+
+import rdt.transformers
+import tests.datasets
+
+DATA_SIZES = [100, 1000, 10000]
+
+# Temporary mapping
+TRANSFORMER_TO_TYPE = {
+    rdt.transformers.null.NullTransformer: None,
+    rdt.transformers.boolean.BooleanTransformer: 'boolean',
+    rdt.transformers.categorical.CategoricalTransformer: 'categorical',
+    rdt.transformers.categorical.OneHotEncodingTransformer: 'categorical',
+    rdt.transformers.categorical.LabelEncodingTransformer: 'categorical',
+    rdt.transformers.datetime.DatetimeTransformer: 'datetime',
+    rdt.transformers.numerical.NumericalTransformer: 'numerical',
+    rdt.transformers.numerical.GaussianCopulaTransformer: 'numerical',
+}
+
+
+def get_subclasses(obj):
+    subclasses = obj.__subclasses__()
+
+    if len(subclasses) == 0:
+        return []
+
+    for sc in subclasses:
+        subclasses.extend(get_subclasses(sc))
+
+    return subclasses
+
+
+def build_generator_map():
+    generators = {}
+
+    for g in tests.datasets.BaseDatasetGenerator.__subclasses__():
+        generators[g.DATA_TYPE] = g.__subclasses__()
+
+    return generators
+
+
+def get_transformer_test_cases():
+    test_cases = []
+
+    transformers = get_subclasses(rdt.transformers.BaseTransformer)
+    generators = build_generator_map()
+
+    for t in transformers:
+        for ds in DATA_SIZES:
+            dt = TRANSFORMER_TO_TYPE[t]
+
+            if dt is None:
+                continue
+
+            gs = generators[dt]
+
+            for g in gs:
+                test_cases.append((t, ds, g))
+
+    return test_cases
+
+
+test_cases = get_transformer_test_cases()
+
+
+@pytest.mark.parametrize('transformer,data_size,dataset_generator', test_cases)
+def test_transformer(transformer, data_size, dataset_generator):
+    # Generate input data
+    input_data = pd.Series(dataset_generator.generate(data_size))
+
+    # Fit
+    t = transformer()
+    t.fit(input_data)
+
+    # Transform
+    transformed = t.transform(input_data)
+    # TODO: Verify transformed data type
+
+    # Reverse transform
+    out = t.reverse_transform(transformed)
+    # TODO: Verify out data type
+
+    if isinstance(out, pd.DatetimeIndex):
+        out = out.to_series(index=pd.RangeIndex(start=0, stop=data_size, step=1))
+    if pd.api.types.is_datetime64_any_dtype(out):
+        out = out.round('us')
+    if isinstance(out, np.ndarray):
+        out = pd.Series(out)
+
+    pd.testing.assert_series_equal(out, input_data, check_dtype=False)
--- a/tests/integration/test_transformers.py
+++ b/tests/integration/test_transformers.py
@@ -2,9 +2,9 @@ import numpy as np
 import pandas as pd
 import pytest
 
+from rdt import HyperTransformer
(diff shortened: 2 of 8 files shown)


tests/integration/test_transformers.py
```diff
@@ -0,0 +1,92 @@
...
+                test_cases.append((t, ds, g))
+
+    return test_cases
+
+
+test_cases = get_transformer_test_cases()
+
+
+@pytest.mark.parametrize('transformer,data_size,dataset_generator', test_cases)
+def test_transformer(transformer, data_size, dataset_generator):
```
csala, line 69: I added a comment about this directly in the issue, so we can discuss there in a more organized way.

tests/integration/test_transformers.py
```diff
@@ -0,0 +1,188 @@
...
+
+    for g in BaseDatasetGenerator.__subclasses__():
+        generators[g.DATA_TYPE] = g.__subclasses__()
+
+    return generators
+
+
+def _find_dataset_generators(data_type, generators):
+    """Find the dataset generators for the given data_type."""
+    if data_type is None or data_type not in generators:
```
amontanez24, line 41: could we do something like 
```python3
return generators.get(data_type, [])
```

tests/integration/test_transformers.py
```diff
@@ -0,0 +1,188 @@
...
+    if isinstance(out, pd.DataFrame):
+        reversed_data = out[TEST_COL]
+    if isinstance(out, pd.DatetimeIndex):
+        reversed_data = out.to_series(index=pd.RangeIndex(start=0, stop=DATA_SIZE, step=1))
+    if isinstance(reversed_data, np.ndarray):
+        out = pd.Series(reversed_data)
+    if pd.api.types.is_datetime64_any_dtype(reversed_data):
+        reversed_data = reversed_data.round('us')
+
+    _validate_composition(transformer, reversed_data, input_data)
```
amontanez24, line 121: If I'm reading this correctly, all these checks only matter if `transformer.is_composition_identity() == True`. Should we move this stuff to after that check inside of `_validate_composition`? Or will it be used for other checks later?

amontanez24 (commented):
Looking good, just some small comments

tests/integration/test_transformers.py
```diff
@@ -0,0 +1,188 @@
...
+TEST_COL = 'test_col'
+
+# Mapping of rdt data type to dtype
+DATA_TYPE_TO_DTYPES = {
+    'boolean': ['b', 'O'],
+    'categorical': ['O', 'i', 'f'],
+    'datetime': ['M'],
+    'numerical': ['f', 'i'],
+    'integer': ['i'],
+    'float': ['f', 'i'],
```
katxiao, line 19: Right now, `OneHotEncodingTransformer` sometimes returns ints, but has an output type of `float`. I included `i` in this list to get the tests to pass. Should we update the output type of the transformer to `numerical`?

tests/integration/test_transformers.py
katxiao: Yes we can move it into `_validate_composition` 👍

tests/integration/test_transformers.py
```diff
@@ -0,0 +1,188 @@
...
+    assert input_data_type is not None
+
+    dataset_generators = _find_dataset_generators(input_data_type, generators)
+    assert len(dataset_generators) > 0
+
+    for dg in dataset_generators:
+        with subtests.test(msg="test_transformer_with_dataset", generator=dg):
+            input_data = pd.Series(dg.generate(DATA_SIZE))
+
+            data = pd.DataFrame({TEST_COL: input_data})
```
csala, line 186: I think we could replace `input_data` with `dg.generate(DATA_SIZE)` and skip the line above.

tests/integration/test_transformers.py
```diff
@@ -0,0 +1,188 @@
...
+
+    Test the transformer end-to-end with at least one generated dataset. Test
+    both the transformer by itself, and by running in the hypertransformer.
+
+    Args:
+        transformer (rdt.transformers.BaseTransformer):
+            The transformer to test.
+    """
+    input_data_type = transformer.get_input_type()
+    assert input_data_type is not None
```
csala, line 177: I think we can skip this validation, and in the `_find_dataset_generators` function below return all the dataset generators that exist.

tests/integration/test_transformers.py
```diff
@@ -0,0 +1,188 @@
...
+    # Reverse transform
+    out = transformer.reverse_transform(transformed)
+    _validate_reverse_transformed_data(transformer, out, input_data.dtypes[TEST_COL])
+
+    _validate_composition(transformer, out, input_data)
+
+
+def _validate_hypertransformer_transformed_data(transformed_data):
+    """Check that the transformed data is of type float."""
+    for dt in transformed_data.dtypes:
```
csala, line 126: nit: `dtype`

tests/integration/test_transformers.py
```diff
@@ -0,0 +1,188 @@
...
+        assert dt.kind in DATA_TYPE_TO_DTYPES['numerical']
```
csala, line 127: I think we should also validate that there are no null values

tests/integration/test_transformers.py
```diff
@@ -0,0 +1,188 @@
...
+
+    This is only applicable if the transformer has the composition
+    identity property.
+    """
+    if not transformer.is_composition_identity():
+        return
+
+    if isinstance(reversed_data, pd.DataFrame):
+        reversed_data = reversed_data[TEST_COL]
+    if isinstance(reversed_data, pd.DatetimeIndex):
```
csala, line 79: I think these should be `elif`

csala (changes requested):
I added a few comments, but in general this is looking great @katxiao !
I think that a few other changes will be required after PR #240 is merged too.

tests/integration/test_transformers.py
katxiao: Multiple of these conditions could apply - for example, if the reversed_data is a DataFrame and a datetime. I could `elif` the two `isinstance` checks.

tests/integration/test_transformers.py
katxiao: Opened https://github.com/sdv-dev/RDT/issues/268

tests/integration/test_transformers.py
```diff
@@ -0,0 +1,188 @@
...
+    """Build a map of data type to data generator.
+
+    Output:
+        dict:
+            A mapping of data type (str) to a list of data
+            generators (rdt.tests.datasets.BaseDatasetGenerator).
+    """
+    generators = {}
+
+    for g in BaseDatasetGenerator.__subclasses__():
```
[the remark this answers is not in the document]
katxiao: I see. Right now we have subclasses like `NumericalGenerator` (that other numerical generators subclass), which don't implement the `generate` method. I agree that we should allow for flexibility for subclasses to have grand-children. I think I will just have to catch the `NotImplementedError` for the case where `generate` is not implemented.

csala (approved):
Let's go!

csala (approved):
Re-approving this, and also re-requesting review from @amontanez24 
@amontanez24 @katxiao This and #240 will clash, because the HyperTransformer API changes a bit. I suggest we merge #240 first, then update this one, and then merge

tests/datasets/base.py
```diff
@@ -20,3 +20,20 @@ def generate(num_rows):
             numpy.ndarray of size ``num_rows``
         """
         raise NotImplementedError()
+
+    @classmethod
+    def get_subclasses(cls):
```
amontanez24, line 25: I notice we do this in a few places. Not really necessary for this PR, but maybe we should make this a utility function somewhere that can take any class and get all the subclasses

amontanez24 (approved):
LGTM!

tests/datasets/base.py
katxiao: Yeah that's what I had originally (put in a utils file), but I think @csala preferred it to be on the transformer. We can revisit this if it comes up again.

PR #239, opened 2021-09-21
Create ways for HyperTransformer to know which transformers to apply to each data type #232

### LINKED ISSUE #232: Create ways for HyperTransformer to know which transformers to apply to each data type
### Problem Description

Since RDT will be expanding the number of data types it supports, we will need a new way for the `HyperTransformer` to determine which transformers to apply to a specific data type. We propose the following additions:

1. A function, `get_transformers_by_type`, which iterates over all the existing Transformers and calls the `get_input_type` method to create a dictionary that indicates which transformers work with which data type.
2. A `DEFAULT_TRANSFORMERS` dict which may indicate what the default transformer for each data type is.
3. If a data type is not in the `DEFAULT_TRANSFORMERS` dictionary, the first transformer found is "the default".

amontanez24:
resolves #232 

This PR adds the following attributes to the `HyperTransformer`

1. `_DTYPES_TO_DATA_TYPES` - a dict mapping the pandas dtypes to an RDT data type
2. `DEFAULT_TRANSFORMERS` - a dict mapping the RDT data types to a default transformer

It also adds the following static method
1. `get_transformers_by_type` - a function that loops through all existing transformers and creates a dict mapping data types to a list of valid transformers for that type.

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -1,6 +1,8 @@
 """Hyper transformer module."""
 
+import inspect
 import re
+import sys
 from copy import deepcopy
 
 import numpy as np
@@ -73,6 +75,19 @@ class HyperTransformer:
         'b': 'boolean',
         'M': 'datetime',
     }
+
+    @staticmethod
+    def get_transformers_by_type():
+        data_type_transformers = dict()
+        transformer_classes = inspect.getmembers(sys.modules[__name__], inspect.isclass)
+        for (_, transformer) in transformer_classes:
+            try:
+                input_type = transformer.get_input_type()
+                transformers_for_type = data_type_transformers.get(input_type, [])
+                transformers_for_type.append(transformer)
+                data_type_transformers.update({input_type, transformers_for_type})
+            except:
+                pass
 
     def __init__(self, transformers=None, copy=True, dtypes=None, dtype_transformers=None):
         self.transformers = transformers
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -93,7 +93,14 @@ class HyperTransformer:
 
     @staticmethod
     def get_transformers_by_type():
-        data_type_transformers = dict()
+        """Build a ``dict`` mapping data types to valid existing transformers for that type.
+
+        Returns:
+                dict:
+                    Mapping of data types to a list of existing transformers that take that
+                    type as an input.
+        """
+        data_type_transformers = {}
         transformer_classes = inspect.getmembers(sys.modules[__name__], inspect.isclass)
         for (_, transformer) in transformer_classes:
             try:
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -75,6 +75,21 @@ class HyperTransformer:
         'b': 'boolean',
         'M': 'datetime',
     }
+    DEFAULT_TRANSFORMERS = {
+        'numerical': NumericalTransformer,
+        'integer': NumericalTransformer(dtype=int),
+        'float': NumericalTransformer(dtype=float),
+        'categorical': CategoricalTransformer(fuzzy=True),
+        'boolean': BooleanTransformer,
+        'datetime': DatetimeTransformer,
+    }
+    _DTYPES_TO_DATA_TYPES = {
+        'i': 'integer',
+        'f': 'float',
+        'O': 'categorical',
+        'b': 'boolean',
+        'M': 'datetime',
+    }
 
     @staticmethod
     def get_transformers_by_type():
@@ -85,9 +100,10 @@ class HyperTransformer:
                 input_type = transformer.get_input_type()
                 transformers_for_type = data_type_transformers.get(input_type, [])
                 transformers_for_type.append(transformer)
-                data_type_transformers.update({input_type, transformers_for_type})
-            except:
+                data_type_transformers.update({input_type: transformers_for_type})
+            except AttributeError:
                 pass
+        return data_type_transformers
 
     def __init__(self, transformers=None, copy=True, dtypes=None, dtype_transformers=None):
         self.transformers = transformers
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -73,14 +73,6 @@ class HyperTransformer:
         'b': 'boolean',
         'M': 'datetime',
     }
-    DEFAULT_TRANSFORMERS = {
-        'numerical': NumericalTransformer,
-        'integer': NumericalTransformer(dtype=int),
-        'float': NumericalTransformer(dtype=float),
-        'categorical': CategoricalTransformer(fuzzy=True),
-        'boolean': BooleanTransformer,
-        'datetime': DatetimeTransformer,
-    }
     _DTYPES_TO_DATA_TYPES = {
         'i': 'integer',
         'f': 'float',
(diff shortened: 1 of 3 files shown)


rdt/hyper_transformer.py
```diff
@@ -73,6 +75,42 @@ class HyperTransformer:
...
+    _DTYPES_TO_DATA_TYPES = {
+        'i': 'integer',
+        'f': 'float',
+        'O': 'categorical',
+        'b': 'boolean',
+        'M': 'datetime',
+    }
+
+    @staticmethod
+    def get_transformers_by_type():
```
csala, line 95: I think it would make more sense to put this as a function inside the `transformers` module, either directly in the `__init__` module or in a `utils` module and then importing it from the `__init__`.

I imagine the use case where someone simply wants to explore what transformers exist, without using the `HyperTransformer`

```
>>> import rdt
>>> rdt.transformers.get_transformers_by_type()
...
```

rdt/hyper_transformer.py
```diff
@@ -73,6 +75,42 @@ class HyperTransformer:
...
+        """Build a ``dict`` mapping data types to valid existing transformers for that type.
+
+        Returns:
+                dict:
+                    Mapping of data types to a list of existing transformers that take that
+                    type as an input.
+        """
+        data_type_transformers = {}
+        transformer_classes = inspect.getmembers(sys.modules[__name__], inspect.isclass)
```
csala, line 104: Rather than using `inspect`, I would navigate the subclasses of `BaseTransformer` by adding a `get_subclasses` classmethod to it.

Something similar is done in SDGym: https://github.com/sdv-dev/SDGym/blob/79321b57a1b2dd416426fc9088e1ff771dd82e9c/sdgym/synthesizers/base.py#L17

Also, as a note, when implementing it here I would merge the `get_subclasses` and `get_baselines` logic, so only the classes that do not inherit from `ABC` directly are included in the output. By doing it this way, we can skip any intermediate abstract Transformers that may end up defining.

rdt/hyper_transformer.py
```diff
@@ -73,6 +75,42 @@ class HyperTransformer:
         'b': 'boolean',
         'M': 'datetime',
     }
+    DEFAULT_TRANSFORMERS = {
```
csala, line 78: I think this could also be moved over to the `transformers` module, and create an additional `get_default_transformer(data_type)` function that has the functionality of looking up the given `data_type` on this dictionary and if not found calls the `get_transformers_by_type` function to choose the first one that it finds that supports it.

rdt/hyper_transformer.py
```diff
@@ -73,6 +75,42 @@ class HyperTransformer:
...
+        'numerical': NumericalTransformer,
+        'integer': NumericalTransformer(dtype=int),
```
csala, line 80: Maybe we should create specific transformers for each configuration, so this dictionary can contain classes instead of instances? I'm opening a new issue to discuss this there.

csala (changes requested):
I added a couple of comments to suggest a different approach to this implementation.

rdt/hyper_transformer.py
amontanez24: I don't like the idea of calling `get_transformers_by_type` every time we don't have a defined transformer.

amontanez24: What I'll do is add a function that creates the default dict using the `DEFAULT_TRANSFORMERS` and `get_transformers_by_type` dicts, and we can just use that dict in the `HyperTransformer`

rdt/transformers/__init__.py
```diff
@@ -25,6 +25,14 @@
     transformer.__name__: transformer
     for transformer in BaseTransformer.__subclasses__()
```
katxiao, line 26: could call `BaseTransformers.get_subclasses()` here, so `TRANSFORMERS` is a list of all transformers.

katxiao: Also could we add `'GaussianCopulaTransformer` to the `__all__` list above?

rdt/transformers/__init__.py
```diff
@@ -77,3 +85,43 @@ def load_transformers(transformers):
...
+            dict:
+                Mapping of data types to a list of existing transformers that take that
+                type as an input.
+    """
+    data_type_transformers = {}
+    transformer_classes = BaseTransformer.get_subclasses()
+    for transformer in transformer_classes:
+        try:
+            input_type = transformer.get_input_type()
+            transformers_for_type = data_type_transformers.get(input_type, [])
```
csala, line 103: nit: We could use a `defaultdict(list)`:

```python3
data_type_transformers = defaultdict(list)
...
data_type_transformers[input_type].append(transformer)
```

Or `setdefault`:

```python3
data_type_transformers.setdefault(input_type, []).append(transformer)
```

katxiao (approved):

rdt/transformers/base.py
```diff
@@ -21,6 +23,22 @@ class BaseTransformer:
...
+
+        Returns:
+            list:
+                List of all subclasses of this class.
+        """
+        subclasses = []
+        for subclass in cls.__subclasses__():
+            if abc.ABC not in subclass.__bases__:
+                subclasses.append(subclass)
+            subclasses += subclass.get_subclasses()
```
csala, line 38: Can we add a blank line above this one?

csala (approved):
LGTM! I just added a couple of minor comments.

katxiao (changes requested):
Oops, meant to request changes.

rdt/hyper_transformer.py
csala: > What I'll do is add a function that creates the default dict using the `DEFAULT_TRANSFORMERS` and `get_transformers_by_type` dicts, and we can just use that dict in the `HyperTransformer`

I had not seen this. I agree, calling it every time sounds overkill, but somehow I do not like the idea of also calling the function when the HyperTransformer is imported, or every time an instance is created.

What do you think about using [functools.cache](https://docs.python.org/3/library/functools.html#functools.cache)?
We could actually have `get_default_transformers()` which caches its result the first time it is called, but then also have `get_default_transformer(transformer)`, which also caches its results, and that under the hood calls the other one and gets the transformer from the dict.

By doing this, the result is the behavior is the same (we end up having the dictionary stored somewhere instead of building it every time), but its initialization is lazy (the dict gets build on the fly the first time it is used)

rdt/transformers/__init__.py
amontanez24: It's already there

rdt/hyper_transformer.py
amontanez24: Makes sense. Let me push a commit and you can take a look

katxiao (approved):

rdt/transformers/__init__.py
```diff
@@ -1,5 +1,8 @@
 """Transformers module."""
 
+from collections import defaultdict
+from functools import lru_cache
```
csala, line 4: I'm not sure if `functools` is part of the standard library. If it is not, we should add it to `setup.py`, even if we already install it because one of our dependencies use it, so if they remove it in the future we do not crash.

csala (approved):
Looking great now! I just added a comment about functools dependency. We may need to do nothing about it, though

rdt/transformers/__init__.py
amontanez24: It is part of the standard library

PR #240, opened 2021-09-22
Update HyperTransformer

### LINKED ISSUE #233: Enable users to specify transformers to use in HyperTransformer
### Problem Description

The current `HyperTransformer` arguments could be expanded to also accept tuples of names instead of strings to indicate multiple fields that together represent a single data type. It also needs to be allow users to specify transformers for fields that are generated by other transformers.

### Expected behavior
The `HyperTransformer` should be able to take in the following parameters:

1. `field_types`: dictionary mapping single names or tuples of names to single data types
2. `data_type_transformers`: dictionary that maps from single data types to single transformers
3. `field_transformers`: Hierarchical way of specifying which transformer to apply to each field along all the process

**Initialization using no inputs:**
The simplest way to initialize a `HyperTransformer` is to use no inputs. If this is done, the `HyperTransformer` will infer which `transformers` to use for each column in the input data using the Pandas `dtype` of the column.

```python3
ht = HyperTransformer()
```
**Initialization using `field_types`:**
The `field_types` dictionary maps single columns or groups of columns, to their data type. This is especially useful if your data type does not easily map to a Pandas `dtype` or is more of a semantic type (ie. email). Defining these types for your fields will enable the `HyperTransformer` to select better transformers for them. If multiple columns are going to be an input for ONE data type, this can be specified by grouping the columns in a tuple.

In this example, your data has the columns: `company_email`, `home_phone`, `cell_phone`, `year_joined`, `month_joined` and `day_joined`.

```python3
ht = HyperTransformer(field_types={
    "company_email": "email",
    ("year", "month", "day"): "datetime",
    "home_phone": "phone_number",
    "cell_phone": "phone_number"
}
```

**Initialization using `data_type_transformers`:**
The `data_type_transformers` argument is useful if you want to use a specific `Transformer` or `Transformer` instance as the transformer for any field of a certain data type. Using this in conjunction with `field_types` allows a lot of control over `Transformer` specification.

In this example, we imagine there is a `AddressPrivateTransformer` that substitutes the street names in an address for privacy reasons.
```python3
ht = HyperTransformer(data_type_transformers={
    "categorical": CategoricalTransformer(fuzzy=True),
    "address": AddressPrivateTransformer,
    "ipv4": IPv4Transformer
})
```

**Initialization using `field_transformers`:**
The `field_transformers` are specified as a dictionary where the keys are the full name of the fields we want to transform and the values are the transformers. If a field is derived from another transformer, then its name is created by concatenating the field it came from to the output field name. Not all of the fields in your data, or their outputs need to be specified.

In this example, the data has a column named `email`.

```python3
ht = HyperTransformer(field_transformers={
    "email": EmailTransformer,
    "email.domain": EmailDomainTransformer,
    "email.username": CategoricalFuzzy,
    "email.domain.extension": OneHotEncoding,
})
```

amontanez24:
Resolves #233


rdt/hyper_transformer.py
```diff
@@ -80,73 +78,62 @@ class HyperTransformer:
...
         'M': 'datetime',
     }
-
-    def __init__(self, transformers=None, copy=True, dtypes=None, dtype_transformers=None):
-        self.transformers = transformers
-        self._transformers = {}
+    _DEFAULT_OUTPUT_TYPES = [
+        'numerical',
+        'float',
+        'int'
```
csala, line 84: The final name is not fully decided, but I suppose we will end up calling it `integer` rather than `int`.

rdt/hyper_transformer.py
```diff
@@ -80,73 +78,62 @@ class HyperTransformer:
...
+        self._transformers_sequence.append(transformer)
+        output_types = transformer.get_output_types()
+        next_transformers = transformer.get_next_transformers()
+        for (output, output_type) in output_types.items():
+            if output_type in self.transform_output_types:
+                self._output_columns.append(output)
+            else:
+                self._generated_columns.append(output)
+                transformed_data = transformer.transform(data)
+                if output not in self.field_transformers and next_transformers is not None:
```
csala, line 130: I think that before doing this we will need to see if the field is included in any multi-column `field_transformer` specification.
We could possibly store those in a separate `self._multi_field_transformers`, where only the fields that are specified as tuples are stored, and here iterate over those calling the `_field_in_data` defined above.

rdt/hyper_transformer.py
```diff
@@ -80,73 +78,62 @@ class HyperTransformer:
...
+                    self.field_transformers[output] = next_transformers[output]
+                if output in self.field_transformers:
+                    next_transformer = self.field_transformers[output]
+                else:
+                    next_transformer = self._default_transformers[output_type]
+                self._fit_field_transformer(transformed_data, output, next_transformer)
```
csala, line 136: Maybe we should do this after the `for` loop finishes, so multi-column field transformers can be applied on mutli-column outputs?

* Field `a` is passed to `TransformerA`, which produces output `{'a.b': 'a_data_type', 'a.c': 'numerical'}`
* `field_transformers` specifies `{('a.b', 'c'): TransformerABC}`, so this is applied

This should include the `next_transformers` also, so the following is also possible:

* Field `a` is pased to `TransformerA`, which produces outputs `{'a.b': 'a_data_type', 'a.c': 'another_data_type'}`
* `TransformerA.get_next_transformers()` returns `{('a.b', 'a.c'): TransformerBC}`, so this is applied.

And this actually makes me think, that the following should also be possible (we may want to add this one in a second iteration, though):

* Field `a` is pased to `TransformerA`, which produces outputs `{('a.b', 'a.c'): 'a_data_type'}`
* `data_type_transformers` specifies `{'a_data_type': TransformerABC}`, so this is applied.

rdt/hyper_transformer.py
```diff
@@ -155,14 +142,22 @@ def fit(self, data):
...
-            transformer.fit(column)
+        self._input_columns = data.columns
+        self._update_field_types(data)
+        fitted_fields = set()
+
+        for field in self.field_transformers:
+            if self._field_in_data(field, data):
+                self._fit_field_transformer(data, field, self.field_transformers[field])
+                fitted_fields.add(field)
+        for (field, data_type) in self.field_types.items():
```
csala, line 153: We could add some blank lines where the indentation is reduced, like above this line, or in line 160.

rdt/hyper_transformer.py
```diff
@@ -180,26 +175,12 @@ def transform(self, data):
...
-
-                else:
-                    data[column_name] = transformed
-
-        if drop_columns:
-            data.drop(drop_columns, axis=1, inplace=True)
+        for transformer in self._transformers_sequence:
+            data = transformer.transform(data, drop=False)
 
+        columns_to_drop = [col for col in data
```
csala, line 181: nit: I'd use hanging indent here, in three lines (output elements, for loop, condition), and also call it column

```python3
columns_to_drop = [
    column
    for column in data
    if column in self._input_columns or column in self._generated_columns
]

File: rdt/hyper_transformer.py [hunk 6]
```diff
@@ -80,73 +78,62 @@ class HyperTransformer:
...
+        transformer = load_transformer(transformer)
+        transformer.fit(data, field)
+        self._transformers_sequence.append(transformer)
+        output_types = transformer.get_output_types()
+        next_transformers = transformer.get_next_transformers()
+        for (output, output_type) in output_types.items():
+            if output_type in self.transform_output_types:
+                self._output_columns.append(output)
+            else:
+                self._generated_columns.append(output)
```
csala, line 128: Do we need this? I thought we only needed `_input_columns` and `_output_columns`

csala (changes requested):

rdt/hyper_transformer.py
amontanez24: My understanding is that if they transform a table with new columns, we don't want to do anything to those. However, how do we distinguish new columns from generated columns? Generated columns aren't in the input or output so I couldn't think of another way to filter them out

csala: Oh! Of course! You mean columns that were generated in an intermediate step, but which were further transformed by another transformer before finishing, right?

rdt/hyper_transformer.py
```diff
@@ -80,73 +78,87 @@ class HyperTransformer:
...
+                self.field_types[field] = self._DTYPES_TO_DATA_TYPES[data[field].dtype.kind]
+
+    def _fit_field_transformer(self, data, field, transformer):
+        transformer = load_transformer(transformer)
+        transformer.fit(data, field)
+        self._transformers_sequence.append(transformer)
+        output_types = transformer.get_output_types()
+        next_transformers = transformer.get_next_transformers()
+        for (output, output_type) in output_types.items():
+            if output in self.field_transformers:
```
csala, line 138: I think we could extract this if/else block into a `_get_next_transformer` method, so the logic in the loop becomes shorter:

```python3
for (output, output_type) in output_types.items():
    next_transformer = self._get_next_transformer(output, output_type, data)
    if next_transformer:
        self._temp_columns.append(output)
        transformed_data = transformer.transform(data)
        self._fit_field_transformer(transformed_data, output, next_transformer)
    else:
        self._output_columns.append(output)
```

rdt/hyper_transformer.py
```diff
@@ -17,136 +10,161 @@ class HyperTransformer:
...
+                Instance of transformer class that will fit the data.
+        """
+        transformer = load_transformer(transformer)
+        transformer.fit(data, field)
+        self._transformers_sequence.append(transformer)
+        output_types = transformer.get_output_types()
+        next_transformers = transformer.get_next_transformers()
+        for (output, output_type) in output_types.items():
+            output_field = self._multi_column_fields.get(output, output)
+            if (output_field not in self.field_transformers
```
amontanez24, line 157: I added this check first because I realized that in every other case we still want to run `transformer.transform`, but there are two cases where the `next_transformer` is `None`: 
1. If the column is part of a multi-column field and the other fields aren't ready
2. If we are done with the output
In the first case, we still need to transform the data but in the second case we don't want to and with the method we discussed it was hard to distinguish

csala: Interesting. As we discussed, I think this approach is right and the only comment here would be to maybe try to make it a bit more readable by renaming `next_transformer = get_next_transformer` to `next_transformer, transform = get_next_steps`, so all the `if/else` conditions are inside the same method.

amontanez24: I am now realizing there is a problem with this approach. We are looping through the output types for the transformer, and the `data = transformer.transform()` call is in the loop, meaning potentially the same transformer might attempt to transform twice. 

I think either we have to just call `transformer.transform()` outside the loop, or check in the beginning if a transform is needed which is if any of the `output_types` are not final or if any of the `output_columns` are in `field_transformers`. Thoughts?

amontanez24: Or a third approach which is to check if the transformation was already done or not. That's probably what I'll go with

amontanez24: I decided to just return a `next_transformer` whenever there is one. That way it is only null if the output field is final. Then the caller can decide whether to use the `next_transformer` or not based on if all the columns necessary are present in the data

rdt/transformers/base.py
```diff
@@ -233,12 +234,14 @@ def _reverse_transform(self, columns_data):
         """
         raise NotImplementedError()
 
-    def reverse_transform(self, data):
+    def reverse_transform(self, data, drop=True):
         """Revert the transformations to the original values.
 
         Args:
-            data (pandas.DataFrame):
+            data (pandas.DateFrame):
```
fealho, line 241: Typo, DataFrame

rdt/hyper_transformer.py
```diff
@@ -17,136 +10,163 @@ class HyperTransformer:
...
+
+        Create a ``HyperTransformer`` passing a ``field_transformers`` dict.
+        (Note: The transformers used in this example may not exist and are just used
+        to illustrate the different way that a transformer can be defined for a field).
 
-        Create a ``HyperTransformer`` passing a ``transformers`` dict.
+        >>> field_transformers = {
+        ...     'email': EmailTransformer(),
+        ...     'email.domain': EmailDomainTransformer(),
+        ...     ('year', 'month', 'day'): DateTimeTransformer()
```
fealho, line 51: I have a question: what's the expected behavior if a field is specified multiple times? For example:

```
field_transformers = {
    'email': EmailTransformer(),
    ('email',): OneHotEncoding(),
    ('email', 'phone_number'): OneHotEncoding(),
}
```
If this is not allowed, maybe we should specify it?

fealho (approved):
Looks good 👍

rdt/hyper_transformer.py
csala: Good point Felipe. I think that at some point we discussed that this would be validated, so the same column name can only be mentioned once, but it seems at the end it was not added.

csala (changes requested):

rdt/hyper_transformer.py
```diff
@@ -17,136 +10,163 @@ class HyperTransformer:
...
+            if field not in provided_fields:
+                self.field_types[field] = self._DTYPES_TO_DATA_TYPES[data[field].dtype.kind]
+
+    def _get_next_transformer(self, output, output_type, next_transformers):
+        next_transformer = None
+        if output in self.field_transformers:
+            next_transformer = self.field_transformers[output]
+
+        elif output_type not in self.transform_output_types:
+            if next_transformers is not None:
```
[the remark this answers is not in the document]
amontanez24: I thought that was required. Will add

rdt/hyper_transformer.py
amontanez24: Maybe I'll add a warning whenever there are multiple definitions.

amontanez24: Actually I don't think this makes sense. A dictionary can't have the same key twice. If they specify a tuple with only one value, that is a nonsensical input. I don't know if that's worth validating

csala: I am thinking that this should be an error that makes the code crash rather than a warning, because if the user really passes the same field twice we cannot work around it

rdt/hyper_transformer.py
```diff
@@ -155,18 +194,28 @@ def fit(self, data):
...
-        if self.transformers is not None:
-            self._transformers = load_transformers(self.transformers)
-        else:
-            self._transformers = self._analyze(data)
+        self._input_columns = list(data.columns)
+        self._update_field_types(data)
+        fitted_fields = set()
+
+        # Loop through field_transformers that are first level
+        for field in self.field_transformers:
```
csala, line 202: I think that should keep track of which input arguments were used during fit and which were not. If the user mistakenly passed a field name that ended up never being used we should issue a warning or raise an error (possibly just a warning):

```python
In [15]: ht = HyperTransformer(field_transformers={
    ...:     'something': SomeTransformer,
    ...:     'something_else': AnotherTransformer,
    ...:     })

In [16]: ht.fit(data)
<ipython-input-11-37d0f57b9965>:5: RuntimeWarning: Field "something_else" was specified in the input arguments was not found in the data
  warnings.warn('Field "something_else" was specified in the input arguments was not found in the data', RuntimeWarning)
```

rdt/hyper_transformer.py
```diff
@@ -155,18 +194,28 @@ def fit(self, data):
...
+            if self._field_in_data(field, data):
+                if self._field_in_set(field, fitted_fields):
+                    warnings.warn(FIELD_ALREADY_FIT_WARNING)
```
csala, line 205: As mentioned above, I think this should raise a `ValueError`
Also, I would try to do this validation in the `__init__`, since at that time we already know all the values given by the user.

csala (changes requested):
I think this is almost ready to go.
Just a comment about the input validations and warning the user about unused arguments

rdt/hyper_transformer.py
```diff
@@ -253,25 +278,13 @@ def reverse_transform(self, data):
...
-                    reversed_data = transformer.reverse_transform(data.rename(columns=rename))
-                    reversed_data = reversed_data[transformer.columns[0]].to_numpy()
-
-                data[column_name] = reversed_data
-                drop_columns.extend(set(columns) - {column_name})
-
-        data.drop(drop_columns, axis=1, inplace=True)
-
-        return data
+        for transformer in reversed(self._transformers_sequence):
```
csala, line 281: It looks like calling the `reverse_transform` methods of the transformers in reverse order produces the reversed columns in the wrong order (exactly the opposite than the one seen during fit). This seems unavoidable, but knowing that this happens I suggest that we add a way to restore the original order before returning.

NOTES:
- with this approach, `_temp_columns` may stop being needed too
- this `_subset` helper proposal can be used in other places of this module, and maybe somewhere else too. Should we create a helpers or utils module with these?

```python3
    @staticmethod
    def _subset(input_list, other_list, not_in=False):
        return [
            element
            for element in input_list
            if (element in other_list) ^ not_in
        ]

    def reverse_transform(self, data):
        unknown_columns = self._subset(data.columns, self._output_columns, not_in=True)
        for transformer in reversed(self._transformers_sequence):
            data = transformer.reverse_transform(data, drop=False)

        reversed_columns = self._subset(self._input_columns, data.columns)
        return reversed_data[unknown + reversed_columns]
```

This will keep all the unknown columns at the beginning of the table, and then add all the reversed columns at the end in the same order that was learned during `fit`. If the entire table is passed without any unknown columns, the result will be identical to the learned one.

csala (changes requested):
Added a comment about the order of the reversed columns

csala (approved):
Ready to go @amontanez24 !

PR #244, opened 2021-09-23
Update learn rounding digits to handle all nan data

katxiao:
Resolves https://github.com/sdv-dev/SDV/issues/574

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -82,7 +82,7 @@ class NumericalTransformer(BaseTransformer):
                 if (roundable_data == roundable_data.round(decimal)).all():
                     return decimal
 
-        else:
+        elif len(roundable_data) > 0:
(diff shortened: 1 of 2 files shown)


rdt/transformers/numerical.py
```diff
@@ -82,7 +82,7 @@ def _learn_rounding_digits(data):
                 if (roundable_data == roundable_data.round(decimal)).all():
                     return decimal
 
-        else:
+        elif len(roundable_data) > 0:
```
amontanez24, line 85: I assume `elif roundable_data` doesn't work because it still creates a series with no rows right?

amontanez24 (approved):
LGTM :shipit:

rdt/transformers/numerical.py
katxiao: Yeah it's an empty series

csala (approved):
Looks good!

PR #245, opened 2021-09-23
Update BaseTransformer

csala:
Proposing a few changes on top of #224 

Basically:
- Reorganize the helper methods to simplify the `fit`, `transform` and `reverse_transform` implementations
- Build the `output_columns` after having called `_fit` to allow the transformer to have information about the data before deciding about the outputs needed.
- Check for `shape` length instead of `pd.Series` type to support numpy outputs from the transformer `_fit` method.
- Reorder methods to follow the usual no-use-before-define standard.

### DIFF
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -1,5 +1,4 @@
 """BaseTransformer module."""
-import pandas as pd
 
 
 class BaseTransformer:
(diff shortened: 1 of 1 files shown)


amontanez24 (approved):
Changes look good to me!

fealho (approved):

PR #246, opened 2021-09-23
Add gaussian copula to init

katxiao:
Currently, `GaussianCopulaTransformer` isn't in the list of rdt `TRANSFORMERS`, because we only get the immediate subclasses of the `BaseTransformer`. Since `GaussianCopulaTransformer` is a subclass of the `NumericalTransformer`, it isn't included. This was causing issues when I tried to call `load_transformer` on the `GaussianCopulaTransformer`.

In order to get all transformers, we could recursively get the subclasses of the `BaseTransformer`.

### DIFF
--- a/rdt/transformers/__init__.py
+++ b/rdt/transformers/__init__.py
@@ -18,12 +18,25 @@ __all__ = [
     'NullTransformer',
     'OneHotEncodingTransformer',
     'LabelEncodingTransformer',
+    'GaussianCopulaTransformer',
 ]
+
+
+def _get_transformer_subclasses(base):
+    subclasses = base.__subclasses__()
+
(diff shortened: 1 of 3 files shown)


fealho (approved):
Looks good. I would've used a set instead of list, just in case of multiple inheritances at some point, but I don't think it really matters.

csala:
Something very similar is actually being implemented as part of #239 already.
I prefer the other implementation, where `get_subclasses` is a method of the `BaseTransformer` and it also takes care of skipping any `ABC` class, so I would opt for closing this PR without merging, in favor of the other one.

katxiao:
Makes sense, I'll close this PR and comment on the other one.

PR #247, opened 2021-09-23
Update the return dtype of boolean transformer

katxiao:
In the BooleanTransformer, we always cast the reverse transformed data to type `object`, because if a boolean series contains NaNs, it will be type `object`. However, if the input is a boolean series without NaNs (therefore type `bool`), the output type will be inconsistent.

We could instead try to first cast the output type to `bool`, and if there is a `ValueError` due to the presence of NaNs, we could cast it to `object`. The alternative is to first check if the series has any null values, but we have previously seen that checking for nulls is not very performant.

### DIFF
--- a/rdt/transformers/boolean.py
+++ b/rdt/transformers/boolean.py
@@ -85,4 +85,9 @@ class BooleanTransformer(BaseTransformer):
 
             data = pd.Series(data)
 
-        return np.round(data).clip(0, 1).astype('boolean').astype('object')
+        try:
+            out = np.round(data).clip(0, 1).astype('boolean').astype('bool')
+        except ValueError:
+            out = np.round(data).clip(0, 1).astype('boolean').astype('object')
+
+        return out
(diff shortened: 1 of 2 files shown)


csala:
Good catch @katxiao 
This may create conflicts with the changes coming in #228 , so I would wait for that to be merged first and then see if we need to add this on top of it.

On the other hand, I think that the implemented approach is fine but it may also produce inconsistent behaviors: One may have fitted the transformer with nulls (so, passing `object` as input) but later on be reverse transforming data that does not contain nulls. In this case, I would rather expect the output to be `object`, like what was seen in `fit`.

Which actually brings us to a new question: Rather than taking care about this here, at the `Transformer` implementation level, maybe we should update the new `BaseTransformer` implementation to capture the ` fit` input `dtypes` and simply cast back to them at the end of `reverse_transform`. By doing this, we make sure that any transformer always produces the same `dtypes` that it saw during `fit`, which are the ones that are truly relevant.

What do you think about it? CCing @amontanez24 @fealho too

katxiao:
Yeah, I think ideally the dtype should be set to whatever the input data's dtype is. We can either merge this first, or just wait until we make the larger `BaseTransformer` change (either is fine with me).

PR #258, opened 2021-09-28
Update hyper transformer tests

amontanez24:


tests/unit/test_hyper_transformer.py
```diff
@@ -1,101 +1,676 @@
...
+        expected = {'a': 'numerical', 'b': 'categorical', 'c': 'boolean', 'd': 'float'}
+        self.assertEqual(ht.field_types, expected)
+
+    @patch('rdt.hyper_transformer.load_transformer')
+    def test__fit_field_transformer(self, load_transformer_mock):
+        """Test the ``_fit_field_transformer`` method.
+
+        This tests that the ``_fit_field_transformer`` behaves as expected.
+        It should fit the transformer it is provided, loops through its
+        outputs, transform the data if the output is not ML ready and call
```
katxiao, line 272: Just a question - How do you determine if the output is ML ready?

tests/unit/test_hyper_transformer.py
```diff
@@ -128,53 +703,89 @@ def test_fit_transform(self):
...
+        # Setup
+        int_transformer = Mock()
+        int_out_transformer = Mock()
+        float_transformer = Mock()
+        categorical_transformer = Mock()
+        bool_transformer = Mock()
+        datetime_transformer = Mock()
         data = pd.DataFrame({
-            'a': [1, 2, 3],
+            'integer.out.value': [1, 2, 1, 3],
```
katxiao, line 732: Would it be worth it to pull out some of this input data / transformed data into a helper method? Seems like it's being reused to some extent.

tests/integration/test_hyper_transformer.py
```diff
@@ -1,9 +1,100 @@
...
+
+
+class DummyTransformerNotMLReady(BaseTransformer):
+
+    INPUT_TYPE = 'datetime'
+    OUTPUT_TYPES = {
+        'value': 'categorical',
+    }
+
+    def _fit(self, data):
```
csala, line 52: Seeing that a no-op `_fit` is added quite often, maybe we want to actually make the one from the `BaseTransformer` a no-op method instead of an abstract one? I would not do it in this PR, but we can open an issue to do evaluate it and maybe do it sometime in the future

tests/integration/test_hyper_transformer.py
```diff
@@ -1,9 +1,100 @@
...
+        pass
+
+    def _transform(self, data):
+        # Convert multiple columns into a single datetime
```
csala, line 56: This comment does not seem right

katxiao (approved):
lgtm with a few nits / questions

tests/integration/test_hyper_transformer.py
```diff
@@ -1,9 +1,100 @@
...
+        data.columns = [c.replace('_str.value', '') for c in data.columns]
+        data = pd.to_datetime(data)
+
+        out = pd.DataFrame(dict(zip(
+            self.output_columns,
+            [
+                data.values.astype(np.float64),
+
```
csala, line 82: :thinking:

tests/integration/test_hyper_transformer.py
```diff
@@ -101,110 +192,193 @@ def get_transformers():
...
-    transformers = get_transformers()
-
-    ht = HyperTransformer(transformers)
+    ht = HyperTransformer()
     ht.fit(data)
     transformed = ht.transform(data)
 
-    expected = get_transformed_nan_data()
+    pd.testing.assert_frame_equal(
+        transformed.sort_index(axis=1),
```
csala, line 245: I have doubts about this `sort_index`. Should the output data not be already sorted like the input by default?

tests/unit/test_hyper_transformer.py
```diff
@@ -1,101 +1,676 @@
...
+
+    @patch('rdt.hyper_transformer.HyperTransformer._create_multi_column_fields')
+    @patch('rdt.hyper_transformer.HyperTransformer._validate_field_transformers')
+    def test___init__(self, validation_mock, multi_column_mock):
         """Test create new instance of HyperTransformer"""
         # Run
         ht = HyperTransformer()
 
         # Asserts
         self.assertTrue(ht.copy)
```
csala, line 52: nit: Since we are already modifying this, I would change the `self.assertXXX` (unit test style) to `assert XXX` (pytest style)

tests/unit/test_hyper_transformer.py
```diff
@@ -1,101 +1,676 @@
...
+        """Test the ``_get_next_transformer method.
+
+        This tests that if the transformer is defined in the
+        ``instance.field_transformers`` dict, then it is returned
+        even if the output type is final.
+
+        Setup:
+            - field_transformers is given a transformer for the
+            output field.
+            - field_types will be given a different transformer
```
csala, line 109: This test (and the ones below) mention `field_types` but I believe this is `data_type_transformers`

tests/unit/test_hyper_transformer.py
```diff
@@ -1,101 +1,676 @@
...
-        hp = HyperTransformer(dtypes=['int', 'complex'])
+        data = pd.DataFrame({'a': [1, 2, 3]})
+        transformed_data1 = pd.DataFrame({
+            'a.out1': [2, 4, 6],
+            'a.out2': ['1', '2', '3']
+        })
+        transformer1 = Mock()
+        transformer2 = Mock()
+        transformer1.get_output_types.return_value = {
+            'a.out1': 'categorical',
```
csala, line 301: Is this correct? In the DF above it seems like `a.out1` is `numerical` and `out2` `categorical`

csala (approved):
Excellent (and intense) test suite @amontanez24 !
I added a couple of minor comments but once these (and @katxiao comments) are addressed  this should be ready to go! :rocket:

tests/unit/test_hyper_transformer.py
amontanez24: Good question! We check the `_transform_output_types` list. I will update this documentation to show that in the other PR. I'm trying to keep this PR strictly about tests.

tests/integration/test_hyper_transformer.py
amontanez24: Good point. I didn't actually sort in the `reverse_transform` method but I suppose we could.

tests/integration/test_hyper_transformer.py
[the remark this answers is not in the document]
amontanez24: Nope, good catch!

tests/unit/test_hyper_transformer.py
amontanez24: good catch!

tests/integration/test_hyper_transformer.py
amontanez24: This was because the way I made the expected output was incorrect. I can change that

tests/integration/test_hyper_transformer.py
```diff
@@ -233,34 +367,6 @@ def test_dtype_category():
...
-    ht.fit(data)
-
-    transformed = ht.transform(data)
-    reverse = ht.reverse_transform(transformed)
-
-    pd.testing.assert_frame_equal(data, transformed)
-    pd.testing.assert_frame_equal(data, reverse)
-
-
 def test_subset_of_columns():
```
csala, line 370: Maybe it would be interesting to add the opposite case: transform and reverse transform data that contains unexpected columns

tests/integration/test_hyper_transformer.py
```diff
@@ -101,110 +177,168 @@ def get_transformers():
...
-    expected = get_transformed_nan_data()
+    pd.testing.assert_frame_equal(transformed, expected_transformed)
 
-    np.testing.assert_allclose(
-        transformed.sort_index(axis=1).values,
-        expected.sort_index(axis=1).values
+    reversed = ht.reverse_transform(transformed)
+
+    pd.testing.assert_frame_equal(
+        expected_reversed.sort_index(axis=1),
```
csala, line 219: I kept thinking about this and did a couple tests, and I realized that the current behavior is not complete: If the entire table is transformed and then reverse transformed, the reversed data should be identical to the input data, including the columns order, which means that these `sort_index` calls should not be necessary.

I'll add a comment on the other PR about addressing this there, and then these tests can be updated

tests/integration/test_hyper_transformer.py
```diff
@@ -3,7 +3,80 @@
...
+        out = pd.DataFrame({
+            'year': datetimes.dt.year,
+            'month': datetimes.dt.month,
+            'day': datetimes.dt.day,
+        })
+
+        return out
 
 
 def get_input_data_with_nan():
```
csala, line 82: I'm thinking that we should complicate things a bit more by making this DataFrame use an index that is not the default 0, 1, 2, 3.
I suggest adding an `index` argument and pass something random, so we validate that the operations below do not mess the index.

csala (changes requested):

tests/integration/test_hyper_transformer.py
amontanez24: After discussion, we will open another issue for this because there does seem to be an indexing error in the `BaseTransformer` class

csala (approved):
This looks ready to go to me!

PR #260, opened 2021-09-29
Transformer subclasses

### LINKED ISSUE #243: Should Transformers subclasses exist for specific configurations?
We currently have several `Transformers` that accept parameters that modify their behavior.

For example, the `CategoricalTransformer` can be configured to be `fuzzy` or not, the `NumericalTransformer` can be configured to learn the `min_value`, `max_value` and `rounding` level, and the `DatetimeTransformer` can be configured to `strip_constant` and round the values to the same units as the input.

So the question here is: Should we create sub-classes of each one of these Transformers pinning the `__init__` parameters to some specific values, so they can also be used as classes without arguments?

Maybe not every parameter needs to be fixed in the subclasses, but at least the ones that produce very different behaviors.

For example, for the Transformers mentioned above, we could create:
* `CategoricalFuzzyTransformer`: Equivalent to `CategoricalTransformer(fuzzy=True)`
* `NumericalBoundedTransformer`: Equivalent `NumericalTransformer(min_value='auto', max_value='auto', rounding='auto')`
* `DatetimeRoundedTransformer`: Equivalent to `DatetimeTransformer(strip_constant=True)`

fealho:
Resolve #243.

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -258,6 +258,34 @@ class CategoricalTransformer(BaseTransformer):
         return self._reverse_transform_by_row(data)
 
 
+class CategoricalFuzzyTransformer(CategoricalTransformer):
+    """Transformer for categorical data.
+
+    This transformer computes a float representative for each one of the categories
+    found in the fit data, adding gaussian noise around the float representation
+    of each class, and then replaces the instances of these categories with
+    the corresponding representative.
+
+    The representatives are decided by sorting the categorical values by their relative
+    frequency, then dividing the ``[0, 1]`` interval by these relative frequencies, and
+    finally assigning the middle point of each interval to the corresponding category.
+
+    When the transformation is reverted, each value is assigned the category that
+    corresponds to the interval it falls in.
+
+    Null values are considered just another category.
+
+    This class behaves exactly as the ``CategoricalTransformer`` with ``fuzzy=True``.
+
+    Args:
+        clip (bool):
+            If ``True``, clip the values to [0, 1]. Otherwise normalize them using modulo 1.
+            Defaults to ``False``.
+    """
+    def __init__(self, clip=False):
+        super().__init__(fuzzy=True, clip=clip)
+
+
 class OneHotEncodingTransformer(BaseTransformer):
     """OneHotEncoding for categorical data.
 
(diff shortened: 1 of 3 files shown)


rdt/transformers/categorical.py
```diff
@@ -271,6 +300,7 @@ class OneHotEncodingTransformer(BaseTransformer):
         error_on_unknown (bool):
             If a value that was not seen during the fit stage is passed to
             transform, then an error will be raised if this is True.
+            Defaults to ``True``.
```
csala, line 303: good catch

rdt/transformers/numerical.py
```diff
@@ -196,6 +196,41 @@ def _reverse_transform(self, data):
         return data.astype(self._dtype)
 
 
+class NumericalBoundedTransformer(NumericalTransformer):
```
csala, line 199: I think we should be adding more options here, so the different configurations can be used combining whether the data is bounded or rounded.
This one, `NumericalBounded` would possibly be equivalent to `NumericalTransformer(min_value='auto', max_value='auto')`, but keeping `rounding=None`.

Another point of discussion is the name of the Transformers. The proposed name may work, but it may also be too verbose, or confusing, once we add all the options: `NumericalBoundedTransformer`, `NumericalRoundedTransformer` and `NumericalRoundedBoundedTransfomer`.
Another option is to call them `NumericalAutoMinMaxTransformer`, `NumericalAutoRoundingTransformer` and `NumericalFullAutoTransformer`, and I'm sure there are more.

Please free to propose better names if you can come up with them @fealho

csala (changes requested):

rdt/transformers/numerical.py
fealho: I'll go with your bounded/rounded suggestion, it's pretty good.

katxiao (approved):
LGTM!

rdt/transformers/categorical.py
```diff
@@ -258,6 +258,35 @@ def _reverse_transform(self, data):
         return self._reverse_transform_by_row(data)
 
 
+class CategoricalFuzzyTransformer(CategoricalTransformer):
+    """Transformer for categorical data.
+
+    This transformer computes a float representative for each one of the categories
```
csala, line 264: I would reorganize this description a bit to separate the fit and the transform more clearly, starting by saying how the transformer computes the category representative, then mentioning that during transform the categories are replaced by their representative with some gaussian noise, and finish on the reverse transform explanation that we already have.

csala (approved):
I added a comment about a description. Other than that, this looks good to go.

Issue #248, opened 2021-09-24
Create `validate_transformer_code_style` function


csala:
A function should be implemented to automatically validate the code style of any new Transformer.

## Function Name and Module

The function should be implemented inside `tests/contributing.py` and should be called `validate_transformer_code_style`

## Inputs

The function should accept a single input:

* `transformer (class or str)`: Transformer class or full Python name of the class (e.g. `DatetimeTransformer` or `"rdt.transformers.time.DatetimeTransformer"`)

## Outputs

The function should return a single `bool` value indicating whether the code style is correct or not.

## Behavior

The function should run all the third party code style checkers (`flake8`, `isort`, `pylint`...) as well custom code analysis checks and report whether all of them pass, as well as feedback about why they do not otherwise.

### List of Checks

The following validations must be performed and reported:

Check | Correct If | What does it check
-- | -- | --
flake8 | flake8 check passes | PEP8 standards and basic static code analysis
isort | isort check passes | Imports are properly sorted
pylint | pylint check passes | Static code analysis and programming style
pydocstyle | pydocstyle check passes | Whether the docstrings are written in the right format
Transformer Name | The Transformer name is correct | Transformer name ends with `Transformer` (e.g. `DatetimeTransformer`)
Unit Tests Exist | The unit tests module exists and is located in the right place | The unit tests module should be located called `tests/unit/transformers/<transformer_family>/<data_type>.py`
Unit Test function names | The unit test function names are correct | The unit test function names should be called `test_<method_name>_<test_case_name>`

</div></b>

### Prints to console

The function prints several messages on the console indicating:
* The path to the source file of the Transformer, which is being analyzed.
* A message indicating whether the code style is correct or not.
* A list of things that were validated and whether they were correct or not (list above).

### Success

![image](https://user-images.githubusercontent.com/6129283/134681632-082d2a75-ec57-4fbb-a665-87d346b2899d.png)

### External tool failure

![image1](https://user-images.githubusercontent.com/6129283/134681447-45dbad16-9edf-470a-8ca8-77dc90babc30.png)

### Custom validation failure

![image8](https://user-images.githubusercontent.com/6129283/134681759-2018862c-58d5-4e79-adee-202d72c28667.png)

pvk-developer:
The current format of our `docstrings` follows the [Google style guide](https://google.github.io/styleguide/pyguide.html), we can add `pydocstyle` and setting the `convention` to `google`. 

[...]

PR #261, opened 2021-09-29
Update performance tests

### LINKED ISSUE #257: Update performance test structure
### Problem Description

Performance tests can become more generalized if we specify the expected runtime per dataset, instead of per individual test cases.

### Expected behavior

The current performance tests are based on test cases that specify the transformer, the dataset generator and the sizes.

This should be changed to a more generic approach where each dataset generator indicates its expected output values for different data sizes, and the performance test simply takes a transformer and runs all the compatible datasets through it, validating against those expected output values.

katxiao:
Resolve #257


csala:
@katxiao Should this be put as ready for review?

tests/performance/test_performance.py
```diff
@@ -33,52 +31,126 @@ def get_fqn(obj):
...
+            A mapping of data type (str) to a list of transformers.
+    """
+    transformers = BaseTransformer.get_subclasses()
+    transformers_map = {}
+
+    for transformer in transformers:
+        input_type = transformer.get_input_type()
+        input_type_transformers = transformers_map.get(input_type, [])
+        input_type_transformers.append(transformer)
+        transformers_map[input_type] = input_type_transformers
```
amontanez24, line 58: minor: you could make `transformers_map` a `default_dict`. So these lines can be changed to something like
```
transformers_map = defaultdict(list)
transformers_map[input_type].append(transformer)
```

Also , there will be a method exactly like this in the `transformers.__init__.py` called `get_transformers_by_type`

amontanez24 (approved):
One small comment, but besides that looks good!

katxiao:
With this update, perf tests now take almost 2 hours, since we are now running more performance tests than before. Maybe we should reduce the number of dataset sizes that we run with.

tests/performance/test_performance.py
katxiao: Re-using the `get_transformers_by_type` method sounds good to me. I'll wait for that PR.

pvk-developer (approved):
LGTM!

Only comments I have are that this may fail with the future lint coming from #248 since there are missing the `get_performance_thresholds` description.

Maybe we can merge and fix later? 

This is the `PR` tha will make this fail in the future: https://github.com/sdv-dev/RDT/pull/276/files

PR #269, opened 2021-10-02
Add flake8-pytest-style

pvk-developer:
As part of #248 , add `flake8-pytest-style` and adapt the current code to pass this lint.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,7 @@ development_requires = [
     'flake8-sfs>=0.0.3,<0.1',
     'isort>=4.3.4,<5',
     'pylint>=2.5.3,<3',
+    'flake8-pytest-style>=1.5.0,<1.6',
 
     # fix style issues
     'autoflake>=1.1,<2',
--- a/tests/unit/transformers/test_boolean.py
+++ b/tests/unit/transformers/test_boolean.py
@@ -15,8 +15,8 @@ class TestBooleanTransformer(TestCase):
         transformer = BooleanTransformer()
 
         # Asserts
-        self.assertEqual(transformer.nan, -1, "Unexpected nan")
-        self.assertIsNone(transformer.null_column, "null_column is None by default")
+        assert transformer.nan == -1
+        assert transformer.null_column is None
 
     def test__fit_nan_ignore(self):
         """Test _fit nan equal to ignore"""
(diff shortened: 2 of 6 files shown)


tests/unit/transformers/test_boolean.py
```diff
@@ -15,8 +15,8 @@ def test___init__(self):
         transformer = BooleanTransformer()
 
         # Asserts
-        self.assertEqual(transformer.nan, -1, "Unexpected nan")
-        self.assertIsNone(transformer.null_column, "null_column is None by default")
+        assert transformer.nan == -1
```
fealho, line 18: @pvk-developer Is deleting all the comments really necessary? Some of them provide useful error messages.

setup.py
```diff
@@ -52,6 +52,7 @@
     'flake8-sfs>=0.0.3,<0.1',
     'isort>=4.3.4,<5',
     'pylint>=2.5.3,<3',
+    'flake8-pytest-style>=1.5.0,<1.6',
```
fealho, line 55: @pvk-developer Maybe cap it at 2?

fealho (approved):
Besides the couple of details, looks good.

amontanez24 (approved):
This looks good but I imagine it is going to create merge conflicts with the `HyperTransformer` tests changes. Please coordinate with me before merging :)

setup.py
csala: Yup, this should be <2 (next major)

tests/unit/transformers/test_boolean.py
csala: My 2 cents: I agree on that having assert comments can be helpful, but I have some doubts about these ones in particular being relevant enough to keep them there, since they basically just repeat what is being asserted. Consider my vote blank here. Your call @pvk-developer @fealho

csala (approved):
Approved, conditioned on the fixing the version range and reaching a conclusion about the comments

tests/unit/transformers/test_boolean.py
pvk-developer: @fealho and @csala I think that the comments in the code while asserting shouldn't be there at all, if we have to write some comments regarding what we are asserting it should be on the docstring of the test itself. My vote goes on this option (move the comments to the docstrings).

fealho: I agree it didn't matter too much either way, so since the comments have already been deleted I think it can be merged as is.

PR #270, opened 2021-10-02
Add pandas-vet flake8 addon

pvk-developer:
As part of #248 , added `pandas-vet` and adapted the code.


rdt/transformers/categorical.py
```diff
@@ -134,9 +134,9 @@ def _transform_by_category(self, data):
         for category, values in self.intervals.items():
             mean, std = values[2:]
             if category is np.nan:
-                mask = data.isnull()
+                mask = data.isna()
             else:
-                mask = (data.values == category)
+                mask = (data.array == category)
```
csala, line 139: I'd use `to_numpy()` instead.

rdt/transformers/categorical.py
```diff
@@ -198,7 +198,7 @@ def _reverse_transform_by_matrix(self, data):
 
         data = np.broadcast_to(data, (num_categories, num_rows)).T
         means = np.broadcast_to(self.means, (num_rows, num_categories))
-        diffs = np.abs(np.subtract(data, means))
+        diffs = np.abs(np.subtract(data, means))  # noqa
```
csala, line 201: why `#noqa`?

rdt/transformers/categorical.py
```diff
@@ -211,7 +211,7 @@ def _reverse_transform_by_category(self, data):
         # loop over categories
         for category, values in self.intervals.items():
             start = values[0]
-            mask = (start <= data.values)
+            mask = (start <= data.array)
```
csala, line 214: `to_numpy()`

rdt/transformers/datetime.py
```diff
@@ -79,21 +79,21 @@ def _find_divider(self, transformed):
...
                 break
 
             self.divider = candidate
 
     def _transform_helper(self, datetimes):
         """Transform datetime values to integer."""
-        nulls = datetimes.isnull()
-        integers = pd.to_numeric(datetimes, errors='coerce').values.astype(np.float64)
+        nulls = datetimes.isna()
+        integers = pd.to_numeric(datetimes, errors='coerce').array.astype(np.float64)
```
csala, line 90: Is the `array` needed here? I suspect this would also work without it

rdt/transformers/null.py
```diff
@@ -85,20 +85,20 @@ def transform(self, data):
...
+                    data[isna] = self._fill_value
                 else:
                     data = data.fillna(self._fill_value)
 
             if self._null_column:
-                return pd.concat([data, isnull.astype('int')], axis=1).values
+                return pd.concat([data, isna.astype('int')], axis=1).to_numpy()
 
-            if self._fill_value in data.values:
+            if self._fill_value in data.array:
```
csala, line 98: I suspect the `.array` is not needed here

tests/unit/transformers/test_boolean.py
```diff
@@ -181,7 +181,7 @@ def test__reverse_transform_not_null_values(self):
         expected = np.array([True, False, True])
 
         assert isinstance(result, pd.Series)
-        np.testing.assert_equal(result.values, expected)
+        np.testing.assert_equal(result.array, expected)
```
csala, line 184: `is_null()`

csala (changes requested):
Added a few comments about things that could possibly be changed.
On the other hand, it seems like the performance tests now fail, so we should figure out which one of the introduced changes slow the code down.

Can you try to figure this out @pvk-developer ?

I suggest `timeit` the following, using different input sizes and dtypes:
- `.values` vs `.array` vs `.to_numpy()`
- `pd.isnull` vs `pd.isna`

fealho (approved):

PR #272, opened 2021-10-02
Add flake8-use-fstring  addon to enforce the f-string usage.

pvk-developer:
As part of #248  added `flake8-use-fstring` to enforce the usage of `f'{variable}'` and changed the code where the old `.format` was being used.


katxiao:
Could we achieve the same thing by removing `SFS1` and  `SFS2` from `extend-ignore` [here](https://github.com/sdv-dev/RDT/blob/master/setup.cfg#L37)?

katxiao (commented):
Could we achieve the same thing by removing `SFS1` and  `SFS2` from `extend-ignore` [here](https://github.com/sdv-dev/RDT/blob/master/setup.cfg#L37)?

amontanez24 (approved):
I like this a lot! :shipit:

setup.py
```diff
@@ -52,6 +52,7 @@
     'flake8-sfs>=0.0.3,<0.1',
     'isort>=4.3.4,<5',
     'pylint>=2.5.3,<3',
+    'flake8-use-fstring>=1.1,<1.2',
```
csala, line 55: I think we can push this to <2

csala (approved):
Approved. We can review the version range, though

csala (changes requested):
Actually taking the approval back.

I agree with @katxiao : The result of this PR is an overlap of two plugins: `flake8-sfs` and `flake8-use-fstring`, which cover the same functionality.

I think that the PR needs to be updated to either:
- Remove the `flake8-sfs` plugin and its ignores from the `setup.cfg`
- Remove the `flake8-use-fstring` plugin and remove the `SFS2` ignore from `setup.cfg`, so only f-strings are accepted. Unless there is a problem with `flake8-sfs` that I am not aware of, I would go for this one, since we are already using this one without issues

PR #273, opened 2021-10-02
Add flake8-multiline-containers addon

pvk-developer:
As part of #248  added `flake8-multiline-containers` addon and adapted our code to follow this style check.


tests/unit/transformers/test_categorical.py
```diff
@@ -32,10 +32,24 @@ def test__get_intervals(self):
 
         # Asserts
         expected_intervals = {
-            'foo': (0, 0.5, 0.25, 0.5 / 6),
-            'bar': (0.5, 0.8333333333333333, 0.6666666666666666, 0.05555555555555555),
-            'tar': (0.8333333333333333, 0.9999999999999999, 0.9166666666666666,
-                    0.027777777777777776)
+            'foo': (
```
katxiao, line 35: Would the original `'foo': (0, 0.5, 0.25, 0.5 / 6),` be valid, since it opens and closes on same line? It seems a bit excessive to line break after every element here, since it fits neatly onto one line.

katxiao (approved):
Looks generally ok to me, but had a question

tests/unit/transformers/test_numerical.py
```diff
@@ -421,12 +421,14 @@ def test__reverse_transform_rounding_none_with_nulls(self):
         - First column of the input array as entered, replacing the indicated value with a Nan.
         """
         # Setup
-        data = pd.DataFrame([
```
amontanez24, line 424: I feel like this syntax show up a lot in our code. Why is this the only one that's changed?

amontanez24 (commented):
I just had a question about one of the changes before approving

tests/unit/transformers/test_numerical.py
pvk-developer: It's because of the `columns=['a', 'b']` they are written in one line and the others are splitted.
When we have a look we usually tend to visually approve that

```
data = pd.DataFrame({
        ...
        ...
        ...
        columns=['a', 'b'],
 })
 ```
 But this is actually becoming a inconsistent usage of the `[]` since you have array which it's values are splitted in multiple lines and another that is one-liner. So if we had no `columns` it would be fine.

File: tests/unit/transformers/test_categorical.py (hunk 1 above)
Comment (pvk-developer) [reply tests/unit/transformers/test_categorical.py:35]: It has to follow the same multi-line for the containers:
```python
        expected_intervals = {
            'foo': (0, 0.5, 0.25, 0.5 / 6),
            'bar': (0.5, 0.8333333333333333, 0.6666666666666666, 0.05555555555555555),
            'tar': (0.8333333333333333, 0.9999999999999999, 0.9166666666666666,
                    0.027777777777777776)  # this is different from the others, so it becomes invalid.
        }
``` 
If it was like this it would be valid:
```python
        expected_intervals = {
            'foo': (0, 0.5, 0.25, 0.5 / 6),
            'bar': (0.5, 0.8333333333333333, 0.6666666666666666, 0.05555555555555555),
            'tar': (0.8333333333333333, 0.9999999999999999, 0.9166666666666666, 0.027777)
        }

```

katxiao: Sounds good, thanks for the update.

Would prefer if a mixed style was allowed, but I'm fine with trying out the consistent styling requirements.

PR #274, opened 2021-10-02
Add flake8-quotes

pvk-developer:
As part of #248 added `flake8-quotes` addon and adapted our code to follow this style check.


setup.py
```diff
@@ -52,6 +52,7 @@
     'flake8-sfs>=0.0.3,<0.1',
     'isort>=4.3.4,<5',
     'pylint>=2.5.3,<3',
+    'flake8-quotes>=3.3.0,<3.4',
```
fealho, line 55: @pvk-developer maybe we should cap it at `<4`? It seems like they will release the 3.4.0 soon.

fealho (approved):

amontanez24 (approved):

setup.py
csala: I agree

PR #275, opened 2021-10-02
Add pep8-naming addon for flake8

pvk-developer:
As part of #248  add `pep8-naming` addon for flake8 and adapted our code to follow this style


amontanez24 (approved):
:shipit:

PR #279, opened 2021-10-04
Preserving index when transforming

### LINKED ISSUE #277: If the input data has a different index, the reverse transformed data may be out of order
### Environment Details

* RDT version: Any
* Python version: Any
* Operating System: Any

### Error Description

If the `DataFrame` input into the`HyperTransformer` has an index that isn't in incremental order, the reverse transformed data might come back in the wrong order. Some transformers reset the index and when the `BaseTransformer` assigns the new columns it is sometimes out of order. This happens in the following lines because the rows of `columns_data` will reorder to match the index of data.
https://github.com/sdv-dev/RDT/blob/10a996eebc8577692b49afc16a9be3417324d9fa/rdt/transformers/base.py#L126-L130

### Steps to reproduce

```python3
data = pd.DataFrame({
    'integer': [1, 2, 1, 3],
    'float': [0.1, 0.2, 0.1, 0.1],
    'categorical': ['a', 'a', 'b', 'a'],
    'bool': [False, False, True, False],
    'names': ['Jon', 'Arya', 'Jon', 'Jon'],
})
ht = HyperTransformer(data_type_transformers={'categorical': CategoricalTransformer})
ht.fit(data)
reversed = ht.reverse_transform(data)
```

amontanez24:
resolves #277

### DIFF
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -1,6 +1,8 @@
 """BaseTransformer module."""
 import abc
 
+import pandas as pd
+
 
 class BaseTransformer:
     """Base class for all transformers.
@@ -124,6 +126,9 @@ class BaseTransformer:
 
     @staticmethod
     def _set_columns_data(data, columns_data, columns):
+        if isinstance(columns_data, pd.Series) or isinstance(columns_data, pd.DataFrame):
+            columns_data.index = data.index
+
         if len(columns_data.shape) == 1:
             data[columns[0]] = columns_data
         else:
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -126,7 +126,7 @@ class BaseTransformer:
 
     @staticmethod
     def _set_columns_data(data, columns_data, columns):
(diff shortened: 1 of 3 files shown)


rdt/transformers/base.py
```diff
@@ -124,6 +126,9 @@ def _get_columns_data(data, columns):
 
     @staticmethod
     def _set_columns_data(data, columns_data, columns):
+        if isinstance(columns_data, pd.Series) or isinstance(columns_data, pd.DataFrame):
```
amontanez24, line 129: Using `assign` didn't fix the problem. I also realized that `columns_data` can be a Series, DataFrame or numpy array. So it seemed easiest to just fix the index if `columns_data` has one.

amontanez24: @csala I timed both using `to_numpy` and this method, and this seemed to be slightly faster for all `DataFrame` sizes

csala (approved):
This looks good!

katxiao (approved):
looks good 👍

PR #280, opened 2021-10-05
Applying the `NullTransformer` in the `HyperTransformer`

### LINKED ISSUE #262: Make the `HyperTransformer` apply the `NullTransformer`
Make the `HyperTransformer` (optionally) apply the `NullTransformer` after transforming everything else. This behavior should be controllable from the `HyperTransformer` `__init__` parameters by adding the following arguments:

- `transform_nulls (bool, default True)` : Whether to cleanup null values at the end of the transformations or not.
- `fill_value (str, object or dict)` : Imputation strategy ("mean" or "mode") or value to replace nulls with. It can be passed as a global value (str or object) or as a dict with values for specific columns.
- `null_column (bool, None or dict)` : Whether to create a null column to flag the nulls or not. It can be passed as a global value (bool or None) or as a dict with values for specific columns.

fealho:
Resolves #262.

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -3,6 +3,7 @@
 import warnings
 
 from rdt.transformers import get_default_transformer, load_transformer
+from rdt.transformers.null import NullTransformer
 
 
 class HyperTransformer:
@@ -118,7 +119,8 @@ class HyperTransformer:
             self._add_field_to_set(field, self._specified_fields)
 
     def __init__(self, copy=True, field_types=None, data_type_transformers=None,
-                 field_transformers=None, transform_output_types=None):
+                 field_transformers=None, transform_output_types=None, transform_nulls=True,
+                 fill_value=None, null_column=None):
         self.copy = copy
         self.field_types = field_types or {}
         self.data_type_transformers = data_type_transformers or {}
@@ -126,6 +128,9 @@ class HyperTransformer:
         self._specified_fields = set()
         self._validate_field_transformers()
         self.transform_output_types = transform_output_types or self._DEFAULT_OUTPUT_TYPES
+        self._transform_nulls = transform_nulls
+        self._fill_value = fill_value
+        self._null_column = null_column
         self._multi_column_fields = self._create_multi_column_fields()
         self._transformers_sequence = []
         self._output_columns = []
@@ -231,6 +236,17 @@ class HyperTransformer:
 
                 data = self._fit_field_transformer(data, field, transformer)
 
+        if self._transform_nulls:
+            output_fields = []
+            for transformer in self._transformers_sequence:
+                output_fields.extend(transformer.get_output_types().keys())
+
+            self._null_transformers = {}
+            for output_field in output_fields:
+                transformer = NullTransformer(self._fill_value, self._null_column)
+                transformer.fit(data[output_field])
+                self._null_transformers[output_field] = transformer
+
         self._validate_all_fields_fitted()
 
     def transform(self, data):
@@ -254,6 +270,11 @@ class HyperTransformer:
             data = transformer.transform(data, drop=False)
 
         transformed_columns = self._subset(self._output_columns, data.columns)
+
+        if self._transform_nulls:
+            for field, transformer in self._null_transformers.items():
+                data[field] = transformer.transform(data[field])
+
         return data.reindex(columns=unknown_columns + transformed_columns)
 
     def fit_transform(self, data):
@@ -281,6 +302,10 @@ class HyperTransformer:
             pandas.DataFrame:
                 reversed data.
         """
+        if self._transform_nulls:
+            for field, transformer in self._null_transformers.items():
+                data[field] = transformer.reverse_transform(data[field])
+
         unknown_columns = self._subset(data.columns, self._output_columns, not_in=True)
         for transformer in reversed(self._transformers_sequence):
             data = transformer.reverse_transform(data, drop=False)


fealho:
**Note**: I'm disabling lint R0902, which complains about `too-many-instance-attributes` in the `HyperTransformer.

setup.cfg
```diff
@@ -63,4 +63,4 @@ max-attributes = 11
 ignore-comments = yes
 ignore-docstrings = yes
 ignore-imports = yes
-disable = R0801, R0903, R0913, R0914, C0209, W0223, W0221, W0237
+disable = R0801, R0902, R0903, R0913, R0914, C0209, W0223, W0221, W0237
```
amontanez24, line 66: what is this for?

amontanez24 (commented):
Just a couple of comments

rdt/hyper_transformer.py
```diff
@@ -231,6 +237,18 @@ def fit(self, data):
...
 
+        if self._transform_nulls:
+            self._null_transformers = {}
+            for output_column in self._output_columns:
+                transformer = NullTransformer(self._fill_value, self._null_column)
+                transformer.fit(data[output_column])
+                self._null_transformers[output_column] = transformer
+
+        elif self._null_column or self._fill_value:
+            raise ValueError('When ``transform_nulls = False``, ``null_column`` and'
```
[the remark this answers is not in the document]
fealho: Yep, I forgot to push that change.

rdt/transformers/null.py
```diff
@@ -81,22 +82,21 @@ def transform(self, data):
         Returns:
             numpy.ndarray
         """
-        if self.nulls:
```
amontanez24, line 84: Does this have anything to do with the performance test failures?

fealho: Hmm, maybe, let me try.

fealho: We can't roll back the change that easily, because the previous code was faulty. I added a check before transforming the data to series, which made the code slightly faster, but not significantly. I'll try to see if I can speed it up somewhere else.

rdt/transformers/null.py
```diff
@@ -81,22 +75,19 @@ def transform(self, data):
...
-            if self._null_column:
-                return pd.concat([data, isna.astype('int')], axis=1).to_numpy()
+        if self.nulls and self._fill_value is not None:
+            if not self.copy:
+                data[isna] = self._fill_value
+            else:
+                data = data.fillna(self._fill_value)
 
-            if self._fill_value in data:
-                warnings.warn(IRREVERSIBLE_WARNING)
```
fealho, line 99: @csala So apparently, generating this warning takes the most time out of the whole process (fitting, transforming, reverse transforming)... if the data has lots of rows with nan's, over 90% of the time is spent generating this warning... for now I just removed the warning altogether, so it passes the test cases.

tests/integration/test_hyper_transformer.py
```diff
@@ -264,7 +235,7 @@ def test_hypertransformer_field_transformers():
     expected_transformed['datetime.value.value'] = [0.375, 0.875, 0.375, 0.375]
     expected_reversed = get_reversed()
 
-    ht = HyperTransformer(field_transformers=field_transformers)
+    ht = HyperTransformer(field_transformers=field_transformers, transform_nulls=False)
```
csala, line 238: Why is this added here? The default parameters should make this test pass

tests/integration/test_hyper_transformer.py
```diff
@@ -332,7 +303,7 @@ def test_hypertransformer_field_transformers_multi_column_fields():
     })
     expected_reversed = data.copy()
 
-    ht = HyperTransformer(field_transformers=field_transformers)
+    ht = HyperTransformer(field_transformers=field_transformers, transform_nulls=False)
```
csala, line 306: Same as above.

rdt/hyper_transformer.py
```diff
@@ -207,6 +220,10 @@ def _validate_all_fields_fitted(self):
             warnings.warn('The following fields were specified in the input arguments but not'
                           + f'found in the data: {non_fitted_fields}')
 
+    @staticmethod
+    def _get_null_column_name(field):
+        return ''.join(field.split('.')[:-1]) + '.is_null'
```
csala, line 226: This does not seem correct: What happens if the columns `something.transformed` and `something_else.transformed` exist? This would return `transformed.is_null` in both cases, which would be wrong? (Note: This scenario should be covered in an integration test)

The null column name should simply be `f{column_name}.is_null}`

tests/integration/test_hyper_transformer.py
```diff
@@ -383,7 +354,7 @@ def test_subset_of_columns():
     """
     data = get_input_data_without_nan()
 
-    ht = HyperTransformer()
+    ht = HyperTransformer(transform_nulls=False)
```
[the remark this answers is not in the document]
fealho: Done. This behavior is a little bit misleading though. For example, say the column `b` is `['abc', np.nan]`. If `null_column = True`, then reverse of `b.value` will be `['abc_transformed', 'filled_value']`, but if `null_column = False`, the reverse is `['abc_transformed', np.nan]`.

Basically, when only `column_name.value` is passed, the reverse transform will reverse the null transform if `null_column = False`, and won't otherwise. IMO, this behavior is a little unintuitive, maybe we should just reverse the null transform regardless of whether `b.is_null` is passed.

rdt/hyper_transformer.py
```diff
@@ -254,6 +284,21 @@ def transform(self, data):
             data = transformer.transform(data, drop=False)
 
         transformed_columns = self._subset(self._output_columns, data.columns)
+
+        if self._transform_nulls:
+            for field in data.columns:
```
amontanez24, line 289: can we pull all of these chunks into their own methods? One for `fit`, one for `transform` and one for `reverse_transform`

rdt/transformers/null.py
```diff
@@ -81,22 +82,22 @@ def transform(self, data):
         Returns:
             numpy.ndarray
         """
-        if self.nulls:
+        if self._null_column or (self.nulls and self._fill_value is not None):
+            isna = data.isna()
             if not isinstance(data, pd.Series):
```
amontanez24, line 87: Should this be moved outside of the if?

rdt/transformers/null.py
```diff
@@ -81,22 +82,22 @@ def transform(self, data):
...
-                if not self.copy:
-                    data[isna] = self._fill_value
-                else:
-                    data = data.fillna(self._fill_value)
+        if not self._null_column and (self._fill_value == data.array).any():
+            warnings.warn(IRREVERSIBLE_WARNING)
 
-            if self._null_column:
-                return pd.concat([data, isna.astype('int')], axis=1).to_numpy()
+        if self.nulls and self._fill_value is not None:
```
amontanez24, line 93: this is used a couple of times. Can we pull it into a variable name? This would also make it easier to understand what this check is for

PR #281, opened 2021-10-06
Validate transformer integration checks

### LINKED ISSUE #250: Create `validate_transformer_integration` function
A function should be implemented to automatically validate the Integration Tests of any new Transformer.

## Function Name and Module

The function should be implemented inside `tests/contributing.py` and should be called `validate_transformer_integration`

## Inputs

## Outputs

The function should return a single `bool` value indicating whether the integration tests passed or not.

## Behavior

This function runs the integration test functions on the Transformer. The integration tests are automated (the user does not write them explicitly) and simply generate data using all the Dataset Generators that have been created for the Transformer data type and then run the Transformer on them in two ways:
* Directly, transforming and reverse transforming every dataset and validating the output types. If the Transformer has been flagged as deterministic, this also validates that it actually is.
* Using the HyperTransformer, going from the Transformer Data Type to Numerical and back.

### List of Checks

Check | Correct if
-- | --
Transformer defines data type | The Transformer correctly defines the data type that it supports
Dataset Generators exist | At least one Dataset Generator exists for the Transformer data type
Transform output types | The Transformer can transform data and produces outputs of the indicated data types
Reverse output type | The Transformer can reverse transform the data it produces, going back to the original data type
Composition is Identity | (only if the Transformer indicates so) Transforming data and reversing it recovers the original data
HyperTransformer can transform | The HyperTransformer is able to use the Transformer and produce float values
HyperTransformer can reverse transform | The HyperTransformer is able to reverse the data that has previously transformer and restore the original data type

### Prints to console

* Whether the Integration tests were successful or not
* Which checks from the Integration tests were correct (list above)

## Usage Example

image4

katxiao:
Output (success): 
```
>>> valid = validate_transformer_integration('rdt.transformers.boolean.BooleanTransformer')
Validating Integration Tests for transformer rdt.transformers.boolean.BooleanTransformer

SUCCESS: The integration tests were successful.

Check                                   Correct    Details
--------------------------------------  ---------  ----------------------------------------------------------------------------------------------------------------------
Dataset Generators                      True       At least one Dataset Generator exists for the Transformer data type
Output Types                            True       The Transformer can transform data and produce output(s) of the indicated data type(s)
Reverse Transform                       True       The Transformer can reverse transform the data it produces, going back to the original data type
Composition is Identity                 True       Transforming data and reversing it recovers the original data, if composition is identity is specified
Hypertransformer can transform          True       The HyperTransformer is able to use the Transformer and produce float values
Hypertransformer can reverse transform  True       The HyperTransformer is able to reverse the data that it has previously transformed and restore the original data type
>>> valid
True
```

Output (validation error):
```
>>> validate_transformer_integration('rdt.transformers.BooleanTransformer')
Validating Integration Tests for transformer rdt.transformers.BooleanTransformer

ERROR: One or more integration tests were NOT successful.

Check               Correct    Details
------------------  ---------  ---------------------------------------------------------------------------------------
Dataset Generators  Yes        At least one Dataset Generator exists for the Transformer data type.
Output Types        Yes        The Transformer can transform data and produce output(s) of the indicated data type(s).
Reverse Transform   No         Reverse transformed data is not the expected data type boolean.
>>> valid
False
```

Output (transformer exception):

```
>>> validate_transformer_integration('rdt.transformers.BooleanTransformer')
Validating Integration Tests for transformer rdt.transformers.BooleanTransformer

ERROR: Transformer errored out with the following error:

Traceback (most recent call last):
  File "/RDT/tests/contributing.py", line 82, in validate_transformer_integration
    validate_transformer(transformer, steps=steps)
  File "/RDT/tests/integration/test_transformers.py", line 249, in validate_transformer
    _test_transformer_with_dataset(transformer, data, steps)
  File "/RDT/tests/integration/test_transformers.py", line 149, in _test_transformer_with_dataset
    transformer.fit(input_data, [TEST_COL])
  File "/RDT/rdt/transformers/base.py", line 168, in fit
    self._fit(columns_data)
  File "/RDT/rdt/transformers/boolean.py", line 65, in _fit
    raise ValueError('some error message')
Exception: some error message

Check               Correct    Details
------------------  ---------  --------------------------------------------------------------------
Dataset Generators  Yes        At least one Dataset Generator exists for the Transformer data type.

>>> valid
False
```
Resolves #250

### DIFF
--- a/tests/datasets/base.py
+++ b/tests/datasets/base.py
@@ -33,12 +33,7 @@ class BaseDatasetGenerator(ABC):
         subclasses = []
         for subclass in cls.__subclasses__():
             if ABC not in subclass.__bases__:
-                try:
-                    subclass()
-                except TypeError:
-                    pass
-                else:
-                    subclasses.append(subclass)
+                subclasses.append(subclass)
 
             subclasses += subclass.get_subclasses()
 
--- a/tests/datasets/boolean.py
+++ b/tests/datasets/boolean.py
@@ -1,5 +1,7 @@
 """Dataset Generators for boolean transformers."""
 
+from abc import ABC
+
 import numpy as np
 
 from tests.datasets.base import BaseDatasetGenerator
@@ -8,7 +10,7 @@ MAX_PERCENT_NULL = 50  # cap the percentage of null values at 50%
 MIN_PERCENT = 20  # the minimum percentage of true or false is 20%
 
 
-class BooleanGenerator(BaseDatasetGenerator):
+class BooleanGenerator(BaseDatasetGenerator, ABC):
     """Base class for generators that generate boolean data."""
 
     DATA_TYPE = 'boolean'
--- a/tests/datasets/categorical.py
+++ b/tests/datasets/categorical.py
@@ -1,5 +1,7 @@
 """Dataset Generators for categorical transformers."""
 
+from abc import ABC
+
 import numpy as np
 
 from tests.datasets.base import BaseDatasetGenerator
@@ -7,7 +9,7 @@ from tests.datasets.datetime import RandomGapDatetimeGenerator
 from tests.datasets.utils import add_nans
 
 
-class CategoricalGenerator(BaseDatasetGenerator):
+class CategoricalGenerator(BaseDatasetGenerator, ABC):
     """Base class for generators that generate catgorical data."""
 
     DATA_TYPE = 'categorical'
--- a/tests/datasets/datetime.py
+++ b/tests/datasets/datetime.py
@@ -1,6 +1,7 @@
 """Dataset Generators for datetime transformers."""
 
 import datetime
+from abc import ABC
 
 import numpy as np
 import pandas as pd
@@ -9,7 +10,7 @@ from tests.datasets.base import BaseDatasetGenerator
 from tests.datasets.utils import add_nans
 
 
-class DatetimeGenerator(BaseDatasetGenerator):
+class DatetimeGenerator(BaseDatasetGenerator, ABC):
     """Base class for generators that generate datatime data."""
 
     DATA_TYPE = 'datetime'
--- a/tests/datasets/numerical.py
+++ b/tests/datasets/numerical.py
@@ -1,12 +1,14 @@
 """Dataset Generators for numerical transformers."""
 
+from abc import ABC
+
 import numpy as np
 
 from tests.datasets.base import BaseDatasetGenerator
 from tests.datasets.utils import add_nans
 
 
-class NumericalGenerator(BaseDatasetGenerator):
+class NumericalGenerator(BaseDatasetGenerator, ABC):
     """Base class for generators that create numerical data."""
 
     DATA_TYPE = 'numerical'


csala (changes requested):
I see a few of problems in the suggested approach that we may want to fix:
1. The output table is not useful unless it provides meaningful messages that explain why the tests failed when they do. We should find a way to add more information by customizing the message to include the dataset in which the Transformer failed and what was wrong with it. Another option is to have a generic error message and print the traceback of the assertion when it fails, but it may be less readable for the user.
2. The `validate_transformer_integration` and `validate_transformer` implementations are too tight together. Even if `validate_transformer_integration` calls the `validate_transformer`, this one should always behave the same way, raising asserts when something fails, and `validate_transformer_integration` should deal with the `assert` to craft a meaningful message for the user.
3. I would rather `fail fast` than continue running all the tests until the end and then report all the errors. The reasons for this are that in most cases the same error will be observed for all the datasets, so reporting it on the first one is enough for the user to address it, and that failure to return a valid `transform` output is likely to also cause an error in the `reverse_transform` output, which could be false.
4. Capturing `asserts` would not be enough, since the Transformer itself can crash when running. The process should be aware of it and report such failures in a clean way.

tests/datasets/base.py
```diff
@@ -33,13 +33,19 @@ def get_subclasses(cls):
         subclasses = []
         for subclass in cls.__subclasses__():
             if ABC not in subclass.__bases__:
-                subclasses.append(subclass)
+                try:
+                    subclass()
```
csala, line 37: I suspect this is here to catch the classes like the [BooleanGenerator](https://github.com/sdv-dev/RDT/blob/50635411bbb7419cab48a436b036eb9702984c27/tests/datasets/boolean.py#L11) that are intended to be inherited by other classes that implement the `generate` method.

This should be achieved by adding `ABC` to the parent classes, so these steps here are not necessary

```
class BooleanGenerator(ABC, DatasetGenerator):
    ...
```

tests/contributing.py
```diff
@@ -0,0 +1,122 @@
...
+
+    result_summaries = []
+    seen_checks = set()
+    failed_step = None if validation_error is None else steps[-1]
+    for i in range(len(steps)):
+        check, details = CHECK_DETAILS[steps[i]]
+        if check in seen_checks:
+            continue
+
+        seen_checks.update([check])
```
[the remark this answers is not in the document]
csala: It seems there is an open issue about this: https://github.com/deppen8/pandas-vet/issues/74
I suggest we ignore `PD005` in `setup.cfg` until they fix it

tests/contributing.py
```diff
@@ -0,0 +1,123 @@
...
+    checks have passed or failed.
+
+    Args:
+        transformer (string or rdt.transformers.BaseTransformer):
+            The transformer to validate.
+    Output:
+        bool:
+            Whether or not the transformer passes all integration checks.
+    """
+    print(f'Validating Integration Tests for transformer {transformer}\n')
```
csala, line 79: I think this will print a weird message if a transformer class is passed instead of the string.
I suggest this is done below, after the if block that gets the class, and then print `transformer.__name__` so it always works.

tests/contributing.py
```diff
@@ -0,0 +1,123 @@
...
+    elif error_trace:
+        print('ERROR: Transformer errored out with the following error:\n')
+        print(error_trace)
+
+    result_summaries = []
+    seen_checks = set()
+    failed_step = None if validation_error is None else steps[-1]
+    for i in range(len(steps)):
+        check, details = CHECK_DETAILS[steps[i]]
+        if check in seen_checks:
```
csala, line 110: cool! so I suppose this skips the `_validate_composition` step if it was not run, right?

tests/contributing.py
```diff
@@ -0,0 +1,123 @@
...
+    elif validation_error:
+        print('ERROR: One or more integration tests were NOT successful.\n')
+    elif error_trace:
+        print('ERROR: Transformer errored out with the following error:\n')
+        print(error_trace)
+
+    result_summaries = []
+    seen_checks = set()
+    failed_step = None if validation_error is None else steps[-1]
+    for i in range(len(steps)):
```
csala, line 108: Why not just `for step in steps`?

tests/contributing.py
katxiao: Yeah `_validate_composition` will not be in the `steps` list if it was not run, so we will not add it to the printed results summary.

This `check in seen_checks` line makes sure that each check only shows up in the table once, since we run some checks multiple times (over multiple dataset generators).

csala: Oh, you are right. I misread this line.
If that's the case, maybe we could using a `set` would fix the duplicates problem without having to check here

PR #282, opened 2021-10-06
Fix performance test thresholds

katxiao:
I saw a few failures on the performance tests runs - update performance test thresholds so tests pass consistently


pvk-developer (approved):
LGTM! I had the same issue.

PR #286, opened 2021-10-06
Validate transformer performance checks

### LINKED ISSUE #251: Create `validate_transformer_performance` function
A function should be implemented to automatically validate the performance of any new Transformer.

## Function Name and Module

The function should be implemented inside `tests/contributing.py` and should be called `validate_transformer_performance`

## Inputs

## Outputs

The function should return `pandas.DataFrame` that contains information about the results obtained by the Transformer. The table contains 6 rows, time and memory for the fit, transform and reverse transform, and 5 columns:
* **Evaluation Metric**: What is being tested: Fit time, Transform memory…
* **Value**: The average amount across all the datasets that it ran on
* **Units**: Units of the amount, normalized by orders of magnitude of rows to produce 
* **Acceptable**: Whether the value us below the rejection threshold
* **Compared to Average**: Percentage of how the performance compares to the average of all the other transformers, where 1 means the performance is the same, >1 means the performance is worse, <1 means the performance is better.

### Output DataFrame Example

Evaluation_metric | Value | Units | Acceptable | Compared to Average
-- | -- | -- | -- | --
fit time | 1.2 | s / 1M rows | Yes | 1.345
transform time | 3.2 | s / 1M rows | Yes | 1.234
reverse time | 4532.3 | s / 1M rows | No | 24.452
fit memory | 10.4 | Mb / 1M rows | Yes | 0.654
transform memory | 14.3 | Mb / 1M rows | Yes | 0.698
reverse memory | 7.3 | Mb / 1M rows | Yes | 0.874

## Behavior

This function runs the Transformer on all the Dataset Generators of the indicated data type
and produces a report about its performance and how it compares to the other Transformers of the same data type.

### Prints to console

* Whether the Performance tests were successful or not

## Usage Example

image3

katxiao:
Resolves #251 

output:
```
>>> validate_transformer_performance('rdt.transformers.DatetimeTransformer')
Validating Performance for transformer DatetimeTransformer

SUCCESS: The Performance Tests were successful.
          Evaluation Metric         Value Acceptable     Units  Compared to Average
0                Fit Memory  9.334700e+01        Yes  Mb / row             0.757455
1                  Fit Time  6.232677e-07        Yes   s / row             0.574041
2  Reverse Transform Memory  1.451382e+02        Yes  Mb / row             0.966153
3    Reverse Transform Time  6.641531e-07        Yes   s / row             1.080660
4          Transform Memory  8.896317e+01        Yes  Mb / row             0.656664
5            Transform Time  5.217231e-07        Yes   s / row             0.484631
```

```
validate_transformer_performance('rdt.transformers.BooleanTransformer')
Validating Performance for transformer BooleanTransformer

SUCCESS: The Performance Tests were successful.
          Evaluation Metric         Value Acceptable     Units  Compared to Average
0                Fit Memory  3.861500e+01        Yes  Mb / row                  NaN
1                  Fit Time  2.393918e-07        Yes   s / row                  NaN
2  Reverse Transform Memory  1.943598e+02        Yes  Mb / row                  NaN
3    Reverse Transform Time  2.238662e-06        Yes   s / row                  NaN
4          Transform Memory  1.170530e+02        Yes  Mb / row                  NaN
5            Transform Time  8.394433e-07        Yes   s / row                  NaN
```

### DIFF
--- a/tests/contributing.py
+++ b/tests/contributing.py
@@ -17,9 +17,9 @@ from tests.code_style import (
     validate_transformer_name, validate_transformer_subclass)
 from tests.datasets import get_dataset_generators_by_type
 from tests.integration.test_transformers import validate_transformer
+from tests.performance import evaluate_transformer_performance, validate_performance
 from tests.quality.test_quality import (
     TEST_THRESHOLD, get_regression_scores, get_results_table, get_test_cases)
-from tests.performance import evaluate_transformer_performance, validate_performance
 
 # Mapping of validation method to (check name, check description).
 CHECK_DETAILS = {


tests/contributing.py
```diff
@@ -121,3 +126,65 @@ def validate_transformer_integration(transformer):
...
+    final_results = pd.DataFrame()
+    final_results['Evaluation Metric'] = total_results['Test']
+    final_results['Value'] = total_results['Value'] / len(dataset_generators)
+    final_results['Units'] = np.where(
+        final_results['Evaluation Metric'].str.contains('Time'),
+        's / row',
+        'Mb / row',
+    )
+    final_results['Acceptable'] = np.where(total_results['Valid'], 'Yes', 'No')
+    average = total_results['Total'] / (len(transformers) * len(dataset_generators))
```
csala, line 187: The average should be computed without including the current transformer.
If there are no other transformers, the `Compared to average` column can be `N/A` or `np.nan`

tests/performance/test_performance.py
```diff
@@ -83,8 +99,55 @@ def _validate_metric_against_threshold(actual, expected_unit, size):
...
+        should_assert (bool):
+            Whether or not to raise AssertionErrors.
+        results (dict or None):
+            If not None, performance results will be collected and stored in results.
+    """
+    transformer_instance = transformer()
+    expected = dataset_generator.get_performance_thresholds()
+
+    dataset_sizes = _get_dataset_sizes(dataset_generator.DATA_TYPE)
+    for sizes in dataset_sizes:
```
csala, line 119: I think that this should be simplified a bit further by re-organizing some things:
- One function should just run the loop over the sizes and produce the maximum metric relative to the size.
- Another function validates the obtained value against the specific thresholds. This can happen only once by metric using the maximum relative value, since we are only interested on whether the maximum is valid or not. This function is the one that is told whether to assert or not.
- Another function, which is the actual test, calls both of the previous functions, telling the second one to assert.
- Another function, which is the user-facing function, calls both of the functions above without asserting, and then aggregates, etc.

Inside `test_performance.py`

```python3
def evaluate_transformer_performance(transformer, dataset_generator):
    transformer_instance = transformer()

    sizes = _get_dataset_sizes(dataset_generator.DATA_TYPE)

    out = []
    for fit_size, transform_size in sizes:
        performance = profile_transformer(
            transformer=transformer_instance,
            dataset_generator=dataset_generator,
            fit_size=fit_size,
            transform_size=transform_size,
        )
        size = np.array([fit_size, transform_size, transform_size] * 2)
        out.append(performance / size)

    return pd.DataFrame(out).max(axis=0)


def validate_performance(performance, dataset_generator, should_assert):
    expected = dataset_generator.get_performance_thresholds()
    out = []
    for test_name, value in performance.items():
        function, metric = test_name.lower().replace(' ', '_').rsplit('_', 1)
        valid = value < expected[function][metric]
        out.append(valid)
        if should_assert and not valid:
            raise AssertionError()

    return out


@pytest.mark.parametrize(('transformer', 'dataset_generator'), test_cases)
def test_performance(transformer, dataset_generator):
    performance = evaluate_transformer_performance(transformer, dataset_generator)
    validate_performance(performance, dataset_generator, should_assert=True)
```

Inside `contributing.py` (See comments above too):

```python
def validate_transformer_performance(desired_transformer):
    ...
    total_results = pd.DataFrame()
        for dataset_generator in dataset_generators:
            performance = evaluate_transformer_performance(transformer, dataset_generator)
            valid = validate_performance(performance, dataset_generator, should_assert=False)

            results = pd.DataFrame({
                'Value': performance.to_numpy(),
                'Valid': valid,
                'transformer': transformer.__name__,
                'dataset': dataset_generator.__name__,
            })
            results['Evaluation Metric'] = performance.index
            total_results = total_results.append(results)

    ... aggregate...
```

tests/contributing.py
```diff
@@ -121,3 +126,65 @@ def validate_transformer_integration(transformer):
...
+    if isinstance(desired_transformer, str):
+        desired_transformer = get_class(desired_transformer)
+
+    print(f'Validating Performance for transformer {desired_transformer.__name__}\n')
+
+    data_type = desired_transformer.get_input_type()
+    transformers = get_transformers_by_type().get(data_type, [])
+    dataset_generators = get_dataset_generators_by_type().get(data_type, [])
+
+    total_results = pd.DataFrame(
```
csala, line 155: I think this loop may be too complex.
We may be able to simplify if we do not make any aggregation until the end, and during the loop we just accumulate results (see more comments further down, on the other file).

```python3
total_results = pd.DataFrame()
    for dataset in datasets:
        results = ...compute results for transformer and whether they are valid or not...
        total_results = total_results.append(results)

other_results = total_results[total_results.transformer != transformer.__name__]
average = other_results.groupby('Evaluation Metric')['Value'].mean()    # this may be empty if there are no others.

total_results = total_results[total_results.transformer == transformer.__name__]
final_results = total_results.groupby('Evaluation Metric').agg({
    'Value': 'mean',
    'Valid': 'any'
}).reset_index()

# From here on, `final_results` is already the output, but needs to be still formatted and cleaned.
```

PR #287, opened 2021-10-08
Rdt quality tests

amontanez24:


tests/quality/test_quality.py
```diff
@@ -0,0 +1,123 @@
...
+        array = array.to_numpy()
+
+    if len(array.shape) == 1:
+        array = array.reshape(-1, 1)
+    return array
+
+
+def get_regression_score(features, target):
+    model = LinearRegression()
+    scores = cross_val_score(model, features, target)
```
csala, line 36: What metric is being used here? If `cross_val_score` uses the model `.score` method it may end up using `mse`, which is not what we want.

tests/quality/test_quality.py
```diff
@@ -0,0 +1,123 @@
...
+}
+
+
+def format_array(array):
+    if not isinstance(array, np.ndarray):
+        array = array.to_numpy()
+
+    if len(array.shape) == 1:
+        array = array.reshape(-1, 1)
+    return array
```
csala, line 31: Blank line above? :-)

tests/quality/test_quality.py
```diff
@@ -0,0 +1,123 @@
...
+        target = numerical_transformer.fit_transform(target, list(target.columns))
+        target = format_array(target)
+        scores = []
+        for transformer in transformers:
+            ht = HyperTransformer(data_type_transformers={data_type: transformer})
+            ht.fit(features)
+            transformed_features = ht.transform(features).to_numpy()
+            scores.append(get_regression_score(transformed_features, target))
+        all_scores[column] = pd.Series(scores, index=index)
+    return all_scores
```
csala, line 69: Can we add a couple of blank lines here?

tests/quality/test_quality.py
```diff
@@ -0,0 +1,123 @@
...
+    means = scores_without_transformer.mean()
+    means_above_threshold = means > THRESHOLD
+    standard_deviations = scores_without_transformer.std()
+    minimum_scores = means[means_above_threshold] - standard_deviations[means_above_threshold]
+
+    assert all(scores.loc[transformer, means_above_threshold] > minimum_scores)
+
+
+def get_test_cases():
+    max_size = 5000000
```
katxiao, line 83: nit: can make this (and types_to_skip) a constant at the top of the file if we're going to keep using it

tests/quality/test_quality.py
```diff
@@ -0,0 +1,123 @@
...
+    test_cases = []
+    types_to_skip = {'numerical', 'float', 'integer'}
```
katxiao, line 85: Why are we skipping these types?

tests/quality/test_quality.py
```diff
@@ -0,0 +1,123 @@
...
+                    test_cases.append((data_type, row['name']))
+
+    return test_cases
+
+
+test_cases = get_test_cases()
+
+
+@pytest.mark.parametrize('test_case', test_cases)
+def test_quality(subtests, test_case):
```
katxiao, line 101: I think you can do something like `@pytest.mark.parameterize(('dataset_name', 'data_type'), test_cases)` so you can pass the variables in separately `def test_quality(subtests, dataset_name, data_type)`. I haven't tried it out with `subtests` but it worked for me otherwise.

tests/quality/test_quality.py
amontanez24: It uses the default for the estimator which is coefficient of determination for `LinearRegression`. 
https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LinearRegression.html
https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_val_score.html

csala: Oh, cool! I thought the default for LR was mse! Good then

tests/quality/test_quality.py
```diff
@@ -0,0 +1,143 @@
...
+import numpy as np
+import pandas as pd
+from sklearn.linear_model import LinearRegression
+from sklearn.model_selection import cross_val_score
+
+from rdt import HyperTransformer
+from rdt.transformers import NumericalTransformer, get_transformers_by_type
+from tests.quality.utils import download_single_table_dataset
+
+THRESHOLD = 0.2
```
amontanez24, line 12: I lowered this because there were no tests that had higher correlations than 0.4 for datetime data types

tests/quality/test_quality.py
```diff
@@ -0,0 +1,143 @@
...
+    transformers_by_type = get_transformers_by_type()
+    test_cases = get_test_cases()
+    data_types_to_test = [
+        data_type
+        for data_type in transformers_by_type.keys()
+        if data_type not in TYPES_TO_SKIP
+    ]
+    tested_data_types = dict.fromkeys(data_types_to_test, False)
+
+    for data_type, dataset_name, table_name in test_cases:
```
csala, line 130: It seems like we may be able to run multiple times on the same dataset, once per `data_type` so I think it would be better to split this in two loops:

```python3
for dataset_name, table_name, data_types in test_cases:
    data, metadata = download_single_table(dataset_name, table_name)
    for data_type in data_types:
        ...
```

tests/quality/test_quality.py
```diff
@@ -0,0 +1,143 @@
...
+        transformers = transformers_by_type[data_type]
+        (data, metadata) = download_single_table(dataset_name, table_name)
+        scores = get_transformer_scores(data, data_type, transformers, metadata)
+        if any(scores.mean() > THRESHOLD):
+            tested_data_types[data_type] = True
+
+        for transformer in scores.index:
```
csala, line 137: This loop should be done after the one above is finished.
I think the logic should be made a bit more similar to the performance tests one, where we:
1. Identify the transformers and datasets that need to be run
2. Evaluate and accumulate all the scores (loop above)
3. Validate the scores for the target transformer (this loop, possibly inside another function, running an ALL the accumulated results from point 2.)

Point 3 is the `subtest` level, where each `transformer` is a `test case`, and there are three possible outcomes for each one of them:
a. There are no valid datasets for the transformer data type, so the test fails.
b. The transformer is too bad when compared to the other datasets, so the test fails.
c. The transformer is good enough when compared to the other datasets, so the test passes.

amontanez24: I actually think this approach may make things very messy unless we change the way the scores are computed/stored. Currently, each transformer has a set of datasets that scores are computed for. For each dataset, multiple scores are computed: one for every numeric column. We agreed on this because one transformer might be able to find correlations with one column, while a different transformer for the same data type finds correlations with another. In this case, both would be valid since they are each capable of producing data that exposes relationships.

The problem is that this makes step 2 hard, because all the scores cannot really be accumulated in one object. It isn't a table that just has the transformer type, dataset and score. There is a column for each target column's score, and those column names differ between datasets. Accumulating everything into one DataFrame would be a mess.

Maybe it could be a nested dict mapping {data_type: {dataset: scores}}. Then, for each transformer, we could get it's datasets and for each of those:
1. Validate at least one of the datasets has a high enough average score for at least one column
2. Compare the score for the transformer to the score of the others on those datasets

csala (approved):
This looks great @amontanez24 !
I think it can be merged

PR #288, opened 2021-10-08
Custom transformers code validations

pvk-developer:
As part of #248 add custom code style checks for the transformers and addon transformers.

### DIFF
--- a/tests/code_style.py
+++ b/tests/code_style.py
@@ -9,7 +9,6 @@ from types import FunctionType
 
 import pytest
 
-import rdt.transformers
 from rdt.transformers import TRANSFORMERS
 from rdt.transformers.base import BaseTransformer
 


tests/code_style.py
```diff
@@ -0,0 +1,88 @@
...
+    return issubclass(transformer, BaseTransformer)
+
+
+def validate_transformer_module(transformer):
+    """Return whether or not the ``Transformer`` is inside the right module."""
+    transformer_file = Path(inspect.getfile(transformer))
+    transformer_folder = transformer_file.parent
+
+    is_valid = False
+    if transformer_folder.match('transformers'):
```
csala, line 28: By this logic, it seems like all these are valid:
- `rdt/transformers/my_module.py`
- `rdt/transformers/my_family/my_module.py`
- `rdt/transformers/addons/my_addon/my_module.py`

What about `rdt/transformers/addons/my_addon/my_family/my_module.py`?
Should it be valid too?

amontanez24: I guess it depends how we want to organize things. I could see the possibility of having a broad `add-on` that has multiple sub-families in it, so I think this should be valid. For example, a location `add-on` that has a `family` for addresses and another one for geolocation or something like that.

tests/code_style.py
```diff
@@ -0,0 +1,217 @@
...
+                module = _load_module_from_path(directory)
+                transformers_found.update({
+                    name: obj
+                    for name, obj in inspect.getmembers(module, inspect.isclass)
+                })
+
+    return transformers_found
+
+
+def find_all_transformers():
```
amontanez24, line 211: what is this used for?

tests/code_style.py
```diff
@@ -0,0 +1,247 @@
...
+
+def get_all_transformers_as_dict():
+    """Return all the transformers."""
+    return {
+        transformer.__name__: transformer
+        for transformer in BaseTransformer.get_subclasses()
+    }
+
+
+def validate_transformer(transformer):
```
csala, line 209: I think this should be renamed to something like `test_transformer_code_style` and parameterized using `rdt.transformers.TRANSFORMERS` as input (values are test cases, keys are test ids):

```python3
@pytest.mark.parametrize('transformer', TRANSFORMERS.values(), ids=TRANSFORMERS.keys())
def test_transformer_code_style(transformer):
```

Inside the function, collecting results is not necessary: we just need to call the functions, which should raise an `AssertionError` if something is wrong.

After this, a `pytest tests/code_style.py -v --disable-warnings --no-header` call can be added to the `lint` command.

tests/code_style.py
```diff
@@ -0,0 +1,247 @@
...
+
+    valid_addon = validate_transformer_addon(transformer)
+    if valid_addon is None:
+        valid_addon = 'NaN'
+
+    result['valid_addon'] = valid_addon
+    return result
+
+
+def validate_all_transformers():
```
csala, line 231: If we run this as a parameterized `test`, this and the `main` block below are not necessary any more.

amontanez24 (approved):
LGTM!

PR #292, opened 2021-10-15
Fix transformer imports

csala:
Reorganize the `transformer` imports to stop skipping the `__init__.py` file in `flake8` and also to ensure that all the transformers are included in both the `rtd.transformers` module and the `rdt.transformers.TRANSFORMERS` dictionary.


rdt/transformers/__init__.py
```diff
@@ -2,61 +2,54 @@
...
+
+def _import_addons():
+    """Import all the addon modules."""
+    addons_path = Path(__file__).parent / 'addons'
+    for addon_json_path in addons_path.glob('*/*.json'):
         with open(addon_json_path, 'r', encoding='utf-8') as addon_json_file:
             transformers = json.load(addon_json_file).get('transformers', [])
             for transformer in transformers:
                 transformer = transformer.split('.')
                 module = '.'.join(transformer[:-1])
```
pvk-developer, line 38: I know this is my old code but I think we can change it to `rsplit` which is better:
```python3
module = transformer.rsplit('.', 1)[0]
```

PR #299, opened 2021-10-19
Create validate_transformer_quality function

### LINKED ISSUE #253: Create `validate_transformer_quality` function
A function should be implemented to automatically validate the data quality of any new Transformer by running the quality tests mentioned in #252 and reporting the results.

## Function Name and Module

The function should be implemented inside `tests/contributing.py` and should be called `validate_transformer_quality`

## Inputs

## Outputs

The function should return `pandas.DataFrame` that contains information about the results obtained by the Transformer. The dataframe contains one row per dataset, and the following columns (column names here):
* Dataset Name: The name of the dataset
* Score: The score obtained on the dataset (TBD)
* Acceptable: Whether the value is above the rejection threshold (TBD)
* Compared to Average: Percentage of how the quality compares to the average of all the other transformers, where 1 means the quality is the same, >1 means the quality is worse, <1 means the quality is better.

### Output DataFrame Example

Dataset Name | Score | Acceptable | Compared to Average
-- | -- | -- | --
Dataset A | 0.6534 | Yes | 1.345
Dataset B | 0.4423 | Yes | 1.234
Dataset C | 0.3498 | No | 0.548

## Behavior

This function runs all the quality tests using the Transformer on all the real world datasets that contain the Transformer data type and produces a report based on how good the correlations are preserved and how good a synthetic data generator (a copulas.GaussianMultivariate?) is when trained on the data produced by this Transformer, also comparing it to the quality of the other transformers of the same Data Type

### Prints to console

* Whether the Quality tests were successful or not

## Usage Example

image

amontanez24:
resolves #253 

Example use:
```
>>> validate_transformer_quality(CategoricalTransformer)
Validating Quality Tests for transformer CategoricalTransformer

SUCCESS: The quality tests were successful.

Dataset                    Score    Compared To Average  Acceptable
----------------------  --------  ---------------------  ------------
adult                   0.223325               0.442723  True
student_placements      0.45749                0.994534  True
student_placements_pii  0.45749                0.982039  True
```

### DIFF
--- a/tests/contributing.py
+++ b/tests/contributing.py
@@ -9,11 +9,14 @@ import traceback
 import pandas as pd
 from tabulate import tabulate
 
+from rdt.transformers import get_transformers_by_type
 from tests.code_style import (
     load_transformer, validate_test_location, validate_test_names, validate_transformer_addon,
     validate_transformer_importable_from_parent_module, validate_transformer_module,
     validate_transformer_name, validate_transformer_subclass)
 from tests.integration.test_transformers import validate_transformer
+from tests.quality.test_quality import (
+    TEST_THRESHOLD, get_regression_scores, get_results_table, get_test_cases)
 
 # Mapping of validation method to (check name, check description).
 CHECK_DETAILS = {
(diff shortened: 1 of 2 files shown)


tests/contributing.py
```diff
@@ -121,3 +124,65 @@ def validate_transformer_integration(transformer):
...
+    a transformer name, dataset name, average score for the dataset,
+    a score comparing the transformer's average score for the dataset to
+    the average of the average score for the dataset across all transformers of
+    the same data type, and whether or not the score passed the test threshold.
+    3. The table described above is printed when this function is run.
+
+    Args:
+        transformer (string or rdt.transformers.BaseTransformer):
+            The transformer to validate.
+    Output:
```
csala, line 156: This should be `Returns:`, with a `blank` line above this one.

tests/contributing.py
```diff
@@ -121,3 +124,65 @@ def validate_transformer_integration(transformer):
...
+        bool:
```
csala, line 157: It seems like this is not describing the actual output.

csala (approved):
Just a comment about a docstring that seems incorrect.
Other than that, this looks good to go

tests/contributing.py
amontanez24: Good catch!

PR #300, opened 2021-10-19
Baseclass unit tests

fealho:
Write test cases for the baseclass.

### DIFF
--- a/tests/unit/transformers/test_base.py
+++ b/tests/unit/transformers/test_base.py
@@ -47,7 +47,7 @@ class TestBaseTransformer:
             containing only a ``INPUT_TYPE`` attribute.
 
         Output:
-            - the string stored in the ``INPUT_TYPE`` attribute.q
+            - the string stored in the ``INPUT_TYPE`` attribute.
         """
         # Setup
         class Dummy(BaseTransformer):


tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+        # Setup
+        dictionary = None
+        base_transformer = BaseTransformer()
+
+        # Run
+        output = base_transformer._add_prefix(dictionary)
+
+        # Assert
+        expected = None
+        assert output == expected
```
csala, line 62: I do not thing a variable is necessary for these trivial cases. An `assert output is None` may be more readable.

tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+
+    def test__add_prefix_dictionary(self):
+        """Test the ``_add_prefix`` method when passed a dictionary.
+
+        Setup:
+            - set the ``column_prefix`` of the ``BaseTransformer`` to ``'colulumn_name'``.
```
csala, line 68: column_name

tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+            -
+
+        Expected behavior:
+            -
+        """
+        pass  # TODO: not sure how to test this one.
+
+    def test_get_input_type(self):
+        """Test the ``get_input_type`` method.
+
```
csala, line 27: Let's add a specific description of what is expected here. For example:

 "The base `get_input_type` method should return the value defined in the `INPUT_TYPE` of the child classes".

tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+            'column_name.month': 'categorical',
+            'column_name.year': 'numerical'
+        }
+        assert output == expected
+
+    def test_get_output_types(self):
+        """Test the ``get_output_types`` method.
+
+        Setup:
+            - set the ``column_prefix`` of the ``BaseTransformer`` to ``'colulumn_name'``.
```
csala, line 100: Let's use a subclass like in the `get_input_type` above.
Notice that this particular implementation is changing the value of the `OUTPUT_TYPES` of the base class, potentially affecting other tests that depend on it.

tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+        expected = {
+            'column_name.value': 'numerical'
+        }
+        assert output == expected
+
+    def test_is_transform_deterministic(self):
+        """Test the ``is_transform_deterministic`` method.
+
+        Setup:
+            - set the ``DETERMINISTIC_TRANSFORM`` of the ``BaseTransformer`` to True.
```
csala, line 127: Same as `get_input_type`

tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+
+        # Assert
+        expected = True
+        assert output == expected
+
+    def test_is_reverse_deterministic(self):
+        """Test the ``is_reverse_deterministic`` method.
+
+        Setup:
+            - set the ``DETERMINISTIC_REVERSE`` of the ``BaseTransformer`` to True.
```
csala, line 147: Same as `get_input_type`

tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+        """Test the ``is_composition_identity`` method.
+
+        Setup:
+            - set the ``COMPOSITION_IS_IDENTITY`` of the ``BaseTransformer`` to True.
+
+        Output:
+            - the booloan value stored in ``COMPOSITION_IS_IDENTITY``.
+        """
+        # Setup
+        base_transformer = BaseTransformer()
```
csala, line 173: Same as `get_input_type`

tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+
+        # Assert
+        expected = True
+        assert output == expected
+
+    def test_get_next_transformers(self):
+        """Test the ``get_next_transformers`` method.
+
+        Setup:
+            - set the ``column_prefix`` of the ``BaseTransformer`` to a string.
```
csala, line 187: Same as `get_output_types`

tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+
+    def test__store_columns_list(self):
+        """Test the ``_store_columns`` method when passed a list.
+
+        The method should store the passed columns in the ``columns`` attribute.
+
+        Input:
+            - a data frame.
+            - a list of a subset of the columns of the dataframe.
+        """
```
csala, line 218: What is the expected side effect?

tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+    def test__store_columns_tuple(self):
+        """Test the ``_store_columns`` method when passed a tuple.
+
+        The method should first convert the tuple to list, an then store the passed columns
+        in the ``columns`` attribute.
+
+        Input:
+            - a data frame.
+            - a tuple of a subset of the columns of the dataframe.
+        """
```
csala, line 245: side effect?

tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+
+        # Run
+        base_transformer._store_columns(columns, data)
+        stored_columns = base_transformer.columns
+
+        # Assert
+        expected = ['a', 'b']
+        assert stored_columns == expected
+
+    def test__store_columns_tuple(self):
```
csala, line 236: We need to distinguish 2 test cases here:
- The tuple exists as a column name within the data frame
- The tuple does not exist as a column within the data frame (but each element within the tuple does)

tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+    def test__store_columns_string(self):
+        """Test the ``_store_columns`` method when passed a string.
+
+        The method should first convert the passed string into a one element list, and then
+        store this one column in the ``columns`` attribute.
+
+        Input:
+            - a data frame.
+            - a string with the name of one of the columns of the dataframe.
+        """
```
csala, line 272: Side effect?

tests/unit/transformers/test_base.py
```diff
@@ -1,10 +1,374 @@
...
+        columns = ['b']
+
+        # Run
+        columns_data = BaseTransformer._get_columns_data(data, columns)
+
+        # Assert
+        expected = pd.Series([4, 5, 6], name='b')
+        pd.testing.assert_series_equal(columns_data, expected)
+
     def test__set_columns_data_preserves_order_series(self):
```
csala, line 372: We can remove the `preserves_order` from this test name and the one below, since these are just test cases for `series` and `dataframe`

csala: We should also add 2 test cases that pass 1d and 2d numpy arrays

tests/unit/transformers/test_base.py
```diff
@@ -72,3 +436,191 @@ def test__set_columns_data_preserves_order_dataframe(self):
...
+            hashtags, with a hashtag added at the end (e.g. ['a', 'b'] -> 'a#b#').
+            - ``output_columns`` should be set to a list of the keys of what's returned from the
+            ``get_output_types`` method.
+        """
+
+    def test_fit(self):
+        """Test the ``fit`` method.
+
+        Setup:
+            - mock ``_fit``.
```
csala, line 479: By `mock` you mean `patch`?

Another option is to define a subclass on the fly that implements a `fit` method that stores the input data, so later on you can assert on it:

```python3
def test_fit(self):
    # setup
    class Dummy(BaseTransformer):
        def fit(self, data):
            self._passed_data = data

    # run
    ...
    dummy = Dummy()
    dummy.fit(data)

    # assert
    assert dummy._passed_data is the expected one
```

tests/unit/transformers/test_base.py
```diff
@@ -72,3 +436,191 @@ def test__set_columns_data_preserves_order_dataframe(self):
...
+            present in the data
+
+        Input:
+            - a dataframe.
+
+        Output:
+            - the original data.
+        """
+
+    # TODO: should we test all data types being passed? i.e. Series, df and np?
```
csala, line 510: Right now only DF can be passed, so a single test should be enough

tests/unit/transformers/test_base.py
```diff
@@ -72,3 +436,191 @@ def test__set_columns_data_preserves_order_dataframe(self):
...
+    def test_transform_drop_false(self):
+        """Test the ``transform`` method when ``drop=False``.
+
+        Setup:
+            - set ``self.columns`` to a list of columns from the data.
+            - set `self.output_columns` to the output column names.
+            - mock self._transform and return some dummy data.
```
csala, line 517: Same as above, one option is to `patch` the method, but we can also define a dummy transformer that implements a `transform` that simply stores the input data and returns a modified copy of it:

```python3
class Dummy(BaseTransformer):
    def _transform(self, data):
        self._passed_data = data.copy()
        return data.copy() * 2
```

tests/unit/transformers/test_base.py
```diff
@@ -72,3 +436,191 @@ def test__set_columns_data_preserves_order_dataframe(self):
...
+        Setup:
+            -
+
+        Input:
+            -
+
+        Expected behavior:
+            -
+        """
+        pass  # TODO: not sure how to test this? just mock fit, transform, and make sure they are each called once?
```
csala, line 566: Right. I think that here the cleanest is to just Mock the entire thing:

```python3
self = Mock(spec_set=BaseTransformer)

input_data = 'input_data'
out = BaseTransformer.fit_transform(self, input_data)

self.fit.assert_called_once_with('input_data')
self.transform.assert_called_once_with('input_data')
assert out == self.transform.return_value
```

tests/unit/transformers/test_base.py
```diff
@@ -1,11 +1,437 @@
...
+            'column_name.year': 'numerical'
+        }
+        assert output == expected
+
+    def test_get_output_types(self):
+        """Test the ``get_output_types`` method.
+
+        Setup:
+            - create a ``Dummy`` class which inherits from the ``BaseTransformer`` where:
+                - ``column_prefix`` is set to ``'colulumn_name'``.
```
csala, line 119: typo

tests/unit/transformers/test_base.py
```diff
@@ -28,16 +455,18 @@ def test__set_columns_data_preserves_order_series(self):
         columns = ['c']
```
csala, line 455: I think that the input dataframe should not contain the output column, since in the real usage when this method is called it will always be creating new columns. The same applies to the other variations of this test.

csala (commented):
I added a couple of minor comments, but other than that this looks good @fealho !

Once we address those minor changes and fix the code style errors this sees ready to go for me.

tests/unit/transformers/test_base.py
```diff
@@ -52,23 +489,546 @@ def test__set_columns_data_preserves_order_dataframe(self):
...
+        # Run
+        BaseTransformer._set_columns_data(data, columns_data, columns)
+
+        # Assert
+        expected = pd.DataFrame({
+            'a': [1, 2, 3],
+            'b': [4, 5, 6],
+            'c': [7, 8, 9]
+        }, index=[2, 0, 1])
+        pd.testing.assert_frame_equal(data, expected, check_dtype=False)
```
fealho, line 547: @csala I added this check because on Windows data becomes `int32`, while the expected is `int64`.

PR #302, opened 2021-10-19
Validate transformer unit tests

pvk-developer:
```python
>>> validate_transformer_unit_tests(IdentityTransformer)
Validating source file /home/developer/Projects/sdv-dev/RDT/rdt/transformers/addons/identity/identity.py

========================================== test session starts ==========================================
collected 3 items                                                                                       

tests/unit/transformers/addons/identity/test_identity.py::TestIdentityTransformer::test__fit PASSED [ 33%]
tests/unit/transformers/addons/identity/test_identity.py::TestIdentityTransformer::test__reverse_transform PASSED [ 66%]
tests/unit/transformers/addons/identity/test_identity.py::TestIdentityTransformer::test__transform PASSED [100%]

===================================== 3 passed, 1 warning in 0.01s ======================================

SUCCESS: The unit tests passed.

ERROR: The unit tests only cover 25.0% of your code.

Full coverage report here:

Name                                           Stmts   Miss  Cover   Missing
----------------------------------------------------------------------------
rdt/transformers/addons/identity/identity.py       8      6    25%   3-13, 25, 37-47
----------------------------------------------------------------------------
TOTAL                                              8      6    25%

>>> value
0.25

```

### DIFF
--- a/tests/contributing.py
+++ b/tests/contributing.py
@@ -1,18 +1,19 @@
 """Validation methods for contributing to RDT."""
 
-
 import importlib
 import inspect
 import subprocess
 import traceback
 
+import coverage
 import pandas as pd
+import pytest
 from tabulate import tabulate
 
 from tests.code_style import (
-    load_transformer, validate_test_location, validate_test_names, validate_transformer_addon,
-    validate_transformer_importable_from_parent_module, validate_transformer_module,
-    validate_transformer_name, validate_transformer_subclass)
+    get_test_location, load_transformer, validate_test_location, validate_test_names,
+    validate_transformer_addon, validate_transformer_importable_from_parent_module,
+    validate_transformer_module, validate_transformer_name, validate_transformer_subclass)
 from tests.integration.test_transformers import validate_transformer
 
 # Mapping of validation method to (check name, check description).
@@ -268,7 +269,7 @@ def validate_transformer_code_style(transformer):
     Args:
         transformer (string or rdt.transformers.BaseTransformer):
             The transformer to validate.
-    Output:
+    Returns:
         bool:
             Whether or not the transformer passes all code style checks.
     """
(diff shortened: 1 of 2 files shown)


tests/contributing.py
```diff
@@ -268,7 +269,8 @@ def validate_transformer_code_style(transformer):
     Args:
         transformer (string or rdt.transformers.BaseTransformer):
             The transformer to validate.
-    Output:
+
+    Returns:
```
amontanez24, line 273: I noticed some of the other validation functions say `Output:` instead of `Returns`. I know that's not necessarily tied to this PR, but since I think this is the last one that will touch this file, can we make them all consistent?

tests/contributing.py
```diff
@@ -304,6 +308,61 @@ def validate_transformer_code_style(transformer):
...
+    if rounded_score < 1.0:
+        print(f'\nERROR: The unit tests only cover {round(score, 3)}% of your code.')
+    else:
+        print(f'\nSUCCESS: The unit tests cover {round(score, 3)}% of your code.')
+
+    cov.html_report()
+    print('\nFull coverage report here:\n')
+    coverage_name = module_name.replace('.', '_')
+    export_dir = Path('htmlcov') / f'{coverage_name}_py.html'
+    print(export_dir.absolute().as_uri())
```
csala, line 361: Minor thing: This will print the full path as a URI, so it is clickable when the output is shown in a command line.
However, in a jupyter notebook this will be a plain string, which will not be clickable.

Could we move this into a function that detects whether we are in jupyter environment or not and, if we are, prints an HTML link that is clickable?

pvk-developer: Actually that can't be done because `jupyter-notebook` blocks the `localfile` links. I have attempted several approaches but those end up generating a `403` error code where you can't access the `css` files and others.

After a `1-1` discussion we decided to just print the path to where the file is expected and let the user copy and paste it in his browser.

katxiao (approved):
lgtm 👍

PR #305, opened 2021-10-19
Improve integration tests

csala:
Fix a problem on the `HyperTransformer` type detection when being passed a `bool` column with `nan` values, which was incorrectly being labeled as `categorical`.

Also re-organize and clean up a bit the HyperTransformer integration tests.

### DIFF
--- a/tests/integration/test_hyper_transformer.py
+++ b/tests/integration/test_hyper_transformer.py
@@ -134,7 +134,6 @@ def get_transformed_data():
     }, index=index)
 
 
-
 DETERMINISTIC_CATEGORICAL = deepcopy(DEFAULT_TRANSFORMERS)
 DETERMINISTIC_CATEGORICAL['categorical'] = CategoricalTransformer
 
--- a/tests/integration/test_hyper_transformer.py
+++ b/tests/integration/test_hyper_transformer.py
@@ -157,16 +157,16 @@ def test_hypertransformer_default_inputs():
         - The transformed data should contain all the ML ready data.
         - The reverse transformed data should be the same as the input.
     """
-    # setup
+    # Setup
     data = get_input_data()
 
-    # run
+    # Run
     ht = HyperTransformer()
     ht.fit(data)
     transformed = ht.transform(data)
     reverse_transformed = ht.reverse_transform(transformed)
 
-    # assert
+    # Assert
     expected_transformed = get_transformed_data()
     pd.testing.assert_frame_equal(transformed, expected_transformed)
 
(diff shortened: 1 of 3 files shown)


tests/integration/test_hyper_transformer.py
```diff
@@ -238,7 +183,7 @@ def test_hypertransformer_field_transformers():
     till it is.
 
     Setup:
-        - The datetime column is set to use a dumm transformer that stringifies
+        - The datetime column is set to use a dummy transformer that stringifies
```
amontanez24, line 186: good catch!

tests/integration/test_hyper_transformer.py
```diff
@@ -332,99 +280,87 @@ def test_hypertransformer_field_transformers_multi_column_fields():
...
 
+    # assert
     pd.testing.assert_frame_equal(subset, reverse)
 
 
 def test_with_unfitted_columns():
     """HyperTransform should be able to transform even if there are unseen columns in data."""
-    data = get_input_data_without_nan()
-
+    # setup
```
amontanez24, line 349: minor: I noticed some files capitalize these. I think we should try and keep consistent, so either have it capitalized everywhere or lower case everywhere

tests/integration/test_hyper_transformer.py
```diff
@@ -83,110 +85,60 @@ def _reverse_transform(self, data):
...
-            'class': 'BooleanTransformer'
-        },
-        'datetime': {
-            'class': 'DatetimeTransformer'
-        },
-        'names': {
-            'class': 'CategoricalTransformer',
-        },
-    }
+DETERMINISTIC_CATEGORICAL = deepcopy(DEFAULT_TRANSFORMERS)
```
amontanez24, line 137: minor: could we make this more explicit. Perhaps `DEFAULTS_WITH_DETERMINISTIC_CATEGORICAL`

tests/integration/test_hyper_transformer.py
```diff
@@ -83,110 +85,60 @@ def _reverse_transform(self, data):
...
-    })
-    if index:
-        transformed.index = index
-
-    return transformed
-
-
-def get_transformed_nan_data():
+def get_transformed_data():
+    index = [4, 6, 3, 8, 'a', 1.0, 2.0, 3.0]
```
amontanez24, line 113: Since this is the same in both the input data and transformed, should we pull it out?

amontanez24 (approved):
A few minor comments but looks good to me!

fealho (approved):
Looks good 👍

PR #306, opened 2021-10-19
Issue 304 DeprecationWarning of `np.float` and `np.int`

### LINKED ISSUE #304: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`
Running the tests on branch v0.6.0-dev issues a huge number of `DeprecationWarnings` about `np.float` not being valid any more.

```python3
tests/unit/transformers/test_numerical.py::TestNumericalTransformer::test__fit
  .../RDT/tests/unit/transformers/test_numerical.py:30: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    transformer = NumericalTransformer(dtype=np.float, nan='nan')
```

We should update the code to replace `np.float` with `float` or `np.float64`.

csala:
Resolve #304 

Change `np.float` to `float` (or `np.float64`), and `np.int` to `int`

### DIFF
--- a/tests/datasets/categorical.py
+++ b/tests/datasets/categorical.py
@@ -49,7 +49,7 @@ class RandomIntegerNaNsGenerator(CategoricalGenerator):
     @staticmethod
     def generate(num_rows):
         """Generate a ``num_rows`` number of rows."""
-        return add_nans(RandomIntegerGenerator.generate(num_rows).astype(np.float))
+        return add_nans(RandomIntegerGenerator.generate(num_rows).astype(float))
 
     @staticmethod
     def get_performance_thresholds():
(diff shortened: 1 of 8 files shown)


amontanez24 (commented):
What about these lines? https://github.com/sdv-dev/RDT/blob/9b118c43be113ca369344844f2506303835712c6/rdt/transformers/__init__.py#L54-L55
I remember we added them because windows had a different default for float or something like that

PR #310, opened 2021-10-20
Validate test coverage

### LINKED ISSUE #309: Add repo-wide automation
### Problem Description

We should validate that the entire repo has 100% test coverage

katxiao:
Check for 100% coverage and add Makefile target to validate repo

Resolves #309


Makefile
```diff
@@ -124,6 +124,9 @@ test: test-unit test-integration test-readme ## test everything that needs test
 test-all: ## test using tox
 	tox -r
 
+.PHONY: test-repo
+test-repo: lint test-unit test-integration test-readme test-performance test-quality ## test everything
```
amontanez24, line 128: could we just use `test` to replace `test-unit`, `test-integration` and `test-readme` since it covers them?

PR #329, opened 2021-12-01
Add a bayesian gaussian mixture model transformer

### LINKED ISSUE #183: Add a `BayesianGMM` Transformer
The main idea of this transformer is to take a continuous value and map it into a continuous + categorical value (which is represented with a one-hot-encoding). To implement this, most of the code can be taken from [here](https://github.com/sdv-dev/CTGAN/blob/98ae39d261a8d2cad709c91c502e7ee0cef4694a/ctgan/data_transformer.py), specifically the `_fit_continuous`, `_transform_continuous`, and `_reverse_transform_continuous` functions.

1. The `fit` method will train the `BayesianGMM` model. A GMM model has multiple "modes", where each "mode" is a Gaussian.
2. The `transform` method will take the raw values and will randomly select a mode (proportional to the probability of the value coming from that mode). Then, it generates a normalized version of the raw value by (approximately) subtracting/dividing by the mean/std.
3. The `reverse_transform` method will select the maximum likelihood mode and reverse the transformation.

Two things to note:
1. Multiple calls to `fit` can return different numbers of output columns. For example, in one run, it might infer 3 modes in the data, resulting in 4 output columns. In another run, if a random seed isn't set, it may infer a different number of modes.
2. Multiple calls to `transform` can generate different outputs. Although the result is always reversible, the mode is randomly selected (proportional to the likelihood).

When implementing this in RDT, we may want to allow the user to enable/disable these.

fealho:
Resolve #183.


rdt/transformers/numerical.py
```diff
@@ -487,3 +488,142 @@ def _reverse_transform(self, data):
...
+        normalized_values = ((data - means) / (4 * stds))[:, self._valid_component_indicator]
+        component_probs = self._bgm_transformer.predict_proba(data)
+        component_probs = component_probs[:, self._valid_component_indicator]
+
+        num_components = self._valid_component_indicator.sum()
+        selected_component = np.zeros(len(data), dtype='int')
+        for i in range(len(data)):
+            component_porb_t = component_probs[i] + 1e-6
+            component_porb_t = component_porb_t / component_porb_t.sum()
+            selected_component[i] = np.random.choice(np.arange(num_components), p=component_porb_t)
```
amontanez24, line 560: Is it possible to redo this section as a vector operation? Something like
```
component_probs = (component_probs + 1e-6) / component_probs.sum()
np.random.choice(np.arange(num_components), p=component_probs)
```

amontanez24 (commented):
Left a couple of comments but it's looking good so far!

rdt/transformers/numerical.py
fealho: The problem is that `p` in `np.random.choice` needs to be 1-D. While there are work arounds (I didn't find any alternative functions that accept multi-dimensional `p`, but we could always write our own), I'm not sure if they are worth it.

rdt/transformers/numerical.py
```diff
@@ -615,12 +615,13 @@ def _reverse_transform(self, data, sigma=None):
...
             pandas.Series
         """
-        data = pd.DataFrame(data)
-        one_hot = np.zeros(shape=(data.shape[0], self._number_of_modes))
-        continuous_name, discrete_name = self.output_columns  # pylint: disable=W0632
-        discrete_column = data[discrete_name].tolist()
-        one_hot[np.arange(data.shape[0]), discrete_column] = 1.0
-        data = np.concatenate([data[continuous_name][:, None], one_hot], axis=1)
+        print(data.shape)
+        one_hot = np.zeros(shape=(data.shape[1], self._number_of_modes))
```
amontanez24, line 619: why are we changing the indexing?

tests/unit/transformers/test_numerical.py
```diff
@@ -1391,60 +1391,78 @@ def test__transform(self):
...
         ])
-        np.testing.assert_allclose(result['continuous'].to_numpy(), expected_continuous, rtol=0.01)
+        np.testing.assert_allclose(result['continuous'].to_numpy(), expected_continuous, atol=0.1)
 
         expected_discrete = np.array([1, 1, 1, 1, 1, 0, 0, 0, 0, 0])
         np.testing.assert_allclose(result['discrete'].to_numpy(), expected_discrete)
 
     def test__reverse_transform_helper(self):
-        """Test '_inverse_transform_continuous' with sigmas != None.
+        """Test '_inverse_transform_helper' with `sigma != None`."""
```
amontanez24, line 1400: Let's use the same format we've been using for unit tests and have a description for `Setup`, `Input` and `Output`.

tests/unit/transformers/test_bayes_gmm.py
```diff
@@ -134,60 +65,58 @@ def test__transform_continuous(self):
...
-        """Test 'transform' on a dataframe with one continuous and one discrete columns.
-
-        It should use the appropriate '_transform' type for each column and should return
-        them concanenated appropriately.
-
-        Setup:
-            - Mock _column_transform_info_list
-            - Mock _transform_discrete
-            - Mock _transform_continuous
+        """Test 'transform'."""
```
amontanez24: minor: can we change all of the `'` to be the ` character?

tests/code_style.py
```diff
@@ -163,7 +163,7 @@ def validate_test_names(transformer):
     module = _load_module_from_path(test_file)
 
     test_class = getattr(module, f'Test{transformer.__name__}', None)
-    assert test_class is not None, 'The expected test class was not found.'
+    assert test_class is not None, f'{module} n,t.hueo {transformer.__name__}'
```
amontanez24, line 166: What is this change for?

fealho: It was for debugging, forgot to change back.

tests/unit/transformers/test_numerical.py
```diff
@@ -1291,61 +1291,58 @@ class TestBayesGMMTransformer(TestCase):
...
-        """Test '_transform_continuous'.
-
-        The `transform` method first computes the probability that the
-        continuous value came from each component, then samples a component based
-        on that probability and finally returns a (value, onehot) tuple, where the
-        onehot vector indicates the component that was selected and the value
-        is a normalized representation of the continuous value based on the mean
-        and std of the selected component.
+    def test__transform(self):
+        """Test `_transform.
```
amontanez24, line 1329: I think this is missing the closing `

tests/unit/transformers/test_numerical.py
```diff
@@ -1291,61 +1291,58 @@ class TestBayesGMMTransformer(TestCase):
 
     @patch('rdt.transformers.numerical.BayesianGaussianMixture')
     def test__fit(self, mock_bgm):
-        """Test '_fit'."""
+        """Test '_fit'.
```
amontanez24, line 1294: minor: can we switch `'` to `. Or whatever we pick, let's just use throughout the file

tests/integration/transformers/test_numerical.py
```diff
@@ -135,3 +136,21 @@ def test_int_nan(self):
 
         reverse = ct.reverse_transform(transformed)
         np.testing.assert_array_almost_equal(reverse, data, decimal=2)
+
+
+class TestBayesGMMTransformer:
+
+    def test_simple(self):
```
amontanez24, line 143: Can we rename this to be more explicit about what it's testing? Also can we add some tests with nulls similar to the others in this file?

tests/unit/transformers/test_numerical.py
fealho: I missed that one, thanks!

rdt/transformers/numerical.py
```diff
@@ -487,3 +490,160 @@ def _reverse_transform(self, data):
...
+        """
+        self._bgm_transformer = BayesianGaussianMixture(
+            n_components=self._max_clusters,
+            weight_concentration_prior_type='dirichlet_process',
+            weight_concentration_prior=0.001,
+            n_init=1
+        )
+
+        super()._fit(data)
+        data = super()._transform(data)
```
amontanez24, line 543: What is the reason for this? Is it because of the null transformations

rdt/transformers/numerical.py
```diff
@@ -487,3 +490,160 @@ def _reverse_transform(self, data):
...
+        if data.ndim > 1:
+            data = data[:, 0]
+
+        self._bgm_transformer.fit(data.reshape(-1, 1))
+        self._valid_component_indicator = self._bgm_transformer.weights_ > self._weight_threshold
```
amontanez24, line 548: what does `_valid_component_indicator` actually mean/ what is it? If possible, could we rename it to something more straightforward?

rdt/transformers/numerical.py
```diff
@@ -487,3 +490,160 @@ def _reverse_transform(self, data):
...
+        self._number_of_modes = self._valid_component_indicator.sum()
+        data = data.reshape((len(data), 1))
```
amontanez24: why do we need this?

rdt/transformers/numerical.py
```diff
@@ -487,3 +490,154 @@ def _reverse_transform(self, data):
...
+
+        Args:
+            data (pandas.Series):
+                Data to transform.
+
+        Returns:
+            numpy.ndarray
+        """
+        data = super()._transform(data)
+        if data.ndim > 1:
```
amontanez24, line 562: what is the scenario where this is true? If this is because of the null column, are we guaranteed to get the columns in the correct order?

rdt/transformers/numerical.py
```diff
@@ -487,3 +490,154 @@ def _reverse_transform(self, data):
...
+                p=component_prob_t
+            )
+
+        aranged = np.arange(len(data))
+        normalized = normalized_values[aranged, selected_component].reshape([-1, 1])
+        normalized = np.clip(normalized, -.99, .99)
+        normalized = normalized[:, 0]
+
+        one_hot = np.zeros_like(component_probs)
+        one_hot[np.arange(len(data)), selected_component] = 1
```
amontanez24, line 589: minor: can we reuse `aranged` here?

amontanez24 (commented):
This is looking pretty good. I just had a few questions.

The main thing I am wondering is if the `NullTransformer` should just be used directly. I believe that you are using the `NumericalTransformer` super methods for that reason, and since eventually the `NullTransformer` will be pulled out, maybe it makes sense to just use it directly and not use the `NumericalTransformer` at all

rdt/transformers/numerical.py
fealho: It's because of the NullTransformer. It should always be the last column, since in the NullTransformer code we concatenate the null_column at the end:

https://github.com/sdv-dev/RDT/blob/b7e7f213044b067d6bcd258baffb6c67c26f9328/rdt/transformers/null.py#L125-L126

rdt/transformers/numerical.py
fealho: Yes, the `BayesianGaussianMixture` doesn't accept null values, so we need to call the transform to remove them.

rdt/transformers/numerical.py
fealho: Each "component" is basically each gaussian that the model believes constitutes the data. Each component is assigned a weight, `_bgm_transformer.weights_`, which tells how important it is in generating the data. We eliminate the components which have a very low weight (smaller than some `_weight_threshold`). So `_valid_component_indicator` is an indicator of only the valid components, i.e. which gaussians are relevant.

I'm not sure what else we could call it, since "component" is the usual term used when discussing mixture models.

rdt/transformers/numerical.py
amontanez24: Maybe we should just use the `NullTransformer` directly as it might be removed from the `NumericalTransformer` later on

fealho: Regarding inheriting from the `NumericalTransformer`, we do use the other parameters (like `rounding`) besides the `NullTransformer`. I also believe it makes sense to inherit from it regardless, the `bgmTransformer` only accepts numerical values as input, and so it is a numerical transformer. From an organizational standpoint it should be together with the other numerical transformers.

For the `NullTransformer`, I added it right now because it fails the integration tests (the already existing ones) which require handling on `NaN` values. The change also doesn't seem that complicated, so doing it in the same PR as the when that removes the `NullTransformer` makes more sense to me than removing it now. I'm also not 100% what the code will look like then, so making the changes right now only to have to change it again later could be a problem.
Finally, since I already did it like this, and the code will work the same way either way, I think we should leave as is regardless, and update it together with the `NumericalTransformer` when we get to it (I've spent way too much time on this PR already...).

fealho: @amontanez24 Also, the `CopulasTransformer` doesn't accept the `rounding`, `min_value`, `max_value` parameters. Do you know why?

fealho:
@amontanez24 The `CopulasTransformer` doesn't accept the `rounding`, `min_value`, `max_value` parameters. Do you know why?

amontanez24:
> @amontanez24 The `CopulasTransformer` doesn't accept the `rounding`, `min_value`, `max_value` parameters. Do you know why? If there is some reason, maybe I should remove from the `bgmTransformer` as well.

> @amontanez24 The `CopulasTransformer` doesn't accept the `rounding`, `min_value`, `max_value` parameters. Do you know why? If there is some reason, maybe I should remove from the `bgmTransformer` as well.

I don't think there is a reason. We just didn't add it when we created those parameters

amontanez24 (approved):
I think this is ready!

rdt/transformers/numerical.py
```diff
@@ -156,7 +157,9 @@ def _transform(self, data):
         Returns:
             numpy.ndarray
         """
-        return self.null_transformer.transform(data)
+        if self.nan is not None:
```
csala, line 160: This does not seem correct.

By applying the NullTransformer *awlays*, we guarantee that no Null values make their way to the output *even if there were no nulls in the fit data*, which stops being true after this change.
On the other hand, bear in mind that `self.nan` being `None` does not mean "do not replace nulls", but rather "choose what you use to replace nulls depending on the dtype of the input data"

rdt/transformers/numerical.py
```diff
@@ -487,3 +490,204 @@ def _reverse_transform(self, data):
...
 
         return super()._reverse_transform(data)
+
+
+class BayesGMMTransformer(NumericalTransformer):
+    """Transformer for numerical data using a Bayesian Gaussian Mixture Model.
+
+    The transformation consists of computing the probability that a value came from each
+    component, then sampling a component based on that probability and finally returning a
+    pd.DataFrame of a value and a label encoding, where the lable encoding indicates the
```
csala, line 500: this description seems to be a bit off, since there is no label encoding and the result of the transformation is not a DataFrame  (the output of the `transform` is, but I'd rather talk about columns instead).

rdt/transformers/numerical.py
```diff
@@ -487,3 +490,204 @@ def _reverse_transform(self, data):
...
+            The maximum number of mixture components. Depending on the data, the model may select
+            fewer components (based on the ``weight_threshold``).
+            Defaults to 10.
+        weight_threshold (int, float):
+            The minimum value a component weight can take to be considered a valid component.
+            ``weights_`` under this value will be ignored.
+            Defaults to 0.005.
+    """
+
+    STD_MULTIPLIER = 4
```
csala, line 547: Should this be parametizable as an `__init__` argument, being `4` the default?

rdt/transformers/numerical.py
```diff
@@ -487,3 +490,204 @@ def _reverse_transform(self, data):
...
+    DETERMINISTIC_REVERSE = True
+    COMPOSITION_IS_IDENTITY = False
+
+    def __init__(self, dtype=None, nan='mean', null_column=None, rounding=None,
+                 min_value=None, max_value=None, max_clusters=10, weight_threshold=0.005):
+        super().__init__(dtype=dtype, nan=nan, null_column=null_column, rounding=rounding,
+                         min_value=min_value, max_value=max_value)
+        self._max_clusters = max_clusters
+        self._weight_threshold = weight_threshold
+        self._number_of_modes = None
```
csala, line 558: I would define these attributes at class level, as this is the style that most of the other classes use.

rdt/transformers/numerical.py
csala: I think that `component` is the correct term, but in other places (like the line below) we talk about `modes`, so maybe we want to uniformize this.

On the other hand I agree with @amontanez24 on this not being intuitive based on the attribute name alone. Maybe we should add an `Attributes` section on the `class` docstring where these things are explained?

rdt/transformers/numerical.py
```diff
@@ -487,3 +490,204 @@ def _reverse_transform(self, data):
...
+
+    def get_output_types(self):
+        """Return the output types supported by the transformer.
+
+        Returns:
+            dict:
+                Mapping from the transformed column names to supported data types.
+        """
+        output_types = {
+            'continuous': 'float',
```
csala, line 570: I would use a meaningful name for the outputs rather than a description of their type.

I propose calling them `normalized` (or `value`, like the `NumericalTransformer` and `GaussianCopulaTransformer` do) and `mode` (or `component`, depending on what we end up deciding)

rdt/transformers/numerical.py
```diff
@@ -487,3 +490,204 @@ def _reverse_transform(self, data):
...
+                np.arange(self._number_of_modes),
+                p=component_prob_t
+            )
+
+        aranged = np.arange(len(data))
+        normalized = normalized_values[aranged, selected_component].reshape([-1, 1])
+        normalized = np.clip(normalized, -.99, .99)
+        normalized = normalized[:, 0]
+
+        one_hot = np.zeros_like(component_probs)
```
csala, line 638: Isn't this block of code redundant (with actually a slight chance of error due to collisions on prob=1)?
If I'm reading this correctly, I think we can skip this block and just make `row = [normalized, selected_component]`

rdt/transformers/numerical.py
```diff
@@ -487,3 +490,204 @@ def _reverse_transform(self, data):
...
+        if sigma is not None:
+            normalized = np.random.normal(normalized, sigma)
+
+        normalized = np.clip(normalized, -1, 1)
+        component_probs = np.ones((len(data), self._max_clusters)) * -np.inf
+        component_probs[:, self._valid_component_indicator] = selected_component_probs
+
+        means = self._bgm_transformer.means_.reshape([-1])
+        stds = np.sqrt(self._bgm_transformer.covariances_).reshape([-1])
+        selected_component = np.argmax(component_probs, axis=1)
```
csala, line 661: Why are we taking all this long route through `one_hot` to end up going back to the `selected_component` value?
I think we can skip all if it and just take this directly from the input.

rdt/transformers/numerical.py
```diff
@@ -487,3 +490,204 @@ def _reverse_transform(self, data):
...
+        one_hot[np.arange(data.shape[0]), discrete_column] = 1.0
+        recovered_data = np.concatenate([data[:, :1], one_hot], axis=1)
+        recovered_data = self._reverse_transform_helper(recovered_data, sigma)
+
+        if self.null_transformer and self.null_transformer.creates_null_column():
+            data = np.stack([recovered_data, data[:, -1]], axis=1)  # noqa: PD013
+        else:
+            data = recovered_data
+
+        return pd.Series(super()._reverse_transform(data))
```
csala, line 693: Why do we need to convert to `pd.Series`? I think that the output type from `super()._reverse_transform` should already be the expected one.

csala (changes requested):

rdt/transformers/numerical.py
fealho: I don't think so. I'm not sure if this value really influences the output significantly, we would have to do some experiments to figure out what, if any, effects it has, which I believe should be done in a separate PR. I only added it as a constant because I didn't want a random magic number in the middle of the code.

rdt/transformers/numerical.py
fealho: You're right, I forgot about that.

rdt/transformers/numerical.py
fealho: Yep, that's a good idea.

rdt/transformers/numerical.py
fealho: I agree, I was actually in the middle of adding the descriptions to the docs 👍

rdt/transformers/numerical.py
fealho: You are right, good catch!

rdt/transformers/numerical.py
fealho: Good catch!

csala (approved):
Thanks for addressing the feedback @fealho.
This looks good to go, now!

tests/performance/test_performance.py
```diff
@@ -76,6 +77,9 @@ def _get_performance_test_cases():
         dataset_generators_for_type = dataset_generators.get(data_type, [])
 
         for transformer in transformers_for_type:
+            if transformer == BayesGMMTransformer:
+                continue
```
amontanez24, line 81: Talking to @kveerama , I think it might be a good idea to create a list of `Sandbox` transformers. These by definition will be transformers that have room for improvement in the performance area. Could you make that list and add this transformer to it?

PR #338, opened 2021-12-17
Performance tests get stuck on Unix if multiprocessing is involved

### LINKED ISSUE #337: Performance tests get stuck on Unix if multiprocessing is involved
### Environment Details

* RDT version: 0.6.*
* Python version: Any
* Operating System: Unix

### Error Description

If a transformer uses multiprocessing at all, then the performance tests get stuck when running on Unix systems. The reason for this is that `Unix` systems use a different starting method than `Windows` and `macOS` for processes (see [here](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods)). 

This can be fixed by setting the context to use directly.
```
ctx = get_context('spawn')
p = ctx.Process(...)
```

### Steps to reproduce

- Remove the code blocking the `BayesianGMMTransformer` from being run on the performance tests
- runt the tests using

```
python -m pytest -v ./tests/performance/test_performance.py
```

amontanez24:
Resolves #337

### DIFF
--- a/tests/datasets/categorical.py
+++ b/tests/datasets/categorical.py
@@ -30,7 +30,7 @@ class RandomIntegerGenerator(CategoricalGenerator):
         return {
             'fit': {
                 'time': 1e-05,
-                'memory': 100.0
+                'memory': 400.0
             },
             'transform': {
                 'time': 3e-06,
@@ -57,7 +57,7 @@ class RandomIntegerNaNsGenerator(CategoricalGenerator):
         return {
             'fit': {
                 'time': 1e-05,
-                'memory': 100.0
+                'memory': 400.0
             },
             'transform': {
                 'time': 5e-06,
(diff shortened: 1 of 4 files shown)


csala (approved):

amontanez24:
@csala Using the `forkserver` context does fix the problem but also fails the memory tests. As an added negative, `forkserver` seems to fail on `Windows`.  The new `spawn` context was greatly increasing the memory used for some transformers, so I'm attempting to change the way we calculate memory to be more precise. Let me know what you think

amontanez24:
> @csala Using the `forkserver` context does fix the problem but also fails the memory tests. As an added negative, `forkserver` seems to fail on `Windows`. The new `spawn` context was greatly increasing the memory used for some transformers, so I'm attempting to change the way we calculate memory to be more precise.
> 
> I don't think the last commit is correct, but I'm going to try and see if we can filter out all irrelevant captured memory blocks

It seems like `tracemalloc` finds more allocated memory blocks when using the `spawn` start method then in `fork`. Not exactly sure why but they still seem relevant and the threshold values are now consistent across all platforms, so I think we can go with the new thresholds.

PR #349, opened 2021-12-23
Patch the `BayesGMMTransformer`

### LINKED ISSUE #348: Patching the `BayesGMMTransformer`
There are a few improvements that need to be made in the `BayesGMMTransformer`:

1. The variable `_valid_component_indicator` should actually actually be public, so the `_` should be removed.
2. The integration tests `test_multiple_components` and `test_nulls` fail randomly, so a random seed should be set.
3. Add an integration test where all the values are nulls.

fealho:
Resolve #348.

### DIFF
--- a/tests/integration/transformers/test_numerical.py
+++ b/tests/integration/transformers/test_numerical.py
@@ -182,16 +182,18 @@ class TestBayesGMMTransformer:
         np.testing.assert_array_almost_equal(reverse, data, decimal=1)
 
     def test_all_nulls(self):
+        np.random.seed(10)
         data = pd.DataFrame([np.nan, None] * 50, columns=['col'])
         bgmm_transformer = BayesGMMTransformer()
         bgmm_transformer.fit(data, list(data.columns))
         transformed = bgmm_transformer.transform(data)
 
-        assert isinstance(transformed, pd.DataFrame)
-        assert transformed.shape == (100, 3)
-        assert all(x == 0.0 for x in transformed['col.normalized'])
-        assert all(x == 0.0 for x in transformed['col.component'])
-        assert all(x == 1.0 for x in transformed['col.is_null'])
+        expected = pd.DataFrame({
+            'col.normalized': [0.0] * 100,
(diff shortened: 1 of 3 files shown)


tests/integration/transformers/test_numerical.py
```diff
@@ -180,6 +181,21 @@ def test_nulls(self):
...
+        data = pd.DataFrame([np.nan, None] * 50, columns=['col'])
+        bgmm_transformer = BayesGMMTransformer()
+        bgmm_transformer.fit(data, list(data.columns))
+        transformed = bgmm_transformer.transform(data)
+
+        assert isinstance(transformed, pd.DataFrame)
+        assert transformed.shape == (100, 3)
+        assert all(x == 0.0 for x in transformed['col.normalized'])
+        assert all(x == 0.0 for x in transformed['col.component'])
+        assert all(x == 1.0 for x in transformed['col.is_null'])
```
amontanez24, line 194: Could you make an expected DataFrame like
```
expected_transformed = pd.DataFrame({
    'col.normalized': [0] * 100,
    'col.component': [0] * 100,
    'col.is_null': [1.0] * 100
})
```
and just compare that transformed is equal to it?

tests/integration/transformers/test_numerical.py
```diff
@@ -180,6 +181,21 @@ def test_nulls(self):
...
+
+        reverse = bgmm_transformer.reverse_transform(transformed)
+        np.testing.assert_array_almost_equal(reverse, data)
```
amontanez24, line 197: Is it possible for this to ever fail? If so, should we use a random seed here too?

amontanez24 (approved):
A couple minor comments, but besides that it looks good to me!

tests/integration/transformers/test_numerical.py
fealho: I don't think so (at least it never happened to me), but i'll add it just in case.

pvk-developer (approved):
LGTM!

PR #359, opened 2022-01-25
Fix bug due to `pandas` release

### LINKED ISSUE #358: `pandas 1.4.0` release causes a small error
The [new pandas release](https://pandas.pydata.org/docs/whatsnew/v1.4.0.html#null-values-are-no-longer-coerced-to-nan-value-in-value-counts-and-mode) no longer coerces null-like values in `value_counts` and `mode` to a NaN value. This change causes an error in the following test:

https://github.com/sdv-dev/RDT/blob/40d2667cab950164916e7ac58a317621f99afce8/tests/unit/transformers/test_categorical.py#L145

That happens because the test assumes both `None` and `np.nan` will be treated the same way. However, the categorical transformer uses `value_counts`, which no longer has that behaviour.

https://github.com/sdv-dev/RDT/blob/40d2667cab950164916e7ac58a317621f99afce8/rdt/transformers/categorical.py#L92

This error can be fixed by simply converting the `None` to a `np.nan`.

**Note**: This issue only shows up on python 3.8. Neither py3.7 nor py3.9 cause this to happen, and I have no clue why.

fealho:
Resolves #358 by converting all null-like values to `np.nan` in the `_get_intervals` method.

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -89,6 +89,7 @@ class CategoricalTransformer(BaseTransformer):
             dict:
                 intervals for each categorical value (start, end).
         """
+        data = data.fillna(np.nan)
         frequencies = data.value_counts(dropna=False)
 
         start = 0
--- a/tests/unit/transformers/test_categorical.py
+++ b/tests/unit/transformers/test_categorical.py
@@ -156,7 +156,7 @@ class TestCategoricalTransformer:
             categorical value (start, end).
         """
         # Setup
-        data = pd.Series(['foo', np.nan, None, 'foo', 'foo', 'tar'])
+        data = pd.Series(['foo', np.nan, np.nan, 'foo', 'foo', 'tar'])
 
         # Run
         result = CategoricalTransformer._get_intervals(data)
--- a/tests/unit/transformers/test_categorical.py
+++ b/tests/unit/transformers/test_categorical.py
@@ -156,7 +156,7 @@ class TestCategoricalTransformer:
             categorical value (start, end).
         """
         # Setup
-        data = pd.Series(['foo', np.nan, np.nan, 'foo', 'foo', 'tar'])
+        data = pd.Series(['foo', np.nan, None, 'foo', 'foo', 'tar'])
 
         # Run
         result = CategoricalTransformer._get_intervals(data)


tests/unit/transformers/test_categorical.py
```diff
@@ -156,7 +156,7 @@ def test__get_intervals_nans(self):
             categorical value (start, end).
         """
         # Setup
-        data = pd.Series(['foo', np.nan, None, 'foo', 'foo', 'tar'])
```
amontanez24, line 159: Can we change the expected value instead of this part? I think it's good to know how this function handles `None`

amontanez24 (changes requested):

tests/unit/transformers/test_categorical.py
fealho: We can't really do it, since there are two distinct behaviors for this input, depending on what version of pandas we are running. For `pandas 1.4.0`, the existing code doesn't even work on this input. Notice that the categorical transformer does the following:

https://github.com/sdv-dev/RDT/blob/40d2667cab950164916e7ac58a317621f99afce8/rdt/transformers/categorical.py#L106-L109

So we actually just overwrite the first `np.nan` with the `None`, which is not the intended behavior. If we do want to keep support for multiple types of nan's, we could cast all null objects no `np.nan` right at the start of the function.

amontanez24: It's weird that `value__counts` doesn't treat `None` as `np.nan`, but `pd.isna` will still return `True` for it. My question is, is it possible for `_get_intervals` to receive data with different null types. If so, we have to handle that which means I think we should convert them all before doing the `value_counts`

rdt/transformers/categorical.py
```diff
@@ -89,6 +89,7 @@ def _get_intervals(data):
             dict:
                 intervals for each categorical value (start, end).
         """
+        data = data.fillna(np.nan)
         frequencies = data.value_counts(dropna=False)
 
         start = 0
```
amontanez24, line 95: Are these lines still necessary? 
https://github.com/sdv-dev/RDT/pull/359/files#diff-a773101f2762f769e16ea74de27ab3ebe5db4fba702dfe4197dedd5060517842L106-L107

amontanez24 (approved):
I think maybe we can remove some lines but besides that LGTM!

rdt/transformers/categorical.py
fealho: Yes. Check #360.

PR #361, opened 2022-01-26
Adding all rdt subpackages

amontanez24:


amontanez24:
@csala I was having problems accessing the `rdt.performance` package in private libraries because it said `ModuleNotFoundError: No module named 'rdt.performance.datasets'`. I think this should fix it, but I'm unsure

csala:
> @csala I was having problems accessing the `rdt.performance` package in private libraries because it said `ModuleNotFoundError: No module named 'rdt.performance.datasets'`. I think this should fix it, but I'm unsure

Yes, this would fix it. We just need to make sure that the `addons` packages are not also included. Can you check it?

amontanez24:
> > @csala I was having problems accessing the `rdt.performance` package in private libraries because it said `ModuleNotFoundError: No module named 'rdt.performance.datasets'`. I think this should fix it, but I'm unsure
> 
> Yes, this would fix it. We just need to make sure that the `addons` packages are not also included. Can you check it?

this did include the `addons` package. I changed it to now only include the subpackages for the `performance` module

PR #378, opened 2022-01-31
Update `fit` parameter `columns` to `column`

### LINKED ISSUE #376: Rename `columns` parameter in `fit` for each individual transformer
### Problem Description

All of the individual transformers have a parameter called `columns` that you call during `fit`. Each accepts a lists of columns.

However, none of the transformers work on >1 columns, so the name is misleading.

### Expected behavior

Call the parameter `column` (singular) instead. It should accept a string instead of a list.

### Additional context

This should apply to all individual column transformers. Intended use:

```python
transformer.fit(data, column='column_name')
```

fealho:
Resolve #376.


tests/integration/transformers/test_base.py
```diff
@@ -132,84 +132,3 @@ def _reverse_transform(self, data):
     })
     pd.testing.assert_frame_equal(expected_transform, transformed)
     pd.testing.assert_frame_equal(reverse, data)
-
-
-def test_dummy_transformer_multi_column_input():
```
fealho, line 137: This test is no longer supported, and since the entire purpose of the test is to show that `multi_column_input` works, I didn't fell like it made much sense to update it to conform with the new code (it would end up just repeat testing what the test above already does).

amontanez24 (approved):
This looks good! Can you create a branch called `v1.0.0.dev` and update this PR to point to that branch instead?

pvk-developer (approved):
LGTM!

PR #381, opened 2022-02-02
Removing evaluate_transformers

amontanez24:
The `evaluate_transformers` method isn't being used anywhere and is being replaced by another method to collect performance data.


fealho:
@amontanez24 Could you add some description of why you are deleting the method?

fealho (approved):

PR #382, opened 2022-02-03
Update `LabelEncoder`

### LINKED ISSUE #366: Update `LabelEncodingTransformer` to make it user friendly
### Problem Description

Let's update `LabelEncodingTransformer` to make it more user friendly.

### Expected behavior

Name: `LabelEncoder`
Functionality:
- Change `columns` (see #376 )
- If there are unseen categories in the forward transform: Add a warning and randomly assign labels
- Clip during reverse transform

```python
>>> from transformers.categorical import LabelEncoder

>>> le = LabelEncoder()
>>> le.fit(data, column='categorical_column')

# if the transform contains new categories unseen during fit
>>> le.transform(data_new_categories)
Warning: The data contains new categories ('hello', 'world') that were not in the original data.
Assigning them random values. If you want to model new categories, please fit the transformer
again with the new data.

# if there is out-of-bounds data in the reverse, clip it silently
>>> le.reverse_transform(out_of_bounds_data)

```

fealho:
Resolve #366.


tests/code_style.py
```diff
@@ -13,12 +13,6 @@
 from rdt.transformers.base import BaseTransformer
 
 
-def validate_transformer_name(transformer):
```
fealho, line 16: Since we are renaming the transformer to `LabelEncoder`, I am deleting the quality test which assured all transformers end with the `Transformer` word.

rdt/transformers/categorical.py
```diff
@@ -460,8 +462,16 @@ def _transform(self, data):
...
 
         Returns:
-            numpy.ndarray:
+            pd.Series.
         """
+        unseen_categories = set(pd.unique(data)) - set(self.categories_to_values.keys())
+        if unseen_categories:
+            warn(
+                f'Warning: The data contains new categories {unseen_categories} that were not '
+                'seen in the original data. Assigning them NaN values. If you want to model '
```
fealho, line 471: I changed the message to say `Assigning them NaN values` instead of `Assigning them random values`, since that's the behavior of the code.

rdt/transformers/categorical.py
```diff
@@ -460,8 +462,16 @@ def _transform(self, data):
...
+                'new categories, please fit the transformer again with the new data.'
+            )
+
         return pd.Series(data).map(self.categories_to_values)
 
     def _reverse_transform(self, data):
```
fealho, line 477: The reverse transform already clips the data silently:

https://github.com/sdv-dev/RDT/blob/7848667a4768a43deaafbdac3e24c3c46ec895a6/rdt/transformers/categorical.py#L505

fealho (commented):
@npatki I implemented as you described, with the caveats described below.

HISTORY.md
```diff
@@ -200,7 +200,7 @@ high number of columns or a high number of categorical values when using one hot
 
 ### Issues closed
 
-* `Boolean`, `Datetime` and `LabelEncoding` transformers fail with 2D `ndarray` - Issue [#160](https://github.com/sdv-dev/RDT/issues/160) by @pvk-developer
+* `Boolean`, `Datetime` and `LabelEncodingTransformer` transformers fail with 2D `ndarray` - Issue [#160](https://github.com/sdv-dev/RDT/issues/160) by @pvk-developer
```
amontanez24, line 203: Should be `LabelEncoder` now right?

rdt/transformers/categorical.py
amontanez24: I don't think it actually says in the issue but I believe we want to change the behavior to return a random value now instead. Is that right @npatki

HISTORY.md
fealho: I don't think so, since it's in the `HISTORY.md` file, so the naming should be as was at the time, right?

rdt/transformers/categorical.py
npatki: Ah yes, @fealho I implied the change in the warning message instead of explicitly stating it outright. We want to assign random values.

Rationale:
(a) not all ML methods can deal with null values; intended usage is not to introduce more nulls
(b) categorical transformers are treating nulls as a separate category; an unexpected category != null in that sense

rdt/transformers/categorical.py
```diff
@@ -455,13 +457,35 @@ def _fit(self, data):
...
+
+            warn(
+                f'Warning: The data contains new categories {unseen_categories} that were not '
+                'seen in the original data. Assigning them random values. If you want to model '
+                'new categories, please fit the transformer again with the new data.'
+            )
+
+            for unseen_category in unseen_categories:
+                self.categories_to_values[unseen_category] = \
+                    np.random.randint(0, len(self.values_to_categories) - 1)
```
amontanez24, line 487: Instead of doing this, you could pass a function into `map` that returns the value in the `self.categories_to_values` dictionary if it exists, and a random int if it doesn't. Something like
```
map_function = lambda x: self.categories_to_values.get(x, np.random.randint(0, len(self.values_to_categories) - 1))
return pd.Series(data).map(map_function)
```

rdt/transformers/categorical.py
```diff
@@ -455,13 +457,35 @@ def _fit(self, data):
...
                 Data to transform.
 
         Returns:
-            numpy.ndarray:
+            pd.Series.
         """
+        # Find the data categories that were not seen during fitting. Convert all nan's
+        # to np.nan to avoid having nan values with different ids.
+        unique_data = {np.nan if pd.isna(x) else x for x in pd.unique(data)}
+        categories = {np.nan if pd.isna(x) else x for x in self.categories_to_values.keys()}
```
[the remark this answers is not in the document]
fealho: You can't convert things to `pd.Series`, otherwise it doesn't interpret the the nan's properly:
```python3
set(pd.Series([np.nan])) - set(pd.Series([np.nan]))  # returns {nan}
set([np.nan]) - set([np.nan])  # returns {}
```

rdt/transformers/categorical.py
fealho: Hmm, we would not be returning the same value for each unseen category though, so if an unseen category shows up twice, each can get assigned a different value.

rdt/transformers/categorical.py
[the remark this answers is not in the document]
fealho: If we really wanted to vectorize things though, I guess we could do something like this:

```python3
null = pd.isna(data)
unique_data = list(pd.unique(data[~null]))
if null.any():
    unique_data.append(np.nan)
```
Then do the same for `categories` and subtract them.

Untitled.ipynb
```diff
@@ -0,0 +1,101 @@
+{
```
amontanez24, line 1: please remove this file

rdt/transformers/categorical.py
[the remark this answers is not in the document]
amontanez24: If that's faster by a significant amount I'd suggest going with that

rdt/transformers/categorical.py
```diff
@@ -455,14 +457,39 @@ def _fit(self, data):
...
+            if len(self.categories_to_values) == 0:
+                raise ValueError('No categories have been fitted.')
+
+            warn(
+                f'Warning: The data contains new categories {unseen_categories} that were not '
+                'seen in the original data. Assigning them random values. If you want to model '
+                'new categories, please fit the transformer again with the new data.'
+            )
+
+        mapped_data = pd.Series(data).map(self.categories_to_values)
```
amontanez24, line 485: I think map takes a function. Why don't we just pass it in here and have the function return the value in the dict if it's present, and a random value otherwise? Do we need to apply two separate functions here?

fealho: Unfortunately, yeah. The actual solution would probably be not to allow nan's as keys of a dictionary (since they don't necessarily have the same id, you can't actually get the nan value back, e.g. `d = {float('nan')}; d[float('nan')]` throws a keyerror).

So basically, we can't just pass your function because even if you fit on the `nans` (so they are present in the `self.categories_to_values` dict) when you try to `get` it thinks the `nan` is not there. There may be some shenanigans which allow you to access them without iterating through the whole dictionary (that was what I was trying to do in the `Untitled.ipynb` notebook which I accidentally submitted), but I couldn't figure out how.

So yeah, best I could do to keep the vectorization spirit was map the seen data first, like it was doing before, and then apply the random values where the nans are.

csala (changes requested):

rdt/transformers/categorical.py
```diff
@@ -455,14 +477,39 @@ def _fit(self, data):
...
+            pd.Series.
         """
-        return pd.Series(data).map(self.categories_to_values)
+        # Find the data categories that were not seen during fitting. Convert all nan's
+        # to np.nan to avoid having nan values with different ids.
+        unique_data = _select_unique_categories(pd.unique(data))
+        categories = _select_unique_categories(pd.Series(self.categories_to_values.keys()))
+        unseen_categories = unique_data - categories
+        if unseen_categories:
+            if len(self.categories_to_values) == 0:
```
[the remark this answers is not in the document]
fealho: @npatki indicated that we shouldn't be worrying about fitting before transforming in this PR. In the future we may go through the models again to assert you don't transform without fitting, but for now I shouldn't make any changes about it.

In this PR though, since it hasn't been discussed what the behavior should be if we fit on an empty dataframe, I added that guarantee there.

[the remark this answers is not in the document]
amontanez24: I wouldn't worry about that in this PR. I also am not sure that's worth raising an error for.

rdt/transformers/categorical.py
```diff
@@ -455,14 +459,39 @@ def _fit(self, data):
...
+            `NotFittedError` when no categories have been fitted and `data` is not empty.
         """
-        return pd.Series(data).map(self.categories_to_values)
+        # Find the data categories that were not seen during fitting. Convert all nan's
+        # to np.nan to avoid having nan values with different ids.
+        mapped = data.fillna(np.nan).map(self.categories_to_values)
+        is_null = mapped.isna()
+        if is_null.any():
+            if len(self.categories_to_values) == 0:
+                raise NotFittedError('No categories have been fitted.')
```
amontanez24, line 480: Can you move this to the first thing? I think we want to raise the error whether or not there are nulls right?

rdt/transformers/categorical.py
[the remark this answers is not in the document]
npatki: +1 In the case where a user accidentally tries to transform before calling fit, we should handle it consistently within all transformers. So that would be a separate change.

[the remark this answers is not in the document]
csala: I agree that this should be covered in a generic way, probably in the `BaseTransformer` (BTW, the `HyperTransformer already covers it!).
In any case, I suggest changing the line to `if not self.categories_to_values` anyway, since in general `if not <iterable>` is the recommended pythonic shortcut for `if len(<iterable>) == 0` (I would even expect PyLint to fail on that one), and it is also slightly more efficient.

csala (commented):
This looks good now.
I just added two minor comments.
I also agree with @amontanez24 recommendation of moving the `no categories` check up.

rdt/transformers/categorical.py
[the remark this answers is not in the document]
fealho: I'm deleting this check. It seems to be causing too much confusion for something that should be deleted in the future either way.

rdt/transformers/categorical.py
fealho: As mentioned above, I'm deleting the check.

amontanez24 (approved):
Looks good now!

Issue #362, opened 2022-01-26
Update names & functionality for handling missing values


npatki:
### Problem Description

Let's update the usage of missing value handling to make it more user friendly. The update would apply to all transformers that handle missing values. Currently, these are:

- `BooleanTransformer`
- `DateTimeTransformer` (+ variants)
- `NumericalTransformer` (+ variants)
- `NullTransformer`

### Expected behavior

1. `nan` becomes `missing_value_replacement`, with updated defaults
```python
# (default) do not replace missing values
nt = NumericalTransformer()

# replace with a static number
nt = NumericalTransformer(missing_value_replacement=0.00)

# replace using a method
nt = NumericalTransformer(missing_value_replacement='mode')
```

2. `null_column` becomes `model_missing_values` that can only accept True/False values; updated functionality & defaults
```python
# (default = False) do not create a new column
nt = NumericalTransformer()
nt = NumericalTransformer(model_missing_values=False)

# True creates a missing column if there is missing data
nt = NumericalTransformer(model_missing_values=True)
nt.fit(data, columns=['column_with_missing_values'])

# otherwise do not create a column and warn the user
nt = NumericalTransformer(model_missing_values=True)
nt.fit(data, columns=['never_null_column'])
Guidance: There are no missing values in column 'never_null_column'. Extra column not created.
 ```

**Comment by amontanez24** [comment]
A couple points:
1. This issue will probably have to be completed at the same as #372 since all these transformers call the `NullTransformer` behind the scenes, so they will crash if the parameters for the `NullTransformer` change and theirs don't
2. Since each transformer will be using the same parameters to define how the `NullTransformer` will be used, perhaps it makes sense to move them to the `BaseTransformer` class. On top of that, we may want to call the `NullTransformer` from within the `BaseTransformer's` `fit`, `transform` and `reverse_transform` methods. Something like

```
def fit(self, data, columns):
    ...
    self.null_Transformer = NullTransformer(missing_value_replacement=self.missing_value_replacement, model_missing_values = self.model_missing_values)
    self._fit(columns_data)
    self.null_transformer.fit(columns_data)
    
def transform(self, data):
    ...
    transformed_data = self._transform(columns_data)
    transformed_data = self.null_transformer.transform(data)

def reverse_transform(self, data):
    ...
    columns_data = self.null_transformer.reverse_transform(columns_data)
    reversed_data = self._reverse_transform(columns_data)
```

PR #384, opened 2022-02-07
Update NullTransformer to make it user friendly

### LINKED ISSUE #372: Update `NullTransformer` to make it user friendly
### Problem Description

Let's update `NullTransformer` to make it more user friendly.

### Expected behavior

Functionality:

- Missing value handling (see #362 )
- Remove `copy`

```python
>>> from transformers.null import NullTransformer

# copy is always false
>>> nt = NullTransformer()
>>> nt.fit(data, column='column_will_nulls')
```

pvk-developer:
### NullTransformer
With the current `PR` the `NullTransformer` contains the following changes:
* Always works with a copy of the data (no changes occur to the original data).
* Adapt unit tests and integration tests.
* New user friendly names for the fill value / nan and null column:
    * `fill_value` or `nan` becomes -> `missing_value_replacement`.
    * `null_column` becomes -> `model_missing_values`.

#### `missing_value_replacement`
By default this value will be `None` which will not replace the values. From python perspective this will return `np.nan` which should be the representation of `Null`.

The second strategy is the user specifying a string `mode` or `mean` , which will use the `mode` of the data or the `mean` value (if possible) as a replacement of the `null` values that are in the data.

The third strategy is the user specifying a value for it which will be used to fill the null values with.

#### `model_missing_values`
By default this value will be `False`. If set to `True` it will generate a new column representing if there was a missing value or not. In the case that the data had no missing values this column will not be generated.

Resolves #372 , #362 , #376

### DIFF
--- a/rdt/performance/performance.py
+++ b/rdt/performance/performance.py
@@ -86,7 +86,6 @@ def evaluate_transformer_performance(transformer, dataset_generator, verbose=Fal
         pandas.DataFrame:
             The performance test results.
     """
-
     transformer_args = TRANSFORMER_ARGS.get(transformer.__name__, {})
     transformer_instance = transformer(**transformer_args)
 
--- a/rdt/performance/performance.py
+++ b/rdt/performance/performance.py
@@ -9,6 +9,42 @@ from rdt.transformers import BaseTransformer
 
 DATASET_SIZES = [1000, 10000, 100000]
 
+# Additional arguments for transformers
+TRANSFORMER_ARGS = {
+    'BooleanTransformer': {
+        'missing_value_replacement': -1,
+        'model_missing_values': True
+    },
+    'DatetimeTransformer': {
+        'missing_value_replacement': 'mean',
+        'model_missing_values': True
+    },
+    'DatetimeRoundedTransformer': {
+        'missing_value_replacement': 'mean',
+        'model_missing_values': True
+    },
+    'NumericalTransformer': {
+        'missing_value_replacement': 'mean',
+        'model_missing_values': True
+    },
+    'NumericalRoundedBoundedTransformer': {
+        'missing_value_replacement': 'mean',
+        'model_missing_values': True
+    },
+    'NumericalBoundedTransformer': {
+        'missing_value_replacement': 'mean',
+        'model_missing_values': True
+    },
+    'GaussianCopulaTransformer': {
+        'missing_value_replacement': 'mean',
+        'model_missing_values': True
+    },
+    'BayesGMMTransformer': {
+        'missing_value_replacement': 'mean',
+        'model_missing_values': True
+    },
+}
+
 
 def _get_dataset_sizes(data_type):
     """Get a list of (fit_size, transform_size) for each dataset generator.
@@ -50,7 +86,9 @@ def evaluate_transformer_performance(transformer, dataset_generator, verbose=Fal
         pandas.DataFrame:
             The performance test results.
     """
-    transformer_instance = transformer()
+
+    transformer_args = TRANSFORMER_ARGS.get(transformer.__name__, {})
+    transformer_instance = transformer(**transformer_args)
 
     sizes = _get_dataset_sizes(dataset_generator.DATA_TYPE)
 
--- a/rdt/transformers/boolean.py
+++ b/rdt/transformers/boolean.py
@@ -47,7 +47,7 @@ class BooleanTransformer(BaseTransformer):
         output_types = {
             'value': 'float',
         }
-        if self.null_transformer and self.null_transformer.creates_model_missing_values():
+        if self.null_transformer and self.null_transformer.models_missing_values():
             output_types['is_null'] = 'float'
 
         return self._add_prefix(output_types)
@@ -63,7 +63,7 @@ class BooleanTransformer(BaseTransformer):
             self.missing_value_replacement,
             self.model_missing_values
         )
-        self.null_transformer.fit(data, self.get_input_column())
+        self.null_transformer.fit(data)
 
     def _transform(self, data):
         """Transform boolean to float.
--- a/rdt/transformers/boolean.py
+++ b/rdt/transformers/boolean.py
@@ -16,15 +16,15 @@ class BooleanTransformer(BaseTransformer):
     Null values are replaced using a ``NullTransformer``.
 
     Args:
-        missing_value_replacement (int or None):
-            Replace null values with the given value. If ``None``, do not replace them.
-            Defaults to ``-1``.
+        missing_value_replacement (object or None):
+            Indicate what to do with the null values. If an object is given, replace them
+            with the given value. If the strings ``'mean'`` or ``'mode'`` are given, replace
+            them with the corresponding aggregation. If ``None`` is given, do not replace them.
+            Defaults to ``None``.
         model_missing_values (bool):
             Whether to create a new column to indicate which values were null or not.
-            If ``None``, only create a new column when the fit data contains null values.
-            If ``True``, always create the new column whether there are null values or not.
-            If ``False``, do not create the new column.
-            Defaults to ``None``.
+            If ``True``, create the new column if there are null values.
+            If ``False``, do not create the new column. Defaults to ``False``.
     """
 
     INPUT_TYPE = 'boolean'
--- a/rdt/transformers/boolean.py
+++ b/rdt/transformers/boolean.py
@@ -33,7 +33,7 @@ class BooleanTransformer(BaseTransformer):
 
     null_transformer = None
 
-    def __init__(self, missing_value_replacement=-1, model_missing_values=True):
+    def __init__(self, missing_value_replacement=None, model_missing_values=False):
         self.missing_value_replacement = missing_value_replacement
         self.model_missing_values = model_missing_values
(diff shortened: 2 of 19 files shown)


rdt/transformers/null.py
```diff
@@ -1,87 +1,79 @@
...
 
-        if fill_value == 'mode':
+        if missing_value_replacement == 'mode':
             return data.mode(dropna=True)[0]
 
-        return fill_value
+        return missing_value_replacement
 
-    def fit(self, data):
+    def fit(self, data, column):
```
amontanez24, line 76: If the only reason we are passing this is for the warning message then I'm not sure we need it. Couldn't we get the column name by doing `data.name`?

rdt/transformers/null.py
```diff
@@ -1,87 +1,79 @@
...
-        self.fill_value = fill_value
-        self.null_column = null_column
-        self.copy = copy
+    def __init__(self, missing_value_replacement=None, model_missing_values=False):
+        self.missing_value_replacement = missing_value_replacement
+        self.model_missing_values = model_missing_values
+        self._model_missing_values = model_missing_values
 
-    def creates_null_column(self):
+    def creates_model_missing_values(self):
```
amontanez24, line 43: minor: I'd call this `models_missing_values`

rdt/transformers/boolean.py
```diff
@@ -16,10 +16,10 @@ class BooleanTransformer(BaseTransformer):
     Null values are replaced using a ``NullTransformer``.
 
     Args:
-        nan (int or None):
+        missing_value_replacement (int or None):
             Replace null values with the given value. If ``None``, do not replace them.
             Defaults to ``-1``.
```
amontanez24, line 21: This isn't true anymore right?

amontanez24 (changes requested):
Looking good. I left a few comments

rdt/performance/performance.py
```diff
@@ -9,6 +9,42 @@
 
 DATASET_SIZES = [1000, 10000, 100000]
 
+# Additional arguments for transformers
+TRANSFORMER_ARGS = {
+    'BooleanTransformer': {
+        'missing_value_replacement': -1,
```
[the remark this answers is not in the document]
pvk-developer: I used the `default` that was before on the `BooleanTransformer` . I think that it's okay to leave it is, however in the following changes to the baseclass we can change this to be the same 
````python
null_transformer_args = { 
    missing_value_replacement: 'mean'
    model_missing_value: True,
}
if transformer.supports_null(): # or something like that ?
    transformer_instance = transformer(**null_transformer_args)
````

File: tests/unit/transformers/test_null.py [hunk 5]
```diff
@@ -255,28 +246,28 @@ def test_fit_null_column_none_and_nulls(self):
...
+        assert transformer._missing_value_replacement == 1.5
 
-    def test_fit_null_column_none_and_no_nulls(self):
+    def test_fit_model_missing_values_none_and_no_nulls(self):
         """Test fit when null column is none and there are NO nulls.
 
-        If there are no nulls in the data and null_column was given as None,
-        then the _null_column attribute should be set to False.
-        Also validate that the null attribute and the _fill_value attributes
+        If there are no nulls in the data and model_missing_values was given as None,
```
fealho, line 268: as False, not none.

tests/unit/transformers/test_numerical.py
```diff
@@ -1228,7 +1328,7 @@ def test__reverse_transform_null_column_none(self):
         ct = GaussianCopulaTransformer()
         ct._univariate = Mock()
         ct._univariate.ppf.return_value = np.array([0.0, 1.0, 2.0, 1.0])
-        ct.null_transformer = NullTransformer(None, null_column=None)
+        ct.null_transformer = NullTransformer(None, model_missing_values=None)
```
fealho, line 1331: not None

rdt/transformers/null.py
```diff
@@ -1,85 +1,74 @@
...
-            return 0
+        missing_value_replacement = self.missing_value_replacement
 
-        if fill_value is None:
-            if pd.api.types.is_numeric_dtype(data):
-                fill_value = 'mean'
-            else:
-                fill_value = 'mode'
+        if missing_value_replacement is None:
+            return np.nan
```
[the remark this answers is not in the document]
pvk-developer: Nice catch, I made a lot of changes, this should return `None` and when we do `fillna` we check if this is not `None`  to actually proceed with the replacement 👍🏻

rdt/transformers/datetime.py
```diff
@@ -15,17 +15,15 @@ class DatetimeTransformer(BaseTransformer):
     Null values are replaced using a ``NullTransformer``.
 
     Args:
-        nan (int, str or None):
-            Indicate what to do with the null values. If an integer is given, replace them
+        missing_value_replacement (object or None):
+            Indicate what to do with the null values. If an object is given, replace them
```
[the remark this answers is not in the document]
pvk-developer: Yes in my mind it made sense to put `str, int, float or None` but then of course `False` or `True` could be an option so I decided to use `object` as a generic word to whatever is being passed will be used to replace the `nan` values.

rdt/transformers/null.py
```diff
@@ -93,15 +82,17 @@ def fit(self, data):
...
-        if self.null_column is None:
-            self._null_column = self.nulls
-        else:
-            self._null_column = self.null_column
+        self._missing_value_replacement = self._get_missing_value_replacement(data)
+        if not self.nulls and self._model_missing_values:
+            self._model_missing_values = False
+            guidance_message = (
+                f'Guidance: There are no missing values in column {data.name}. '
+                'Extra column not created.'
```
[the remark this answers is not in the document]
pvk-developer: I mention it on the docstring but let me think of a better message.

tests/unit/transformers/test_numerical.py
```diff
@@ -129,7 +133,8 @@ def test__learn_rounding_digits_less_than_15_decimals(self):
         should be returned.
 
         Input:
-        - An array that contains floats with a maximum of 3 decimals and a NaN.
+        - An array that contains floats with a maximum of 3 decimals and a
+          missing_value_replacement.
```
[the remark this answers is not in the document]
pvk-developer: 😅 make sure to use case sensitive when replacing 🤦🏻‍♂️

rdt/transformers/boolean.py
```diff
@@ -16,15 +16,16 @@ class BooleanTransformer(BaseTransformer):
...
-            Replace null values with the given value. If ``None``, do not replace them.
-            Defaults to ``-1``.
-        null_column (bool):
-            Whether to create a new column to indicate which values were null or not.
-            If ``None``, only create a new column when the fit data contains null values.
-            If ``True``, always create the new column whether there are null values or not.
-            If ``False``, do not create the new column.
+        missing_value_replacement (object or None):
+            Indicate what to do with the null values. If an object is given, replace them
+            with the given value. If the strings ``'mean'`` or ``'mode'`` are given, replace
```
amontanez24, line 21: I guess we should remove `mean` since that doesn't really make sense for booleans

amontanez24 (approved):
One minor comment but once you finish addressing that and the other comments it should be good to go!

tests/unit/transformers/test_numerical.py
fealho, line 137: After you put the `NaN`'s back I think it's good to go.

tests/unit/transformers/test_numerical.py
```diff
@@ -594,8 +648,8 @@ def test__reverse_transform_rounding_none_with_nulls_dtype_int(self):
         Input:
         - 2d Array of multiple float values with decimals and a column setting at least 1 null.
         Output:
-        - First column of the input array rounded, replacing the indicated value with a Nan,
```
fealho, line 597: here.

tests/unit/transformers/test_numerical.py
```diff
@@ -177,13 +182,13 @@ def test__learn_rounding_digits_negative_decimals_integer(self):
 
         assert output == -1
 
-    def test__learn_rounding_digits_all_nans(self):
-        """Test the _learn_rounding_digits method with data that is all NaNs.
```
fealho, line 181: here

tests/unit/transformers/test_numerical.py
```diff
@@ -177,13 +182,13 @@ def test__learn_rounding_digits_negative_decimals_integer(self):
...
+    def test__learn_rounding_digits_all_missing_value_replacements(self):
+        """Test the _learn_rounding_digits method with data that is all missing_value_replacements.
 
-        If the data is all NaNs, expect that the output is None.
```
fealho, line 183: here

tests/unit/transformers/test_numerical.py
```diff
@@ -177,13 +182,13 @@ def test__learn_rounding_digits_negative_decimals_integer(self):
...
+        If the data is all missing_value_replacements, expect that the output is None.
 
         Input:
-        - An array of NaNs.
```
fealho, line 186: here

fealho (approved):
A few more NaN's to change back.

PR #386, opened 2022-02-08
Update `FrequencyEncoder`

### LINKED ISSUE #364: Update `CategoricalTransformer` to make it user friendly
### Problem Description

Let's update `CategoricalTransformer` to make it more user friendly.

### Expected behavior

Name: `FrequencyEncoder`
Functionality:
- Change `columns` (see #376 )
- Change `fuzzy` to `add_noise`
- Remove `clip`
- Add warning for unseen categories

```python
>>> from transformers.categorical import FrequencyEncoder
# (default) No noise
>>> fe = FrequencyEncoder() 

>>> fe = FrequencyEncoder(add_noise=True)
>>> fe.fit(data, column='categorical_column')

# if the transform contains new categories unseen during fit
>>> fe.transform(data_new_categories)
Warning: The data contains new categories ('hello', 'world') that were not in the original data.
Assigning them random values. If you want to model new categories, please fit the transformer
again with the new data.

# reverse transform: always clip without warning (ie never use modulo)
>>> fe.reverse_transform(model_data)
```

fealho:
Resolve #364.


rdt/transformers/categorical.py
```diff
@@ -185,16 +181,6 @@ def _transform(self, data):
 
         return self._transform_by_row(data)
 
-    def _normalize(self, data):
```
amontanez24, line 188: @npatki I remember the doc said something about allowing for normalization as an option, but the issue doesn't mention it. Are we just always going to clip now?

amontanez24 (commented):
This looks good! Just pending on what @npatki says

rdt/transformers/categorical.py
npatki: Yes always clip. If the model is generating too many out-of-bounds values, that is an issue with the model.

Normalizing can have other unintended consequences.

rdt/transformers/categorical.py
```diff
@@ -182,20 +172,28 @@ def _transform(self, data):
...
-        if len(self.means) < len(data):
-            return self._transform_by_category(data)
+        fit_categories = pd.Series(self.intervals.keys())
+        has_nan = pd.isna(fit_categories).any()
 
-        return self._transform_by_row(data)
+        def find_unseen_data(value):
+            if np.isin(value, fit_categories) or (pd.isna(value) and has_nan):
+                return False
+            return True
```
amontanez24, line 181: I think you can vectorize this like
```
unseen_data = data[~(data.isin(fit_categories) | (pd.isna(data) & has_nan))]
```

fealho: Oh, I missed this 😅, thanks!

rdt/transformers/categorical.py
```diff
@@ -182,21 +172,23 @@ def _transform(self, data):
         Returns:
             numpy.ndarray
         """
+        fit_categories = pd.Series(self.intervals.keys())
+        has_nan = pd.isna(fit_categories).any()
+        unseen_indexes = ~(data.isin(fit_categories) | (pd.isna(data) & has_nan))
+        if unseen_indexes.any():
+            unseen_categories = set(data[unseen_indexes])
+            warnings.warn(
+                f'Warning: The data contains new categories {unseen_categories} that were not '
```
pvk-developer, line 181: Should we limit the `unseen_categories` message to the first `5` since otherwise we will flood the log... ?
Also, do we want to keep the `Warning:` ? The message output ends up being like this:
```
UserWarning: Warning: The data contains new categories {'hello', 'world'} that were not seen in the original data. Assigning them random values. If you want to model new categories, please fit the transformer again with the new data.
```

Issue #369, opened 2022-01-26
Update `DateTimeRoundedTransformer` to make it user friendly


npatki:
### Problem Description

Let's update `DateTimeRoundedTransformer` to make it more user friendly.

### Expected behavior

Name: `OptimizedTimestampEncoder`
Functionality:
- Missing value handling (#362 )
- Change `columns` (see #376 )
- Add `format`

```python
>>> from transformers.datetime import OptimizedTimestampEncoder

# never strips the constant
>>> ote = OptimizedTimestampEncoder()

# format specifies how the data is written (eg. in this case 'Jan 01 2021')
>>> ote = OptimizedTimestampEncoder(format='%b %d %Y')
>>> ote.fit(data, column='datetime_column')
>>> transformed = ote.transform(data)
# reverse back into the correctly formatted string
>>> reversed_data = ote.reverse_transform(transformed)
```

PR #388, opened 2022-02-10
Updating Datetime Transformers

### LINKED ISSUE #368: Update `DateTimeTransformer` to make it user friendly
### Problem Description

Let's update `DateTimeTransformer` to make it more user friendly.

### Expected behavior

Name: `UnixTimestampConverter`
Functionality:
- Missing value handling (#362 )
- Remove `strip_constant`
- Reverse to the correctly formatted string
- Change `columns` to `column`

```python
>>> from transformers.datetime import UnixTimestampConverter

# never strips the constant
>>> utc = UnixTimestampConverter()

# format specifies how the data is written (eg. in this case 'Jan 01 2021')
>>> utc = UnixTimestampConverter(format='%b %d %Y')
>>> utc.fit(data, column='datetime_column')
>>> transformed = utc.transform(data)
# reverse back into the correctly formatted string
>>> reversed_data = utc.reverse_transform(transformed)
```

amontanez24:
resolves #368 and #369

### DIFF
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -71,16 +71,6 @@ class UnixTimestampEncoder(BaseTransformer):
 
         return self._add_prefix(output_types)
 
-    def _find_divider(self, transformed):
-        self.divider = 1
-        multipliers = [10] * 9 + [60, 60, 24]
-        for multiplier in multipliers:
-            candidate = self.divider * multiplier
-            if (transformed % candidate).any():
-                break
-
-            self.divider = candidate
-
     def _convert_to_datetime(self, data):
         if data.dtype == 'object':
             try:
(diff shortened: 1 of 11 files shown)


rdt/transformers/datetime.py
```diff
@@ -25,13 +25,7 @@ class DatetimeTransformer(BaseTransformer):
...
             there are null values. If ``False``, do not create the new column even if there
             are null values. Defaults to ``False``.
-        strip_constant (bool):
-            Whether to optimize the output values by finding the smallest time unit that
-            is not zero on the training datetimes and dividing the generated numerical
-            values by the value of the next smallest time unit. This, a part from reducing the
-            orders of magnitued of the transformed values, ensures that reverted values always
-            are zero on the lower time units.
-        format (str):
+        datetime_format (str):
```
pvk-developer, line 28: In the issue says `format` do we want to change it to `datetime_format` ?

pvk-developer (approved):
Overall looks good, just minor (maybe I'm out of context) the issue says `format` and not `daytime_format` , which one should we keep ?

fealho (approved):
Looks good 💯

rdt/transformers/datetime.py
amontanez24: I don't think I can make it `format` because `format` is a special python function

PR #394, opened 2022-02-16
Handle same name transformers

### LINKED ISSUE #352: Improve RDT to be able to handle transformers with the same name
### Problem Description

It is possible down the line the some RDTs will have the same name, since we are moving away from the `Transformer` suffix naming convention.

In the `__init__` file for the `transformers` module, the transformers are loaded into a dictionary that uses their names as the keys. The problem is that it doesn't use the full module + class name, so it is possible for there to be overlap. 
https://github.com/sdv-dev/RDT/blob/dbc550fc5e6cd62a032ab412aa8e2687bac82647/rdt/transformers/__init__.py#L45-L48

### Expected behavior

Methods that use the `TRANSFORMERS` dict should be able to load transformer classes that have the same name but belong to different modules

fealho:
Resolve #352.

### DIFF
--- a/rdt/transformers/__init__.py
+++ b/rdt/transformers/__init__.py
@@ -40,12 +40,10 @@ def _import_addons():
 
 _import_addons()
 
-TRANSFORMERS = {
-    transformer.__name__: transformer
-    for transformer in BaseTransformer.get_subclasses()
-}
+TRANSFORMERS = defaultdict(list)
+for transformer in BaseTransformer.get_subclasses():
+    TRANSFORMERS[__name__].append(transformer)
 
-globals().update(TRANSFORMERS)
 __all__.extend(TRANSFORMERS.keys())
 
 DEFAULT_TRANSFORMERS = {
@@ -62,7 +60,7 @@ def get_transformer_class(transformer):
     """Return a ``transformer`` class from a ``str``.
 
     Args:
-        transforemr (str):
+        transformer (str):
             Python path or transformer's name.
 
     Returns:
@@ -70,10 +68,16 @@ def get_transformer_class(transformer):
             BaseTransformer subclass class object.
     """
     if len(transformer.split('.')) == 1:
-        return TRANSFORMERS[transformer]
+        if len(TRANSFORMERS[transformer]) == 1:
+            return TRANSFORMERS[transformer][0]
+        if len(TRANSFORMERS[transformer]) > 1:
+            raise ValueError(
+                f'There are multiple transformers named {transformer}. Please '
+                f'specify which one you want: {TRANSFORMERS[transformer]}.'
+            )
 
     package, name = transformer.rsplit('.', 1)
-    return TRANSFORMERS.get(name, getattr(importlib.import_module(package), name))
+    return getattr(importlib.import_module(package), name)
 
 
 def get_transformer_instance(transformer):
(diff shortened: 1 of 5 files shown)


rdt/transformers/__init__.py
```diff
@@ -35,18 +46,48 @@ def _import_addons():
...
 TRANSFORMERS = {
-    transformer.__name__: transformer
+    get_transformer_name(transformer): transformer
     for transformer in BaseTransformer.get_subclasses()
 }
 
-globals().update(TRANSFORMERS)
-__all__.extend(TRANSFORMERS.keys())
+__all__.extend({
+    transformer.__name__
```
amontanez24, line 88: why doesn't this use the `get_transformer_name` method? What happens if this list has multiple entries that are the same?

rdt/transformers/__init__.py
```diff
@@ -35,18 +46,48 @@ def _import_addons():
...
+
 TRANSFORMERS = {
-    transformer.__name__: transformer
+    get_transformer_name(transformer): transformer
     for transformer in BaseTransformer.get_subclasses()
 }
 
-globals().update(TRANSFORMERS)
-__all__.extend(TRANSFORMERS.keys())
+__all__.extend({
```
amontanez24, line 87: Aren't these names already added to `__all__` on line 55?

fealho: I forgot to delete this line, thanks!

rdt/transformers/__init__.py
```diff
@@ -35,18 +46,44 @@ def _import_addons():
...
                 module = transformer.rsplit('.', 1)[0]
                 if module not in sys.modules:
-                    importlib.import_module(module)
+                    module = importlib.import_module(module)
+                    for name, obj in inspect.getmembers(module):
+                        # Inject the addon transformer into this scope
+                        if inspect.isclass(obj) and issubclass(
+                                obj, BaseTransformer) and obj is not BaseTransformer:
+                            globals()[name] = obj
+                            __all__.append(name)
```
amontanez24, line 55: What happens here if there are addons with the same name? Is it possible for us to import them by module and name?

fealho: With the current code, if there are two add-ons with the same name, only the last add-on would be available to be imported from `rdt.transformers`. However, you could still import it by the full name if you wanted to (with `rdt.transformers.addons.X`).

amontanez24: Should we just import them by full name here? Idk if it makes sense to overwrite other transformers

fealho: You mean to just add the add-ons manually at the top of the file? I thought we didn't want to explicitly add them as imports since they're add-ons and not always imported.

pvk-developer, line 55: I think that this is a cleaner way to achieve that, we already have the name of the addons class just we were skipping it. 
```python3
                module, class_name = transformer.rsplit('.', 1)
                if module not in sys.modules:
                    module = importlib.import_module(module)
                    globals()[name] = getattr(module, class_name)
                    __all__.append(name)
```

pvk-developer (approved):
👍🏻

amontanez24 (approved):
Thanks for addressing the changes

PR #396, opened 2022-02-16
Support dropping column transformer

### LINKED ISSUE #393: Support dropping a column trough a transformer
### Problem Description
Currently our transformers are expected to always return a `transformed` column, but in some cases we would like to learn something from the column, drop it and then populate it with values during the `reverse_transform`.

### Expected behavior

```python
class PIITransformer(BaseTransformer):
    OUTPUT_TYPES = {
        'pii': 'categorical'
    }
    def __init__(self):
        self.length = None
        self.faker = Faker()
    
    def _fit(self, columns_data):
        self.length = len(columns_data)
        
    def _transform(self, columns_data):
        return None  # Drop the column
    
    def _reverse_transform(self, columns_data):
        return [self.faker.name() for i in range(self.length)]  # Generate the values
```

Which should lead to the following outcomes:
```python
data = pd.DataFrame({
    'name': ['John', 'Doe', 'John Doe', 'John Doe John'],
    'donation': [20, 30, 40, 50]
})

pii_transformer = PIITransformer()
transformed = pii_transformer.fit_transform(data, 'name')
transformed

   donation
0        20
1        30
2        40
3        50
```
Which would contain only the `donation` value (the `name` has been dropped).

When `reverse_transforme`:
```python
data = pii_transformer.reverse_transform(transformed)
data

   donation              name
0        20    Gregory Rivera
1        30  Melissa Peterson
2        40   Mrs. Lori Pitts
3        50       Susan Payne
```

A column `name` has been returned with random values, the data contains both `name` and `donation` columns.

pvk-developer:
Resolves #393

This pr also brings the following changes:
* Cap the `scipy` version to `<1.8` since it causes performance issues.
* Update `github-actions` setup to `v2` since `Windows` does not have `py36` env.
* Update the `get_output_columns` to call `get_output_types` which now always returns a dict (even if `OUTPUT_TYPES` is None).

### DIFF
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -52,7 +52,7 @@ class BaseTransformer:
 
     def _add_prefix(self, dictionary):
         if not dictionary:
-            return None
+            return {}
 
         output = {}
         for output_columns, output_type in dictionary.items():
@@ -67,7 +67,7 @@ class BaseTransformer:
             dict:
                 Mapping from the transformed column names to the produced data types.
         """
-        return self._add_prefix(self.OUTPUT_TYPES) if self.OUTPUT_TYPES else {}
+        return self._add_prefix(self.OUTPUT_TYPES)
 
     def is_transform_deterministic(self):
         """Return whether the transform is deterministic.
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -67,7 +67,7 @@ class BaseTransformer:
             dict:
                 Mapping from the transformed column names to the produced data types.
         """
-        return self._add_prefix(self.OUTPUT_TYPES)
+        return self._add_prefix(self.OUTPUT_TYPES) if self.OUTPUT_TYPES else dict()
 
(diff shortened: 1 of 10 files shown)


rdt/transformers/base.py
```diff
@@ -147,9 +147,9 @@ def _set_columns_data(data, columns_data, columns):
         if isinstance(columns_data, (pd.DataFrame, pd.Series)):
             columns_data.index = data.index
 
-        if len(columns_data.shape) == 1:
+        if columns_data is not None and len(columns_data.shape) == 1:
```
amontanez24, line 150: Another option is to check if `columns_data` is `None` at the top and just return, so that we don't even bother with the check on line 147

amontanez24 (approved):
one small comment. Up to you if you want to change it, otherwise LGTM!

csala (approved):
I added a small comment about a change that we may want to do, but even without the change this looks ok to go.

PR #410, opened 2022-02-24
Correctly sort the columns after transforming/reverse_transforming in the `HyperTransformer`

### LINKED ISSUE #405: `HyperTransformer` should sort columns after `transform` and `reverse_transform`
### Problem Description

If the input columns don't match the order of the output columns, then users might think there is some mistake in the `transform`.

### Expected behavior

If the original data has a column order `A, B, C, ...` then:

1. The transformed data should have the same column order with respect to the original column positions
eg. `A.value, B.value, B.is_null, C.value, C.is_null, ...` (it is still the case that A is before B is before C...)
2. The reversed transformed data should have the same column order as the original

### Additional context

Recreate this issue using the code below

```python
import numpy as np
import pandas as pd

login_dates = ['2021-06-26', '2021-02-10', 'NAT', '2020-09-26', '2020-12-22']

df = pd.DataFrame(data={
    'last_login': [np.datetime64(i) for i in login_dates],
    'email_optin': [False, False, False, True, np.nan],
    'credit_card': ['VISA', 'VISA', 'AMEX', np.nan, 'DISCOVER'],
    'age': [29, 18, 21, 45, 32],
    'dollars_spent': [99.99, np.nan, 2.50, 25.00, 19.99]
})
```
image

```python

from rdt.transformers import datetime, boolean, categorical, numerical

field_transformers = {
    'last_login': datetime.DatetimeTransformer(null_column=True),
    'email_optin': boolean.BooleanTransformer(null_column=False, nan=0),
    'credit_card': categorical.CategoricalTransformer(fuzzy=True),
    # 'age': numerical.NumericalTransformer(null_column=False),
    'dollars_spent': numerical.NumericalTransformer(null_column=False, nan=0)
}

ht = HyperTransformer(field_transformers=field_transformers)

ht.fit_transform(df)
```
image

Notice how `age.value` is at the very end now.

fealho:
Resolve #405.


rdt/hyper_transformer.py
```diff
@@ -382,6 +382,24 @@ def _validate_all_fields_fitted(self):
...
+        This method sorts ``_output_columns`` such that the generated output columns
+        have the same order as the original ``_input_columns``.
+        """
+        def _sort_key(output_column):
+            best_i, match_len = float('inf'), 0
+            for i, input_column in enumerate(self._input_columns):
+                if output_column.startswith(input_column):
+                    if match_len < len(input_column):
+                        best_i = i
+                        match_len = len(input_column)
```
amontanez24, line 397: We have a method called `get_final_output_columns` that returns the output columns derived from a specified input column. Instead of checking that the names match, we can use that to see if it is a match

amontanez24 (changes requested):
Left one comment

rdt/hyper_transformer.py
fealho: That method doesn't handle combined columns (e.g. `day#month#year.value`). We don't really have any transformers that need it for now, so I'm not sure if we care about it?

amontanez24: I don't think we need to worry about that for now

rdt/hyper_transformer.py
```diff
@@ -288,7 +288,7 @@ def get_final_output_columns(self, field):
             else:
                 final_outputs.append(output)
 
-        return final_outputs
+        return sorted(final_outputs, reverse=True)
```
fealho, line 291: This method returns values like `[column.is_null, column.value]`, but since we expect the column order to be the other way around (i.e. `[column.value, column.is_null]`) I'm reversing the order.

rdt/hyper_transformer.py
```diff
@@ -370,10 +370,6 @@ def _fit_field_transformer(self, data, field, transformer):
                 if self._field_in_data(output_field, data):
                     self._fit_field_transformer(data, output_field, next_transformer)
 
-            else:
-                if output_name not in self._output_columns:
-                    self._output_columns.append(output_name)
```
fealho, line 375: Deleting this because we need to reset the `_output_columns` when sorting them (this also doesn't get used before the transform/reverse_transform, so moving it later in the fit doesn't affect anything).

tests/integration/test_hyper_transformer.py
```diff
@@ -47,44 +47,6 @@ def _reverse_transform(self, data):
         return data.astype('datetime64')
 
 
-class DummyTransformerMultiColumn(BaseTransformer):
```
fealho, line 50: Deleting these tests because we are dropping support for multi-column transformers.

tests/unit/test_hyper_transformer.py
```diff
@@ -14,6 +14,52 @@
 
 class TestHyperTransformer(TestCase):
 
+    def test__add_field_to_set_string(self):
```
fealho, line 17: Need to add these tests to get 100% coverage (I guess only the multi-column tests used the tuple case of this method, so deleting the tests decreased the coverage.)

tests/unit/test_hyper_transformer.py
```diff
@@ -363,7 +408,6 @@ def test__fit_field_transformer(self, get_transformer_instance_mock):
             'a.out1': ['2', '4', '6'],
             'a.out2': [1, 2, 3]
         })
-        assert ht._output_columns == ['a.out1.value', 'a.out2']
```
fealho, line 366: Fit_field_transformers doesn't set _output_columns anymore.

amontanez24 (approved):
LGTM!

katxiao (approved):

PR #411, opened 2022-02-28
Add PIIAnonymizer transformer

### LINKED ISSUE #397: Add `PIIAnonymizer`
### Problem Description

In our efforts to move more `PII` logic into `RDT`, we should add a `PIIAnonymizer` transformer that provides deep integration with Faker.

### Expected behavior

Add a `PIIAnonymizer` object. Params:
- `provider_name` the name of the provider in [Faker](https://faker.readthedocs.io/en/master/providers/faker.providers.credit_card.html#faker.providers.credit_card.Provider.credit_card_number) -- `None` means use the [base provider](https://faker.readthedocs.io/en/master/providers/baseprovider.html#)
- (required) `function_name` the name of the function to use within the provider -- default = `'lexify'`
- `locales` optional list of [localized providers](https://faker.readthedocs.io/en/master/locales.html) to use instead of the global provider. eg. `['en_US', 'fr_CA']`
- `function_kwargs` optional keyword args to pass into the `function_name`

Then:
- The forward transform will drop the column
- The reverse transform will recreate the column using: `<locale> faker.providers.<provider_name>.<function_name>(<function_kwargs>)`

The object would reside in a new, `pii` subsection.

```python
>>> from rdt.transformers.pii import PIIAnonymizer
```

By default, the Faker would use the [lexify method](https://faker.readthedocs.io/en/master/providers/baseprovider.html#faker.providers.BaseProvider.lexify) to that generates random character strings.

```python
>>> default_faker = PIIAnonymizer()
>>> default_faker.fit(data, column='username')

# drops the column 'username'
>>> transformed_data = default_faker.transform(data)

# recreates 'username' using faker.providers.BaseProvider.lexify()
>>> default_faker.reverse_transform(transformed_data)

# print with the default arguments
>>> print(default_faker)
<PIIAnonymizer(function_name='lexify')>
```

The user can override this to use any of Faker's functions.

```python
# easy ability to specify ALL the fields
>>> credit_faker = PIIAnonymizer(provider_name='credit_card',
                function_name='credit_card_number',
                function_kwargs={'card_type': 'visa'})
>>> credit_faker.fit(data, column='credit_card')

# drops the column 'credit_card'
>>> transformed_data = credit_faker.transform(data)

# recreates 'credit_card' using faker.providers.credit_card.credit_card_number(card_type='visa')
>>> credit_faker.reverse_transform(transformed_data)

# print out the object
>>> print(credit_faker)
<PIIAnonymizer(provider_name="credit_card", function_name="credit_card_number", function_kwargs{"card_type": "visa"})> 
```

### Error & Warning States
```python
# user specifies a provider_name but no function_name
>>> faker = PIIAnonymizer(provider_name='credit_card')
Error: Please specify the function name to use from the 'credit_card' provider.

# user specifies a function_name that is not in the provider
>>> faker = PIIAnonymizer(provider_name='credit_card', function_name='address')
Error: The 'credit_call' module does not contain a function named 'address'.
Refer to the Faker docs to find the correct function: https://faker.readthedocs.io/en/master/providers.html

# provider is not available for particular locales
>>> faker = PIIAnonymizer(provider_name='phone_number', function_name='phone_number',
                          locales=['ar_AA', 'en_IE', 'en_GB'])
Warning: Locales ['ar_AA', 'en_IE'] do not support provider 'phone_number' and function 'phone_nunber'.
In place of these locales, 'en_US' will be used instead. Please refer to the localized provider docs
for more information: https://faker.readthedocs.io/en/master/locales.html
```

pvk-developer:
Resolves #397


rdt/transformers/pii/anonymizer.py
```diff
@@ -0,0 +1,138 @@
...
+        ], dtype=object)
+
+        if self.null_transformer.models_missing_values():
+            reverse_transformed = np.column_stack((reverse_transformed, columns_data))
+
+        return self.null_transformer.reverse_transform(reverse_transformed)
+
+    def __repr__(self):
+        """Represent the class instance only with the modified default arguments."""
+        printable = [
```
amontanez24, line 120: minor: is there a way to get this list programmatically from the `__init__` method? In case any parameters change

rdt/hyper_transformer.py
```diff
@@ -78,6 +78,7 @@ class HyperTransformer:
         'O': 'categorical',
         'b': 'boolean',
         'M': 'datetime',
+        'p': 'pii',
```
amontanez24, line 81: This is used to pick the `RDT` data type based on a pandas `dtype`. I don't think they have one that is `p` so this won't do anything. We can get rid of this line

tests/integration/transformers/pii/test_faker.py
```diff
@@ -0,0 +1,105 @@
+
+import numpy as np
+import pandas as pd
+
+from rdt.transformers.pii import PIIAnonymizer
+
+
+def test_piifaker():
```
amontanez24, line 8: should be `piianonymizer`. Can we do one more pass through to make sure all of the names are correct?

amontanez24 (approved):
This looks really good! Just a few minor comments then should be good ready to go

rdt/transformers/pii/anonymizer.py
```diff
@@ -0,0 +1,138 @@
...
+            dict:
+                Mapping from the transformed column names to supported data types.
+        """
+        output_types = {}
+        if self.null_transformer and self.null_transformer.models_missing_values():
+            output_types['is_null'] = 'float'
+
+        return self._add_prefix(output_types)
+
+    def _fit(self, columns_data):
```
fealho, line 93: `columns_data` should just be `data` (we changed it in the other transformers as well since this is always a series). 

Also, add the documentation for fit/transform/reverse_transform (just copy paste from another transformer).

rdt/transformers/pii/anonymizer.py
```diff
@@ -0,0 +1,138 @@
...
+        self.null_transformer = NullTransformer(
+            self.missing_value_replacement,
+            self.model_missing_values
+        )
+        self.null_transformer.fit(columns_data)
+        self.data_length = len(columns_data)
+
+    def _transform(self, columns_data):
```
fealho, line 101: `columns_data` ->`data`

rdt/transformers/pii/anonymizer.py
```diff
@@ -0,0 +1,138 @@
...
+        if self.null_transformer and self.null_transformer.models_missing_values():
+            return self.null_transformer.transform(columns_data)[:, 1].astype(float)
+
+        return None
+
+    def _reverse_transform(self, columns_data):
```
fealho, line 107: `columns_data` ->`data`

fealho (approved):
Looks good! I left a few small comments for you to take a look.

rdt/transformers/pii/anonymizer.py
pvk-developer: 👍🏻

rdt/transformers/pii/anonymizer.py
pvk-developer: Yes there it is 😅 , I have also changed the `BaseTransformer` one because of the following 'issue' that I found:

With the current implementation:
```python
class Foo:
    def __init__(self, param=None):
        self.param = param
        self.object = object()
        
    def __repr__(self):
        class_name = self.__class__.__name__
        args = []
        default = self.__class__()
        for arg, value in self.__dict__.items():
            if default.__dict__[arg] != value:
                if isinstance(value, str):
                    value = f"'{value}'"
                args.append(f'{arg}={value}')

        args_string = ', '.join(args)
        return f'{class_name}({args_string})'
```
When running this code:
```
Foo('a')
Foo(param='a', object=<object object at 0x7f3e6ceca620>)
```

However the `object` itself shouldn't be appearing there, it doesn't form part of the `__init__`. In order to fix it we can use `inspect` to check which are the `init` parameters and the defaults if there are any:
```python
def __repr__(self):
        class_name = self.__class__.__name__
        custom_args = []
        args = inspect.getfullargspec(self.__init__)
        keys = args.args[1:]
        defaults = args.defaults or []
        defaults = dict(zip(keys, defaults))
        instanced = {key: getattr(self, key) for key in keys}

        if defaults == instanced:
            return f'{class_name}()'

        for arg, value in instanced.items():
            if defaults[arg] != value:
                value = f"'{value}'" if isinstance(value, str) else value
                custom_args.append(f'{arg}={value}')

        args_string = ', '.join(custom_args)
        return f'{class_name}({args_string})'
```

PR #412, opened 2022-02-28
Update `HyperTransformer` defaults

### LINKED ISSUE #398: Update `HyperTransformer` default transformers
### Problem Description

We have renamed and modified the individual transformers. As a result:
- We need to rename them in `HyperTransformer`
- We should also update the default functionalities in `HyperTrasnformer`

### Expected behavior

`HyperTransformer` should start with the following defaults per data type:
image

fealho:
Resolve #398.


rdt/__init__.py
```diff
@@ -20,15 +20,13 @@
 
 
 def get_demo(num_rows=5):
-    """Generate random demo data with multiple data types.
+    """Generate demo data with multiple data types.
+
+    The first five rows are hard coded. The rest is randomly generated.
```
pvk-developer, line 25:
```  
    # Randomly generate the remaining rows
    np.random.seed(42)
```  
I think we should say something like: The rest is randomly generated using ``np.random.seed(42)``.

rdt/hyper_transformer.py
```diff
@@ -47,7 +47,7 @@ class HyperTransformer:
 
         >>> field_data_types = {
         ...     'a': 'categorical',
-        ...     'b': 'numerical
+        ...     'b': 'numerical'
```
pvk-developer, line 50: nice catch 👍🏻

pvk-developer (approved):
LGTM!

tests/integration/test_hyper_transformer.py
```diff
@@ -124,14 +124,37 @@ def get_transformed_data():
...
+        '2010-01-01',
+        '2010-02-01',
+        '2010-01-01',
+        '2010-01-01',
+    ])
+    data = pd.DataFrame({
+        'integer': [1, 2, 1, 3, 1, 4, 2, 3],
+        'float': [0.1, 0.2, 0.1, np.nan, 0.1, 0.4, np.nan, 0.3],
+        'categorical': ['a', 'a', np.nan, 'b', 'a', 'b', 'a', 'a'],
+        'bool': [np.nan, np.nan, np.nan, True, np.nan, True, True, np.nan],
```
amontanez24, line 150: I think we should probably handle this problem in another issue. I don't think mean values should all be replaced with nans

amontanez24 (approved):
LGTM!

PR #415, opened 2022-03-01
Improve the way we print an individual transformer

### LINKED ISSUE #395: Improve the way we print an individual transformer
### Problem Description

Users who are inspecting a transformer want to know the hyperparameters that are being used to create it. To improve usability, we should improve the way each individual transformer is printed.

### Expected behavior

Every individual transformer should print:
- name of the transformer
- hyperparameters used to create it (ie all hyperparameters that are _not_ the default)

**Goal**: User should know exactly how to create that object.

```python
# Default: No parameters are printed because they are all defaults
>>> from transformers.numerical import FloatFormatter
>>> default_transformer = FloatFormatter()
>>> print(default_transformer)
<FloatFormatter()>

# Overriding parameter(s): Only print the overridden parameter(s)
>>> from transformers.boolean import BinaryEncoder
>>> transformer = BinaryEncoder(missing_value_replacement='mode')
>>> print(transformer)
<BinaryEncoder(missing_value_replacement="mode")>
```

### Additional context

One motivation for this change is the `HyperTransformer`. When it returns the (first) transformer for each column, it will be nice if users can see which transformers are being used.

amontanez24:
resolves #395

### DIFF
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -165,6 +165,23 @@ class BaseTransformer:
             self.column_prefix += '#'
             self.output_columns = list(self.get_output_types().keys())
 
+    def __str__(self):
+        """Print full initialization of transformer.
+
+        Returns:
+            str:
+                The name of the transformer followed by any non-default parameters.
+        """
+        class_name = self.__class__.__name__
+        args = []
+        default = self.__class__()
+        for arg, value in self.__dict__.items():
+            if default.__dict__[arg] != value:
+                args.append(f'{arg}: {value}')
+
+        args_string = ', '.join(args)
+        return f'{class_name}({args_string})'
+
     def _fit(self, columns_data):
         """Fit the transformer to the data.
 
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -177,7 +177,7 @@ class BaseTransformer:
         default = self.__class__()
         for arg, value in self.__dict__.items():
             if default.__dict__[arg] != value:
-                args.append(f'{arg}: {value}')
+                args.append(f'{arg}={value}')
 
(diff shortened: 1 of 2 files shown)


fealho (approved):
Pretty good 👍

rdt/transformers/base.py
```diff
@@ -165,6 +165,32 @@ def _build_output_columns(self, data):
...
+        Returns:
+            str:
+                The name of the transformer followed by any non-default parameters.
+        """
+        class_name = self.__class__.__name__
+        args = []
+        default = self.__class__()
+        for arg, value in self.__dict__.items():
+            if default.__dict__[arg] != value:
+                args.append(f'{arg}={value}')
```
pvk-developer, line 180: I think that we should add the following:
```python
if isinstance(value, str):
    value = f"'{value}'"
```
The reason why is that currently with this implementation if I do the following:
```python
In [1]: from rdt.transformers import FloatFormatter

In [2]: trans = FloatFormatter(missing_value_replacement='mean')

In [3]: print(trans)
FloatFormatter(missing_value_replacement=mean)
```
Where I think the expected result should be `FloatFormatter(missing_value_replacement='mean')`

tests/unit/transformers/test_base.py
```diff
@@ -115,6 +115,66 @@ def test__add_prefix_dictionary(self):
...
+
+        Expected behavior:
+            - The mock should be called once.
+        """
+        # Setup
+        transformer = BaseTransformer()
+        transformer.__str__ = Mock()
+
+        # Run
+        transformer.__repr__()
```
pvk-developer, line 173: here we could use `repr(transformer)` (similar to `str`)

pvk-developer (approved):

PR #417, opened 2022-03-02
Add `detect_initial_config` method

### LINKED ISSUE #399: Create method to auto detect `HyperTransformer` config from data
### Problem Description
When given data, `HyperTransformer` should be able to auto detect the config (data types and first transformers) & show them to the user. That way, the user can update them **before** the expensive `fit` computation. 

Changes:
- Add method `detect_initial_config` that has 1 required param `data` (a `pandas.DataFrame`)
- Rename `data_types` to `sdtype` (semantic data type)

### Expected behavior

```python
>>> ht = HyperTransformer()
>>> ht.detect_initial_config(data=my_pandas_dataframe)
Detecting a new config from the data ... SUCCESS
Setting the new config ... SUCCESS

Config:
{
  'sdtypes': {
    'colA': 'numerical',
    'colB': 'categorical'
  },
  'transformers': {
    'colA': <FloatFormatter(missing_value_replacement="mean")>,
    'colB': <FrequencyEncoder()>
  }
}

```

(See #395 for how to print out transformers.)

### Additional context

- We expect the user to use this method _only once_ before they fit their data
- Calling this method again with erase any pre-existing config and auto-detect it from the new data being passed in

fealho:
Resolve the first part of #399.


tests/integration/test_hyper_transformer.py
```diff
@@ -207,6 +207,9 @@ def test_hypertransformer_default_inputs():
     assert isinstance(ht._transformers_tree['names']['transformer'], FrequencyEncoder)
     assert ht._transformers_tree['names']['outputs'] == ['names.value']
 
+    get_default_transformers.cache_clear()
```
fealho, line 210: We need these lines because `get_default_transformer(s)` caches the output of the function, but since this test patches the method, it ends up caching the patched, incorrect value, which makes any tests that run after it fail.

Deleting the cache fixes the problem.

rdt/hyper_transformer.py
```diff
@@ -320,6 +324,40 @@ def get_transformer_tree_yaml(self):
...
+        self.field_transformers = {}
+
+        # Set the sdtypes and transformers of all fields to their defaults
+        for field in data:
+            self._set_field_data_type(data, field)
+            field_sdtype = self.field_data_types[field]
+            self.field_transformers[field] = get_default_transformer(field_sdtype)
+
+        print('Detecting a new config from the data ... SUCCESS')  # noqa: T001
+        print('Setting the new config ... SUCCESS')  # noqa: T001
```
amontanez24, line 351: Is there a way to add this warning for the whole method instead of on each line where we print?

rdt/hyper_transformer.py
```diff
@@ -320,6 +324,40 @@ def get_transformer_tree_yaml(self):
...
+
+        config = {
+            'sdtypes': self.field_data_types,
+            'transformers': {k: repr(v) for k, v in self.field_transformers.items()}
+        }
+
+        print('Config:')  # noqa: T001
+        print(json.dumps(config, indent=4))  # noqa: T001
```
amontanez24, line 359: should we print as json or yaml? @npatki

amontanez24 (approved):
This looks good to me! Let's wait till we merge in the changes concerning the workflows before merging this

rdt/hyper_transformer.py
pvk-developer: I'm a bit skeptical about the whole printing but here is my proposal, I know I'm iterating twice on the data but I don't see a better way to do it:
```python
        # Set the sdtypes and transformers of all fields to their defaults
        print('Detecting a new config from the data ... ', end='')  # noqa: T001
        try:
            for field in data:
                self._set_field_data_type(data, field)
                field_sdtype = self.field_data_types[field]

            print('SUCCESS')  # noqa: T001
        except Exception e:
            print('ERROR')  # noqa: T001

        print('Setting the new config ... ', end='')  # noqa: T001
        try:
            for field in data:
                self.field_transformers[field] = get_default_transformer(field_sdtype)

             print('SUCCESS')  # noqa: T001
        except Exception e:
            print('ERROR')  # noqa: T001
```

This way we can print `ERROR` or `SUCCESS` and actually not just 'straight forward'  print a message. @npatki  @amontanez24 your thoughts on this ?

rdt/hyper_transformer.py
pvk-developer: @fealho nice print ! 😸 I didn't know you could do this! I usually used `preatyprint`  but this is a nice hack! 👍🏻

pvk-developer (approved):
Looks good to me!

rdt/hyper_transformer.py
fealho: I'm not sure. If it exists I wasn't able to find it.

PR #425, opened 2022-03-08
Correctly generate `_output_columns`

### LINKED ISSUE #423: Duplicate `_output_columns` during fitting
When fitting multiple times, `_output_columns` should be reset during each `fit`. Instead, it currently only adds the new output columns to the previous ones, as shown in the code below:

```python3
import pandas as pd

data = pd.DataFrame({'a': [1,2,3]})
ht = HyperTransformer()

ht.fit(data)
ht.fit(data)
ht.fit(data)
ht.transform(data)
```
This produces the following output:

  a.value | a.value | a.value 
-- | -- | -- 
1 | 1 | 1
2 | 2 | 2
3 | 3 | 3

fealho:
Resolve #423.


katxiao (approved):

rdt/hyper_transformer.py
```diff
@@ -381,6 +381,7 @@ def _validate_all_fields_fitted(self):
 
     def _sort_output_columns(self):
         """Sort ``_output_columns`` to follow the same order as the ``_input_columns``."""
+        self._output_columns = []
```
amontanez24, line 384: Since this method uses `get_final_output_columns` which uses the `_transformers_tree`, I think we need to reset the `_transformers_tree` at the beginning of `fit`. Otherwise, if we fit onto data with different output columns, I think the output columns from the previous fit will show up as well.

rdt/hyper_transformer.py
```diff
@@ -188,8 +193,8 @@ def get_config(self):
                 - transformers: A dictionary mapping column names to their transformer instances.
         """
         return {
-            'sdtypes': self.field_sdtypes,
-            'transformers': self.field_transformers
+            'sdtypes': {**self.provided_field_sdtypes, **self._fitted_field_sdtypes},
```
amontanez24, line 196: I think we may want to have a method that correctly combines these two dictionaries into one. Then anywhere that needs to loop over the whole dict can just call that method. We should also probably have a method for setting things in that dictionary that removes from the `_fitted` one if necessary. So something like:
```
_get_field_sdtypes() => returns combination of both dicts
_set_field_sdtype() => adds to provided dict and removes from _fitted
```

Either that or we could make a class to manage all of this but that may be overkill.

rdt/hyper_transformer.py
```diff
@@ -367,30 +372,35 @@ def detect_initial_config(self, data):
...
+        self.provided_field_transformers = {}
+        self._fitted_field_transformers = {}
 
         # Set the sdtypes and transformers of all fields to their defaults
         for field in data:
             self._set_field_sdtype(data, field)
-            field_sdtype = self.field_sdtypes[field]
-            self.field_transformers[field] = get_default_transformer(field_sdtype)
+            field_sdtype = self._fitted_field_sdtypes[field]
+            self._fitted_field_transformers[field] = get_default_transformer(field_sdtype)
```
amontanez24, line 384: I think all of this logic (except for resetting the provided dicts) can be moved into a `_learn_config` method that can also be called at the beginning of `fit`.

rdt/hyper_transformer.py
```diff
@@ -458,15 +468,19 @@ def fit(self, data):
             data (pandas.DataFrame):
                 Data to fit the transformers to.
         """
+        self._unfit()
         self._input_columns = list(data.columns)
         self._populate_field_sdtypes(data)
 
         # Loop through field_transformers that are first level
-        for field in self.field_transformers:
+        for field in self.provided_field_transformers:
```
amontanez24, line 476: If we use the `_learn_config` method described above then we can just loop through the combined dictionaries here by calling `_get_field_transformers()`

amontanez24 (changes requested):

rdt/hyper_transformer.py
```diff
@@ -220,8 +228,9 @@ def update_field_sdtypes(self, field_sdtypes):
                 update the existing ``field_sdtypes`` values. Calling this method will
                 require ``fit`` to be run again.
         """
-        self.provided_field_sdtypes.update(field_sdtypes)
-        self._unfit()
+        self._provided_field_sdtypes.update(field_sdtypes)
+        self.field_sdtypes.update(field_sdtypes)
+        self._unfit()  # delete?
```
pvk-developer, line 233: Yes I think that this unfit statements should be deleted from here.

rdt/hyper_transformer.py
```diff
@@ -151,20 +156,21 @@ def _field_in_data(field, data):
...
-        self._fitted_field_sdtypes[field] = self._DTYPES_TO_SDTYPES[kind]
+        assert field not in self.field_sdtypes
+        self.field_sdtypes[field] = self._DTYPES_TO_SDTYPES[kind]
 
     def _populate_field_sdtypes(self, data):
         # get set of provided fields including multi-column fields
         provided_fields = set()
-        for field in self.provided_field_sdtypes.keys():
-            self._add_field_to_set(field, provided_fields)
+        for field in self.field_sdtypes.keys():
```
pvk-developer, line 165: We can iterate directly over the `dict` no  needs for `keys()`.

fealho: Eh, I prefer to be explicit, so I usually use `keys()` when iterating over dicts (except if the line is over 100 characters...)

rdt/hyper_transformer.py
```diff
@@ -353,28 +341,51 @@ def get_transformer_tree_yaml(self):
...
+    def _set_field_sdtype(self, data, field):
+        clean_data = data[field].dropna()
+        kind = clean_data.infer_objects().dtype.kind
+        self.field_sdtypes[field] = self._DTYPES_TO_SDTYPES[kind]
+
+    def _learn_config(self, data):
+        for field in data:
+            if field not in self.field_sdtypes:
+                self._set_field_sdtype(data, field)
+            if field not in self.field_transformers:
```
amontanez24, line 353: Shouldn't this only set if it isnt in the `provided` dicts? I thought we were going to relearn anything that wasn't provided

amontanez24 (commented):
This is looking good! I left one question about the `_learn_config` method

rdt/hyper_transformer.py
fealho: Yes, at this point in the code field_sdtypes is the same as _provided_field_sdtypes, because they just got reset by _unfit. I can move the _unfit inside this method to make that more clear.

amontanez24: Not always. This is also called by `fit` which doesn't reset the `provided` dicts and that's how it should be. We want to relearn things that weren't provided, even if they were previously learned

amontanez24 (approved):
LGTM!

pvk-developer (approved):

PR #428, opened 2022-03-08
Implement update_transformer

### LINKED ISSUE #402: Update transformers by column name in `HyperTransformer`
### Problem Description

We should allow users to: update any of the transformers assigned to each of the column names -- the param name `column_name_to_transformer`

### Expected behavior

```python
>>> ht.update_transformers(column_name_to_transformer={
       'age': numerical.NumericalFormatter(missing_value_replacement='mean', model_missing_values=True),
       'last_login': datetime.OptimizedTimestampConverter(),
})
```

(See #395 for how we print a transformer object) 

### Edge Cases

**Scenario 1:** User tries update transformers without first using `detect_initial_config` (see #399 )
```python
>>> ht = HyperTransformer()
>>> ht.update_transformers(column_name_to_transformer={
       'last_login': datetime.OptimizedTimestampConverter()
})
Error: Nothing to update. Use the `detect_initial_config` method to pre-populate all the sdtypes
and transformers from your dataset.
```

**Scenario 2:** The user tries to update to a transformer for a different sdtype (per our records)
```python
# assume the following sdtypes that 'last_login' is a datetime
# print out a warning per column that this triggers for
>>> ht.update_transformers({'last_login': numerical.NumericalFormatter()})
Warning: You are assigning a numerical transformer to a datetime column ('last_login').
If the transformer doesn't match the sdtype, it may lead to errors.
```

**Scenario 3:** User updates the transformers after calling `fit` or `fit_transform` on the data already.
```python
>>> ht.fit(data)
>>> ht.update_transformers({'last_login': datetime.OptimizedTimestampConverter()})
Warning: For this change to take effect, please refit your data using 'fit' or 'fit_transform'.
```

pvk-developer:
Resolves #402

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -272,8 +272,6 @@ class HyperTransformer:
 
             self.field_transformers[column_name] = transformer
             self._provided_field_transformers[column_name] = transformer
-            self.field_sdtypes[column_name] = transformer.get_input_type()
-            self._provided_field_sdtypes[column_name] = transformer.get_input_type()
 
     def set_first_transformers_for_fields(self, field_transformers):
         """Set the first transformer to use for certain fields.
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -86,6 +86,16 @@ class HyperTransformer:
     ]
 
     @staticmethod
+    def print_tip(text):
+        """Print a text with ``Tip: `` at the start of the text.
+
+        Args:
+            text (str):
+                Text to print.
+        """
+        print(f'Tip: {text}')  # noqa: T001
+
+    @staticmethod
     def _add_field_to_set(field, field_set):
         if isinstance(field, tuple):
             field_set.update(field)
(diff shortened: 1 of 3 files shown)


rdt/hyper_transformer.py
```diff
@@ -212,6 +212,44 @@ def update_default_data_type_transformers(self, new_data_type_transformers):
...
+        Args:
+            column_name_transformer(dict):
+                Dict mapping column names to transformers to be used for that column.
+        """
+        if self._fitted:
+            warning_msg = (
+                "Warning: For this change to take effect, please refit your data using 'fit' "
+                "or 'fit_transform'."
+            )
+            print(warning_msg)  # noqa: T001
```
amontanez24, line 227: should we create custom tip and warning methods that just print the text with the label so that we can just ignore the lint problem in one place? eg:
```
def tip(text):
    print(f'Tip: {text}') # noqa:T001
````

File: rdt/hyper_transformer.py [hunk 2]
```diff
@@ -212,6 +212,44 @@ def update_default_data_type_transformers(self, new_data_type_transformers):
...
+            tip_msg = (
+                'Tip: Use the `detect_initial_config` method to pre-populate all the sdtypes '
+                'and transformers from your dataset.'
+            )
+            print(tip_msg)  # noqa: T001
+
+        for column_name, transformer in column_name_transformer.items():
+            current_transformer = self.field_transformers.get(column_name)
+            if current_transformer:
+                current_transformer_type = current_transformer.get_input_type()
```
amontanez24, line 239: Instead of checking the current transformer I think we should check the `field_sdtypes` dict and compare to what is set there

rdt/hyper_transformer.py
```diff
@@ -232,6 +242,39 @@ def update_default_sdtype_transformers(self, new_sdtype_transformers):
...
+
+        if len(self.field_transformers) == 0:
+            self.print_tip(
+                'Use the `detect_initial_config` method to pre-populate all the sdtypes '
+                'and transformers from your dataset.'
+            )
+
+        for column_name, transformer in column_name_transformer.items():
+            current_sdtype = self.field_sdtypes.get(column_name)
+            if current_sdtype and current_sdtype != transformer.get_input_type():
```
amontanez24, line 266: @npatki Does this method still run through if the user hasn't set the config or used `detect_initial_config`? If so, then should this error also raise if the `sdtype` currently in the config is `None`?

tests/unit/test_hyper_transformer.py
```diff
@@ -1159,6 +1175,191 @@ def test_update_default_sdtype_transformers(self):
...
+            'and transformers from your dataset.'
+        )
+        mock_print.assert_called_once_with(expected_message)
+        assert instance.field_transformers['my_column'] == mock_transformer
+        assert instance.field_sdtypes['my_column'] == 'datetime'
+        assert instance._provided_field_sdtypes == {'my_column': 'datetime'}
+        assert instance._provided_field_transformers == {'my_column': mock_transformer}
+
+    @patch('rdt.hyper_transformer.print')
+    def test_update_transformers_field_transformer_exists(self, mock_warnings):
```
amontanez24, line 1312: minor: can we change this name to reflect that it tests if the sdtype and transformer don't match?

amontanez24 (approved):
This LGTM! Just a few minor comments and waiting to see @npatki's response

rdt/hyper_transformer.py
```diff
@@ -232,6 +242,39 @@ def update_default_sdtype_transformers(self, new_sdtype_transformers):
         """
         self.default_sdtype_transformers.update(new_sdtype_transformers)
 
+    def update_transformers(self, column_name_transformer):
```
amontanez24, line 245: I believe the issue says it should be `column_name_to_transformer`

rdt/hyper_transformer.py
```diff
@@ -232,6 +242,39 @@ def update_default_sdtype_transformers(self, new_sdtype_transformers):
...
+                warnings.warn(
+                    f'You are assigning a {transformer.get_input_type()} transformer '
+                    f'to a {current_sdtype} column ({column_name}). '
+                    "If the transformer doesn't match the sdtype, it may lead to errors."
+                )
+
+            self.field_transformers[column_name] = transformer
+            self._provided_field_transformers[column_name] = transformer
+            self.field_sdtypes[column_name] = transformer.get_input_type()
+            self._provided_field_sdtypes[column_name] = transformer.get_input_type()
```
amontanez24, line 276: @npatki should the `sdtype` also be updated based on the transformer?

rdt/hyper_transformer.py
npatki: +1 for consistency's sake, we can do the same error that we do when setting transformers by `sdtype`

rdt/hyper_transformer.py
```diff
@@ -232,6 +242,37 @@ def update_default_sdtype_transformers(self, new_sdtype_transformers):
...
+        """
+        if self._fitted:
+            warnings.warn(
+                "For this change to take effect, please refit your data using 'fit' "
+                "or 'fit_transform'."
+            )
+
+        if len(self.field_transformers) == 0:
+            self.print_tip(
+                'Use the `detect_initial_config` method to pre-populate all the sdtypes '
```
amontanez24, line 260: We are changing this to now raise an error and crash if no config has been set

PR #429, opened 2022-03-09
Error Handling

### LINKED ISSUE #408: Input validation & error handling in `HyperTransformer`
### Expected behavior

We should check input and provide friendly tips, warnings and errors to guide the user into correct `HyperTransformer` usage.

**Scenario 1:** User calls `fit` or `fit_transform` directly without using the new `auto_detect_config` method (see #399)

```python
>>> ht = HyperTransformer()
>>> ht.fit(data)
Error: No config detected. Set the config using `set_config` or pre-populate it automatically from your data
using `detect_initial_config` prior to fitting your data.
```

**Scenario 2:** User tries to transform a dataset that does not have the same column names or sdtypes as the fit data.
```python
>>> ht.transform(different_data)
Error: The data you are trying to transform has different columns than the original data. Column names and their sdtypes
must be the same. Use the method 'get_config()' to see the expected values. 
```
_Note: This should trigger any time that the set of columns is different. If the set is the same -- but the columns are just in a different order -- then we should proceed with the transformation._

**Scenario 3:** User tries to fit a dataset that does not have the same columns as the config
```python
>>> ht.fit(different_data)
Error: The data you are trying to transform has different columns than the original data. Column names and their sdtypes
must be the same. Use the method 'get_config()' to see the expected values. 
```
_Note: This should trigger any time that the set of columns is different. If the set is the same -- but the columns are just in a different order -- then we should proceed with the fit._

fealho:
Resolve #408.

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -147,6 +147,7 @@ class HyperTransformer:
         self._fitted_fields = set()
         self._fitted = False
         self._transformers_tree = defaultdict(dict)
+        self._config_detected = False
 
     @staticmethod
     def _field_in_data(field, data):
@@ -388,6 +389,7 @@ class HyperTransformer:
         self.default_sdtype_transformers = {}
         self._provided_field_sdtypes = {}
         self._provided_field_transformers = {}
+        self._config_detected = True
 
         # Set the sdtypes and transformers of all fields to their defaults
         self._learn_config(data)
@@ -467,6 +469,19 @@ class HyperTransformer:
             output_columns = self.get_final_output_columns(input_column)
             self._output_columns.extend(output_columns)
 
+    def _validate_detect_config_called(self):
+        """Warn the user if they fit without calling ``detect_initial_config``.
+
+        Note that this tip will only be printed until the first usage of
+        ``detect_initial_config`` method. After that method has been used once
+        the tip will no longer be shown (for an instance of the HyperTransformer).
+        """
+        if not self._config_detected:
+            warnings.warn(
+                "Tip: You can use the method 'detect_initial_config' to inspect "
+                "the sdtypes and transformers first before fitting the data."
+            )
+
     def fit(self, data):
         """Fit the transformers to the data.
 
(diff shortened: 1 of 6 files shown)


rdt/hyper_transformer.py
```diff
@@ -467,13 +469,38 @@ def _sort_output_columns(self):
...
+    def _validate_detect_config_called(self, data):
+        """Warn the user if they fit without calling ``detect_initial_config``.
+
+        Note that this tip will only be printed until the first usage of
+        ``detect_initial_config`` method. After that method has been used once
+        the tip will no longer be shown (for an instance of the HyperTransformer).
+        """
+        if not self._config_detected:
+            warnings.warn(
+                "Tip: You can use the method 'detect_initial_config' to inspect "
```
amontanez24, line 481: Although the issue specifies to raise this if the detect method wasn't called, I think this warning should show whenever a config isn't present. Instead of checking if the method was called just check if there are any values set in the `field_transformers` and `field_sdtypes`

rdt/hyper_transformer.py
```diff
@@ -467,13 +469,38 @@ def _sort_output_columns(self):
...
+
+        else:
+            fields = list(self.field_sdtypes.keys())
+            unknown_columns = self._subset(data.columns, fields, not_in=True)
+            if unknown_columns:
+                raise NotFittedError(
+                    'The data you are trying to fit has different columns than the original '
+                    f'detected data (unknown columns: {unknown_columns}). Column names and their '
+                    "sdtypes must be the same. Use the method 'get_config()' to see the expected "
+                    'values.'
```
amontanez24, line 493: Can we move this message to a variable since it is duplicated?

rdt/hyper_transformer.py
```diff
@@ -467,13 +469,38 @@ def _sort_output_columns(self):
...
 
+    def _validate_detect_config_called(self, data):
+        """Warn the user if they fit without calling ``detect_initial_config``.
+
+        Note that this tip will only be printed until the first usage of
+        ``detect_initial_config`` method. After that method has been used once
+        the tip will no longer be shown (for an instance of the HyperTransformer).
+        """
+        if not self._config_detected:
+            warnings.warn(
```
amontanez24, line 480: The issue was changed to make this an error

amontanez24 (commented):
This is close! Some minor changes

pvk-developer (approved):
LGTM!

amontanez24 (approved):
I think this is good to go!

PR #445, opened 2022-03-22
Datetime reverse transform crashes if datetime_format is specified

### LINKED ISSUE #438: Datetime reverse transform crashes if `datetime_format` is specified
### Environment Details

* RDT version: 1.0.0 (dev branch)

### Error Description

For both `UnixTimestampEncoder` and `OptimizedTimestampEncoder`: If you specify a `datetime_format` string, the transformers are not able to reverse transform the data.

### Steps to reproduce

```python
from rdt import get_demo
from rdt.transformers.datetime import UnixTimestampEncoder

# update demo data with a different datetime format
data2 = get_demo()
data2['last_login'] = ['Jun 26, 2021', 'Feb 02, 2021', np.nan, 'Sep 26, 2020', 'Dec 22, 2020']

# use a datetime transformer on this data: give it the explicit format string
ute = UnixTimestampEncoder(datetime_format='%b %d, %Y')
transformed = ute.fit_transform(data2, column='last_login')

# this step crashes
ute.reverse_transform(transformed)

ValueError: time data '1624665600000000000' does not match format '%b %d, %Y' (match)
```

amontanez24:
resolves #438

### DIFF
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -144,7 +144,7 @@ class UnixTimestampEncoder(BaseTransformer):
         """
         data = self._reverse_transform_helper(data)
         datetime_data = pd.to_datetime(data)
-        if type(datetime_data) != pd.Series:
+        if not isinstance(datetime_data, pd.Series):
             datetime_data = pd.Series(datetime_data)
 
         if self.datetime_format:
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -102,9 +102,6 @@ class UnixTimestampEncoder(BaseTransformer):
         if self.missing_value_replacement is not None:
             data = self.null_transformer.reverse_transform(data)
 
-        if isinstance(data, np.ndarray) and (data.ndim == 2):
-            data = data[:, 0]
-
         data = np.round(data.astype(np.float64))
         return data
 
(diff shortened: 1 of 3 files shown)


rdt/transformers/datetime.py
```diff
@@ -102,9 +102,6 @@ def _reverse_transform_helper(self, data):
         if self.missing_value_replacement is not None:
             data = self.null_transformer.reverse_transform(data)
 
-        if isinstance(data, np.ndarray) and (data.ndim == 2):
-            data = data[:, 0]
```
amontanez24, line 106: I believe this is from this [PR](https://github.com/sdv-dev/RDT/pull/161/files)
I'm not sure why we would ever need this functionality. The transformer should only get more than one column in the case of nulls which is handled above

pvk-developer (approved):
LGTM!

katxiao (approved):

rdt/transformers/datetime.py
```diff
@@ -146,8 +143,14 @@ def _reverse_transform(self, data):
             pandas.Series
         """
         data = self._reverse_transform_helper(data)
+        datetime_data = pd.to_datetime(data)
+        if type(datetime_data) != pd.Series:
+            datetime_data = pd.Series(datetime_data)
```
amontanez24, line 148: `to_datetime` will return a `Series` if `data` is a `Series`, otherwise it returns a `DatetimeIndex`. This becomes a problem on line 151 because `DatetimeIndex` can directly call `strftime` but `Series` can only do it through the `.dt` attribute. We also should make sure a series is always returned, so it's best just to convert anything that isn't a `Series` here.

pvk-developer (approved):

PR #453, opened 2022-03-24
Update functionality when model_missing_values is False

### LINKED ISSUE #435: Update functionality when `model_missing_values=False`
### Problem Description

In any transformer that's capable of modeling missing values, setting this value to `False` leads to unintuitive behavior:
1. It takes note of the scheme used to replace missing values eg. using the mean or mode
2. On the reverse transform, it checks explicitly for the mean or mode and turns them back to missing values

This can lead to inaccurate results. Especially if you use the mode, many values will show up as missing even if they aren't supposed to be.

Additionally, it prints out a warning during the `fit` and `transform` stages:
```
UserWarning: Replacing nulls with existing value without `model_missing_values`,
which is not reversible. Use `model_missing_values=True` to ensure that the transformation
is reversible.
```

This is also not desirable because adding more columns leads to performance issues. We should not be encouraging users to set it to `True` unless absolutely necessary for quality.

### Expected behavior

If `model_missing_values=False`, we should do the following:
1. On forward transform, note down the % of values that are missing
2. On reverse transform, randomly add missing values according to the %

This way the missing values are not modeled, but they are still present in the right proportions

Additionally, we should remove the `UserWarning` that happens on the `fit` and `transform` calls.

### LINKED ISSUE #455: `NullTransformer` `reverse_transform` alters input data due to not copying
### Environment Details

* RDT version: v1.0.0.dev
* Python version: 3.8
* Operating System: PopOS! 20.24

### Error Description

Currently the `data.copy` occurs only when we are modeling the missing values, otherwise it does not.
Since `pandas` uses `numpy` views under the hood, it actually ends up altering the original `numpy.array` in memory which leads to altering the original input data.

### Steps to reproduce

```
import pandas as pd

from rdt.transformers import NullTransformer

data = pd.Series([1, 2, 3, np.nan])
nt = NullTransformer(missing_value_replacement='mean')

nt.fit(data)

transformed = nt.transform(data)
UserWarning: Replacing nulls with existing value without `model_missing_values`, which is not reversible. Use `model_missing_values=True` to ensure that the transformation is reversible.
  warnings.warn(IRREVERSIBLE_WARNING)

transformed
array([1., 2., 3., 2.])

nt.reverse_transform(transformed)

0    1.0
1    NaN
2    3.0
3    NaN
dtype: float64

transformed
array([ 1., nan,  3., nan])
```

amontanez24:
resolves #435
resolves #455

### DIFF
--- a/rdt/__init__.py
+++ b/rdt/__init__.py
@@ -18,6 +18,8 @@ __all__ = [
     'transformers'
 ]
 
+RANDOM_SEED = 42
+
 
 def get_demo(num_rows=5):
     """Generate demo data with multiple sdtypes.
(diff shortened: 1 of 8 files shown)


rdt/__init__.py
```diff
@@ -52,7 +52,8 @@ def get_demo(num_rows=5):
         return data.iloc[:num_rows]
 
     # Randomly generate the remaining rows
-    np.random.seed(42)
+    random_state = np.random.get_state()
+    np.random.set_state(np.random.RandomState(42).get_state())
```
katxiao, line 56: nit: could pull the 42 out as a constant at the top of the file

rdt/__init__.py
```diff
@@ -68,6 +69,8 @@ def get_demo(num_rows=5):
     dollars_spent = np.around(np.random.uniform(0, 100, size=num_rows), decimals=2)
     dollars_spent[np.random.random(size=num_rows) > 0.8] = np.nan
 
+    np.random.set_state(random_state)
```
katxiao, line 72: You could put this in a 
```
try:
     <code>
finally:
     np.random.set_state(random_state)
```
block to be extra sure that the random state is set to the original value.

katxiao (approved):
Accept with comments

PR #458, opened 2022-03-25
Learn datetime format and allow `%-` usage.

### LINKED ISSUE #437: Correct datetime format is not recovered on `reverse_transform`
### Environment Details

* RDT version: 1.0.0 (dev branch)

### Error Description

For both `UnixTimestampEncoder` and `OptimizedTimestampEncoder`: If you do not specify a `datetime_format` string, the transformers are still capable of understanding the values. However, during `reverse_transform`, they do not recover the same format.

The reverse transformed data should have the same datetime format as the original. We should learn the format during `fit` and then make sure we apply it during `reverse_transorm`.

### Steps to reproduce

```python
from rdt import get_demo
from rdt.transformers.datetime import UnixTimestampEncoder

# update demo data with a different datetime format
data2 = get_demo()
data2['last_login'] = ['Jun 26, 2021', 'Feb 02, 2021', np.nan, 'Sep 26, 2020', 'Dec 22, 2020']

# use a datetime transformer on this data: it's able to correctly read the new format
ute = UnixTimestampEncoder()
transformed = ute.fit_transform(data2, column='last_login')

# notice that the reversed data is in a different format: '2021-06-26', '2021-02-02', etc.
ute.reverse_transform(transformed)
```

### LINKED ISSUE #439: Can we support additional datetime formats?
### Problem Description

Right now, the `UnixTimestampEncoder` and `OptimizedTimestampEncoder` are able to accept a `datetime_format` string that contains the format codes listed in the [Python datetime docs](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes).

However, there are some extended parameters that `strftime` can support. See [cheat sheet](https://strftime.org/). Can we support these?

### Example

Python provides support for the `'%d'` code, which specifies a padded day of month. For example in `Feb 02, 2021`.

In the extended scheme, you can also provide `'%-d'`, for a non-padded day of month. For example `Feb 2, 2021`. However I cannot use the `'%-d'` code in our transformers.

```python
from rdt import get_demo
from rdt.transformers.datetime import UnixTimestampEncoder

# update demo data with a different datetime format
data2 = get_demo()
data2['last_login'] = ['Jun 26, 2021', 'Feb 2, 2021', np.nan, 'Sep 26, 2020', 'Dec 22, 2020']

# use a datetime transformer on this data: give it the explicit format string
ute = UnixTimestampEncoder(datetime_format='%b %-d, %Y')

# this crashes
ute.fit(data2, column='last_login')

ValueError: Data does not match specified datetime format.
```

pvk-developer:
Resolve #437 
Resolve #439

### DIFF
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -74,7 +74,11 @@ class UnixTimestampEncoder(BaseTransformer):
     def _convert_to_datetime(self, data):
         if data.dtype == 'object':
             try:
-                data = pd.to_datetime(data, format=self.datetime_format)
+                pandas_datetime_format = None
+                if self.datetime_format:
+                    pandas_datetime_format = self.datetime_format.replace('%-', '%')
+
+                data = pd.to_datetime(data, format=pandas_datetime_format)
 
             except ValueError as error:
                 if 'Unknown string format:' in str(error):
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -1,6 +1,7 @@
 """Transformer for datetime data."""
 import numpy as np
 import pandas as pd
+from pandas.core.tools.datetimes import guess_datetime_format
 
 from rdt.transformers.base import BaseTransformer
 from rdt.transformers.null import NullTransformer
@@ -112,6 +113,10 @@ class UnixTimestampEncoder(BaseTransformer):
             data (pandas.Series):
                 Data to fit the transformer to.
         """
+        if self.datetime_format is None and data.dtype == 'object':
+            date_string = data[data.notna()].astype(str)
+            self.datetime_format = guess_datetime_format(date_string.to_numpy()[0])
+
         transformed = self._transform_helper(data)
         self.null_transformer = NullTransformer(
             self.missing_value_replacement,
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -1,7 +1,7 @@
 """Transformer for datetime data."""
 import numpy as np
 import pandas as pd
-from pandas.core.tools.datetimes import guess_datetime_format
+from pandas.core.tools.datetimes import _guess_datetime_format_for_array
 
 from rdt.transformers.base import BaseTransformer
 from rdt.transformers.null import NullTransformer
@@ -118,8 +118,7 @@ class UnixTimestampEncoder(BaseTransformer):
                 Data to fit the transformer to.
         """
         if self.datetime_format is None and data.dtype == 'object':
-            date_string = data[data.notna()].astype(str)
-            self.datetime_format = guess_datetime_format(date_string.to_numpy()[0])
+            self.datetime_format = _guess_datetime_format_for_array(data.to_numpy())
 
         transformed = self._transform_helper(data)
         self.null_transformer = NullTransformer(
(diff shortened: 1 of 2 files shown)


rdt/transformers/datetime.py
```diff
@@ -73,7 +74,11 @@ def get_output_sdtypes(self):
     def _convert_to_datetime(self, data):
         if data.dtype == 'object':
             try:
-                data = pd.to_datetime(data, format=self.datetime_format)
+                pandas_datetime_format = None
+                if self.datetime_format:
+                    pandas_datetime_format = self.datetime_format.replace('%-', '%')
```
amontanez24, line 79: I don't think this works the desired way. `%-d` for example is supposed to mean you can omit the first digit of the date if it's not necessary. If someone has data like `2-2-2020`, they might give a format like `%-m-%-d-%Y`. If you replace the `-` then you end up with this as the format instead: `02-02-2020`. If you pass data and a datetime format like I originally said, I don't think this will work. If it parses, does it return data in the original format?

rdt/transformers/datetime.py
```diff
@@ -112,6 +117,9 @@ def _fit(self, data):
             data (pandas.Series):
                 Data to fit the transformer to.
         """
+        if self.datetime_format is None and data.dtype == 'object':
+            self.datetime_format = _guess_datetime_format_for_array(data.to_numpy())
```
amontanez24, line 121: Should we use the public method that this uses instead? https://github.com/pandas-dev/pandas/blob/258cfccf8106814504ca07a118037d3f32073732/pandas/core/tools/datetimes.py#L131
They basically just filter to make sure they aren't using a `nan` value to check. I don't know if it is risky to use hidden methods

rdt/transformers/datetime.py
pvk-developer: The `self.datetime_format` ends up 'formating' the string at the end, this line here is just for pandas to parse it to `datetime`. The reason why I'm replacing the `-` is that `pandas` raises an error (it doesn't support this kind of format to parse it to datetime). Since the `-` is just a `end-formating` type (strftime), then when returning on the revese transform the `-` will be applied to the data:

```python
In [1]: from rdt.transformers.datetime import UnixTimestampEncoder
In [2]: ute = UnixTimestampEncoder(datetime_format='%-m-%-d-%Y')
In [3]: data = pd.DataFrame({'date': ['02-02-2020', '1-10-2010', '05-5-2020']})
In [4]: transformed = ute.fit_transform(data, 'date')
In [5]: ute.reverse_transform(transformed)
Out[5]:
        date
0   2-2-2020
1  1-10-2010
2   5-5-2020
```

amontanez24 (approved):
Ready to go after adding unit test!

tests/unit/transformers/test_datetime.py
```diff
@@ -442,6 +464,31 @@ def test__reverse_transform_datetime_format(self):
...
+        ute = UnixTimestampEncoder(missing_value_replacement=None)
+        ute.datetime_format = '%b %-d, %Y'
+        transformed = np.array([1.5778368e+18, 1.5805152e+18, 1.5830208e+18])
+
+        # Run
+        output = ute._reverse_transform(transformed)
+
+        # Assert
+        expected = pd.Series(['Jan 1, 2020', 'Feb 1, 2020', 'Mar 1, 2020'])
+        if 'windows' not in platform.system().lower():
```
amontanez24, line 489: The code itself doesn't crash on windows right?

pvk-developer: Doesn't crash, however, it doesn't produce the expected `strftime`

PR #459, opened 2022-03-26
Fix setting with copy warning

### LINKED ISSUE #436: Update `FrequencyEncoder.transform` so that pandas won't throw a warning
### Environment Details

* RDT version: 0.6.4 and the new 1.0.0 (dev)

### Description

Every time you transform data using a `FrequencyEncoder`, there's a warning thrown by `pandas`

```
/usr/local/lib/python3.7/dist-packages/rdt/transformers/categorical.py:188: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  data[unseen_indexes] = np.random.choice(fit_categories, size=unseen_indexes.size)
```

If we start setting values using the pandas recommended approach, we can get rid of this warning.

pvk-developer:
Resolves #436

### DIFF
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -141,7 +141,7 @@ class BaseTransformer:
         if len(columns) == 1:
             columns = columns[0]
 
-        return data[columns]
+        return data[columns].copy()
 
     @staticmethod
     def _set_columns_data(data, columns_data, columns):


amontanez24 (approved):
LGTM! Let's not change the `pd.options.mode.chained_assignment` to  `None` unless it continues to be a problem

PR #467, opened 2022-04-01
Add more functionality to the Config class

### LINKED ISSUE #457: HyperTransformer Config class
Since a `Config` class has been introduced to `rdt.HyperTransformer` and following the suggestions on this comment: https://github.com/sdv-dev/RDT/pull/456#discussion_r834862616  we should consider moving some part of the logic thats currently held by the `HyperTransformer` to this `Config` class.

### Expected behavior

The `Config` class should be responsible for updating and validating the `field_transformers`, `_provided_field_transformers`, `field_sdtypes` and `_provided_field_sdtypes`.

### Additional context

The end user should be able to work with `rdt.HyperTransformer` as it is now, no new api changes are required.
The `Config` object shouldn't be necessary for the end user to create an instance or work with the `HyperTransformer`.
The `Config` object will be generated and stored in the `hypertransformer` as `self.config` or similar variable.
From the `config` we will perform the rest of the calls to get or use the transformers.

pvk-developer:
Resolve #457

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -147,14 +147,6 @@ class Config:
             if provided_transformer:
                 self._provided_field_transformers[column_name] = transformer
 
-    def get_field_transformers(self):
-        """Return the fields transformers."""
-        return self.field_transformers
-
-    def get_field_sdtypes(self):
-        """Return the fields sdtypes."""
-        return self.field_sdtypes
-
     def to_dict(self):
         """Return a `dict` object of `Config`."""
         return {
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -12,11 +12,13 @@ from rdt.transformers import (
     get_default_transformer, get_transformer_instance, get_transformers_by_type)
 
 
-class Config(dict):
-    """Config dict for ``HyperTransformer`` with a better representation."""
+class Config:
+    """Config class for ``HyperTransformer``."""
 
     _provided_field_transformers = None
     _provided_field_sdtypes = None
+    field_sdtypes = None
+    field_transformers = None
 
     @staticmethod
     def _validate_config(config):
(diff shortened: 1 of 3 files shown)


rdt/hyper_transformer.py
```diff
@@ -188,10 +286,7 @@ def get_config(self):
                 - sdtypes: A dictionary mapping column names to their ``sdtypes``.
                 - transformers: A dictionary mapping column names to their transformer instances.
         """
-        return Config({
-            'sdtypes': self.field_sdtypes,
-            'transformers': self.field_transformers
-        })
+        return self.config
```
amontanez24, line 289: One question is whether or not we should expose the actual config class to the user. Maybe we just return the dict as it used to be to them, but when they call `set_config` we create the special class.

rdt/hyper_transformer.py
```diff
@@ -417,14 +469,15 @@ def _learn_config(self, data):
...
+            if field not in self.config['field_sdtypes']:
                 self._set_field_sdtype(data, field)
-            if field not in self.field_transformers:
-                sdtype = self.field_sdtypes[field]
+            if field not in self.config['field_transformers']:
+                sdtype = self.config['field_sdtypes'][field]
                 if sdtype in self._default_sdtype_transformers:
-                    self.field_transformers[field] = self._default_sdtype_transformers[sdtype]
+                    default_transformer = self._default_sdtype_transformers[sdtype]
+                    self.config['field_transformers'][field] = default_transformer
```
amontanez24, line 478: I was thinking of adding methods to let the `HyperTransformer` interact with the `Config` instead of doing it directly. That way people don't accidently forget to update the `_provided` dicts if necessary. So maybe an `add_transformer(provided)` method that adds it to `config.field_transformers` and if `provided` is `True`, then also `config._provided_transformers`. Same for `field_sdtypes`

rdt/hyper_transformer.py
```diff
@@ -15,11 +15,132 @@
...
+                - transformers: A dictionary mapping column names to their transformer instances.
+        """
+        self._validate_config(config)
+        self._provided_field_sdtypes = config['sdtypes']
+        self['field_sdtypes'].update(config['sdtypes'])
+        self._provided_field_transformers = config['transformers']
+        self['field_transformers'].update(config['transformers'])
+
+    def __init__(self, config=None):
+        super().__init__()
```
amontanez24, line 58: The class probably no longer needs to inherit from `dict`. It could take a dict in the `__init__` and set things up accordingly. This would discourage people from using the class like a dict and encourage them to use the setter methods we create instead

amontanez24 (changes requested):
Looking good! Just a few comments about adding some setter methods and not exposing this class to the user

rdt/hyper_transformer.py
```diff
@@ -12,16 +12,155 @@
...
+
+        Update the `self.field_transformers` with the provided column name and transformer, if
+        `provided_transformer` is `True` will automatically update the
+        `self._provided_field_transformers` which will not reset.
+
+        Args:
+            column_name_to_transformer(dict):
+                Dict mapping column names to transformers to be used for that column.
+            provided_transformer(bool):
+                Wether or not to add to `self._provided_field_transformers`.
```
amontanez24, line 135: typo: whether

rdt/hyper_transformer.py
```diff
@@ -12,16 +12,155 @@
...
+                'pre-populate all the sdtypes and transformers from your dataset.'
+            )
+
+        for field, field_sdtype in self.field_sdtypes.items():
+            if field_sdtype == sdtype:
+                self._provided_field_transformers[field] = transformer
+                self.field_transformers[field] = transformer
+
+    def update_transformers(self, column_name_to_transformer, provided_transformer=False):
+        """Update `self.field_transformers`.
```
amontanez24, line 125: I think we agreed to do double ` characters instead of just one.

rdt/hyper_transformer.py
```diff
@@ -12,16 +12,155 @@
...
+    def __init__(self, config=None):
+        self._provided_field_sdtypes = {}
+        self._provided_field_transformers = {}
+        self.field_transformers = {}
+        self.field_sdtypes = {}
+        if config:
+            self.set_config(config)
+
+    def reset(self):
+        """Reset the `field_sdtypes` and `field_transformers`."""
```
amontanez24, line 68: use double `

rdt/hyper_transformer.py
amontanez24: I think this should return a custom version of the dict (what the config class used to be) that prints nicely. I don't think we want to expose this whole class to them since it might be confusing

pvk-developer:
We may rework this in the future.

PR #481, opened 2022-04-07
Validate the `transform` method

### LINKED ISSUE #479: User validation for `transform`
### Problem Description

When a user uses `transform`, we should validate their input & throw appropriate warnings/errors.

### Expected behavior

Perform the following 3 validation checks:

#### 1. Transforming data before fitting it

Note: This also applies if the HyperTransformer was previously fit but the config has changed since then.

```python
ht = HyperTransformer()
ht.transform(data)
Error: The HyperTransformer is not ready to use. Please fit your data first using 'fit' or 'fit_transform'.
```

#### 2. Transforming the data before setting a config
```python
ht = HyperTransformer()
ht.transform(data)
Error: No config detected. Set the config using 'set_config' or pre-populate it automatically from your data using 'detect_initial_config' prior to fitting your data.
```

#### 3. The data does not have the exact same column names as the config (& fit data)
See #474 

```python
ht = HyperTransformer()
ht.fit(data)
ht.transform(different_data)
Error: The data you are trying to transform has different columns than the original data. Column names and their sdtypes
must be the same. Use the method 'get_config()' to see the expected values. 
```

fealho:
Resolve #479.

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -583,6 +583,7 @@ class HyperTransformer:
             raise NotFittedError(
                 'The data you are trying to transform has different columns than the '
                 'original data. Column names and their sdtypes must be the same.'
+                "Use the method 'get_config()' to see the expected values."
             )
 
     def transform(self, data):
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -580,11 +580,9 @@ class HyperTransformer:
         unknown_columns = self._subset(data.columns, self._input_columns, not_in=True)
         is_subset = any(column not in data.columns for column in self._input_columns)
         if unknown_columns or is_subset:
-            unknown_text = f' (unknown columns: {unknown_columns})' if unknown_columns else ''
             raise NotFittedError(
-                'The data you are trying to transform has different columns than the original '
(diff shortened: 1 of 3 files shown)


rdt/hyper_transformer.py
```diff
@@ -565,17 +570,19 @@ def _validate_correctly_fitted(self, data):
...
+            )
 
         unknown_columns = self._subset(data.columns, self._input_columns, not_in=True)
         is_subset = any(column not in data.columns for column in self._input_columns)
         if unknown_columns or is_subset:
-            unknown_text = f' (unknown columns: {unknown_columns})' if unknown_columns else ''
             raise NotFittedError(
-                'The data you are trying to transform has different columns than the original '
-                f'fitted data{unknown_text}. Column names and their sdtypes must be the same. '
-                "Use the method 'get_config()' to see the expected values."
```
amontanez24, line 578: This line should still be included according to the issue

amontanez24 (changes requested):
One small text change, but once that's done it should be ready to go!

amontanez24 (approved):
LGTM!

pvk-developer (approved):
:shipit:

PR #484, opened 2022-04-11
Fix print

### LINKED ISSUE #468: Fix how get_config prints transformers
### Environment Details

* RDT version: v1.0.0
* Python version: Any
* Operating System: Any

### Error Description

When the output of the `get_config` method is printed (as the cell output), the dictionary contains the string representation of the transformer (surrounded by double quotes) instead of the transformer object.

Furthermore, the dictionary with strings is "copy/pasted" into a new dictionary and then passed to the `set_config` method, it crashes

pvk-developer:
Resolves #468


fealho (approved):

rdt/hyper_transformer.py
```diff
@@ -21,7 +21,14 @@ def __repr__(self):
             'sdtypes': self['sdtypes'],
             'transformers': {k: repr(v) for k, v in self['transformers'].items()}
         }
-        return json.dumps(config, indent=4)
+
+        printed = json.dumps(config, indent=4)
+        for transformer in self['transformers'].values():
+            quoted_transformer = f'"{transformer}"'
```
amontanez24, line 27: Is there any chance that the quotes would be single quotes? If so should we check for that?

amontanez24 (approved):
Just one questions, but otherwise looks good!

rdt/hyper_transformer.py
pvk-developer: I think not since the accepted standart by `JSON` is `"`.

PR #487, opened 2022-04-11
User validation for `update_transformers_by_sdtype`

### LINKED ISSUE #477: User validation for `update_transformers_by_sdtype`
### Problem Description

When a user uses `update_transformers_by_sdtype`, we should validate their input & throw appropriate warnings/errors.

### Expected behavior

Perform the following 4 validations when using this function.

#### 1. Users attempts to use `update_transformers_by_sdtype` after using fit

Throw a warning, same as #466 

```python
ht = HyperTransformer()
ht.fit()

ht.update_transformers_by_sdtype(
  sdtype='categorical',
  transformer=FrequencyEncoder
)
Warning: For this change to take effect, please refit your data using 'fit' or 'fit_transform'.
```

#### 2. Invalid sdtype
Throw an error if a user provides an sdtype that cannot be recognized
_If they are using open source RDT, then premium sdtypes will not be recognized either._

```python
ht.update_transformers_by_sdtype(
  sdtype='phone_number',
  transformer=FrequencyEncoder()
)
Invalid sdtype. If you are trying to use a premium sdtype, contact info@sdv.dev about RDT Add-Ons.
```

#### 3. Invalid transformer
The transformer must be a transformer object, transformer class instance or `None`.

```python
ht.update_transformers_by_sdtype(
  sdtype='numerical',
  transformer="FrequencyEncoder()"
)
Error: Invalid transformer. Please input an rdt transformer object.
```

#### 4. Transformer is not compatible with the sdtype

The transformer must be compatible. Note that it's ok to provide `transformer=None` for all sdtypes.

```python
ht.update_transformers_by_sdtype(
  sdtype='numerical',
  transformer=FrequencyEncoder()
)
Error: The transformer you've assigned is incompatible with the sdtype.
```

amontanez24:
resolves #477

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -9,7 +9,7 @@ import yaml
 
 from rdt.errors import Error, NotFittedError
 from rdt.transformers import (
-    get_default_transformer, get_transformer_instance, get_transformers_by_type)
+    BaseTransformer, get_default_transformer, get_transformer_instance, get_transformers_by_type)
 
 
 class Config(dict):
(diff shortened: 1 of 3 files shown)


pvk-developer (approved):
LGTM!

rdt/hyper_transformer.py
```diff
@@ -225,23 +225,32 @@ def update_transformers_by_sdtype(self, sdtype, transformer):
...
+        if self._fitted:
+            warnings.warn(self._REFIT_MESSAGE)
+
         if not self.field_sdtypes:
             raise Error(
                 'Nothing to update. Use the `detect_initial_config` method to '
                 'pre-populate all the sdtypes and transformers from your dataset.'
             )
 
+        if sdtype not in self._get_supported_sdtypes():
```
fealho, line 237: I don't think this handles RDT-plus, so even if it is installed, it will still raise an error.

amontanez24: Looks like we need to do 2 things:
1. Update the private transformers to use the correct naming
2. Clear the cache when searching transformers
I'll add the second to this PR

fealho (approved):

PR #492, opened 2022-04-14
Error when fitting on subset of config

amontanez24:
This PR fixes the requirement for #408 where fitting on a subset of the config's data should raise an error


fealho (approved):
Looks good

PR #493, opened 2022-04-15
Issue 475 update transformers

fealho:


tests/unit/test_hyper_transformer.py
```diff
@@ -2202,6 +2197,33 @@ def test_update_sdtypes_different_sdtype_transformer_provided(self, mock_warning
         assert instance.field_transformers == {'a': transformer}
         instance._user_message.assert_called_once_with(user_message, 'Info')
 
+    def test__validate_transformers(self):
```
amontanez24, line 2200: Can we add a line to `test_update_transformers` that makes sures the proper values are passed to `_validate_transformers` and that it is called?

amontanez24 (approved):
One small comment about adding to a unit test, but besides that I think this is good to go!

PR #494, opened 2022-04-15
Validate `set_config`

### LINKED ISSUE #478: User validation for `set_config`
### Problem Description

When a user uses `set_config`, we should validate their input & throw appropriate warnings/errors.

### Expected behavior

Perform the following 6 validations when using this function.

#### 1. Users attempts to use `set_config` after using fit

See #466 

#### 2. There must be an `sdtypes` dict and a `transformers` dict
Both dicts must be present. No other dicts are allowed.

```python
ht = HyperTransformer()
ht.set_config({})
Error: Invalid config. Please provide 2 dictionaries named 'sdtypes' and 'transformers'.
```

#### 3. Column names in `sdtypes` and `transformers` dict do not match

Both dicts should have the exact same set of column names as keys. (They do not have to be in the same order.)

```python
ht = HyperTransformer
ht.set_config({
  'sdtypes': {
    'column_A': 'numerical',
    'column_B': 'categorical'
  },
  'transformers': {
    'column_A': FloatFormatter(),
    'column_C': BinaryEncoder
  }
})
Error: The column names in the 'sdtypes' dictionary must match the column names in the 'transformers' dictionary.
```

#### 4. Invalid sdtypes
Throw an error if a user provides an sdtype that cannot be recognized
_If they are using open source RDT, then premium sdtypes will not be recognized either._

```python
ht.set_config({
  'sdtypes': {
    'column_A':  'unknown',
    'column_B': 'phone_number'
  }
  ...
})
Invalid sdtypes: ['unknown', 'phone_number']. If you are trying to use a premium sdtype, contact info@sdv.dev about RDT Add-Ons.
```

#### 5. Invalid transformers
```python
ht.set_config({
  'sdtypes': { ... }
  'transformers': {
    'column_A': "FrequencyEncoder",
    'column_B': 4.0
  }
})
Invalid transformers for columns: ['column_A', 'column_B']. Please assign an rdt transformer object to each column name.
```

#### 6. Transformers are not compatible with the sdtypes
```python
ht.set_config({
  'sdtypes': {
    'column_A': 'categorical', 
    'column_B': 'datetime'
  }, 'transformers': {
    'column_A': FloatFormatter(),
    'column_B': BinaryEncoder()
  }
})
Error: Some transformers you've assigned are not compatible with the sdtypes. Please change the following columns: ['column_A', 'column_B']
```

fealho:
Resolve #478.

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -216,6 +216,30 @@ class HyperTransformer:
             )
 
     @staticmethod
+    def _validate_sdtypes(sdtypes):
+        """Validate the given sdtypes are valid.
+
+        Args:
+            sdtypes (dict):
+                Dict mapping column names to sdtypes to be used for that column.
+
+        Raises:
+            Error:
+                Raises an error if ``sdtypes`` contains one or more invalid sdtype.
+        """
+        supported_sdtypes = HyperTransformer._get_supported_sdtypes()
+        unsupported_sdtypes = []
+        for sdtype in sdtypes.values():
+            if sdtype not in supported_sdtypes:
+                unsupported_sdtypes.append(sdtype)
+
+        if unsupported_sdtypes:
+            raise Error(
+                f'Invalid sdtypes: {unsupported_sdtypes}. If you are trying to use a '
+                'premium sdtype, contact info@sdv.dev about RDT Add-Ons.'
+            )
+
+    @staticmethod
     def _validate_config(config):
         if list(config.keys()) != ['sdtypes', 'transformers']:
             raise Error(
@@ -231,18 +255,7 @@ class HyperTransformer:
                 "column names in the 'transformers' dictionary."
             )
 
-        unsupported_sdtypes = []
-        supported_sdtypes = HyperTransformer._get_supported_sdtypes()
-        for sdtype in sdtypes.values():
-            if sdtype not in supported_sdtypes:
-                unsupported_sdtypes.append(sdtype)
-
-        if unsupported_sdtypes:
-            raise Error(
-                f'Invalid sdtypes: {unsupported_sdtypes}. If you are trying to use a '
-                'premium sdtype, contact info@sdv.dev about RDT Add-Ons.'
-            )
-
+        HyperTransformer._validate_sdtypes(sdtypes)
         HyperTransformer._validate_transformers(transformers)
 
         mismatched_columns = []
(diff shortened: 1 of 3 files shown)


rdt/hyper_transformer.py
```diff
@@ -171,18 +171,58 @@ def _field_in_data(field, data):
...
+                    invalid_transformers_columns.append(column_name)
+
+        if invalid_transformers_columns:
+            raise Error(
+                f'Invalid transformers for columns: {invalid_transformers_columns}. '
+                'Please assign an rdt transformer object to each column name.'
+            )
+
+        mismatched_columns = []
         for column, transformer in transformers.items():
```
amontanez24, line 214: Can we combine these two loops? Just populate the invalid and mismatched lists at the same time

tests/unit/test_hyper_transformer.py
```diff
@@ -1946,7 +2105,7 @@ def test_update_transformers_missmatch_sdtypes(self, mock_warnings):
         assert instance.field_transformers['my_column'] == mock_transformer
 
     @patch('rdt.hyper_transformer.warnings')
-    def test_update_sdtypes_fitted(self, mock_warnings):
+    def t(self, mock_warnings):
```
amontanez24, line 2108: fix this test name

amontanez24 (changes requested):

pvk-developer (approved):

amontanez24 (approved):
One small comment but otherwise looks good to me!

rdt/hyper_transformer.py
```diff
@@ -171,18 +171,57 @@ def _field_in_data(field, data):
...
-                if input_sdtype != sdtype:
-                    warnings.warn(
-                        f'You are assigning a {input_sdtype} transformer to a {sdtype} '
-                        f"column ('{column}'). If the transformer doesn't match the "
-                        'sdtype, it may lead to errors.'
-                    )
+                try:
+                    get_transformer_instance(transformer)
+                except ValueError:
+                    invalid_transformers_columns.append(column_name)
```
pvk-developer, line 207: There is a small bug, if we leave it as `ValueError` we are not contemplating other kind of exceptions, if we pass down an `object()` instance this will produce an `AttributeError` : 
`AttributeError: 'object' object has no attribute 'rsplit'`.

pvk-developer (changes requested):

rdt/hyper_transformer.py
amontanez24: Can we actually make this change in the `_validate_transformers` method you added in the other issue and just call that method here?

amontanez24 (approved):
I think this is ready!

pvk-developer (approved):

Issue #475, opened 2022-04-06
User validation for `update_transformers`


npatki:
### Problem Description

When a user uses `update_transformers`, we should validate their input & throw appropriate warnings/errors.

### Expected behavior

Perform the following 4 validations when using this function.

#### 1. Users attempts to use `update_transformers` after using fit
Throw a warning, same as #466 

```python
ht = HyperTransformer()
ht.detect_initial_config(data)
ht.fit()

ht.update_transformers(column_name_to_transformer={
  'column_A': FrequencyEncoder()
})
Warning: For this change to take effect, please refit your data using 'fit' or 'fit_transform'.
```

#### 2. Invalid column names

The column names must be in the config.

```python
ht.update_transformers(column_name_to_transformer={
  'unknown_column_1': FrequencyEncoder(),
  'unknown_column_2': LabelEncoder()
})
Error: Invalid column names: ['unknown_column_1', 'uknown_column_2']. These columns do not exist in the config. Use 'set_config' to write and set your entire config at once.
```

#### 3. Invalid transformers
The user must provide either a transformer object or `None`. 

```python
ht.update_transformers(column_name_to_transformer={
  'column_A': "FrequencyEncoder()",
  'column_B': 4.5
})
Error: Invalid transformers for columns: ['column_A', 'column_B']. Please assign an rdt transformer object to each column name.
```
#### 4. Transformers are not compatible with the sdtypes
Check the sdtypes and throw a warning in case of incompatibility.

```python
ht.update_transformers(column_name_to_transformer={
  'datetime_column': FrequencyEncoder(),
  'numerical_column': BinaryEncoder()
})
Warning: Some transformers you've assigned are not compatible with the sdtypes. Use 'update_sdtypes' to update: ['datetime_column', 'numerical_column']
```

PR #495, opened 2022-04-20
Fix columns missing in config and update transformers to None

pvk-developer:
* Fix columns missing in config  as part of #475 
* Fix `update_transformers` to set `transformers` to `None`

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -184,6 +184,14 @@ class HyperTransformer:
                         'sdtype, it may lead to errors.'
                     )
 
+    def _validate_update_columns(self, update_columns, config_columns):
+        unknown_columns = self._subset(update_columns, config_columns, not_in=True)
+        if unknown_columns:
+            raise Error(
+                f'Invalid column names: {unknown_columns}. These columns do not exist in the '
+                "config. Use 'set_config()' to write and set your entire config at once."
+            )
+
     @staticmethod
     def _get_supported_sdtypes():
         get_transformers_by_type.cache_clear()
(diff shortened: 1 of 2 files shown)


rdt/hyper_transformer.py
```diff
@@ -346,16 +346,25 @@ def update_transformers(self, column_name_to_transformer):
         if len(self.field_transformers) == 0:
             raise Error(self._DETECT_CONFIG_MESSAGE)
 
+        invalid_columns = list(
+            set(column_name_to_transformer).difference(set(self.field_transformers)))
+        if invalid_columns:
+            raise Error(
+                f'Invalid column names: {invalid_columns}. These columns do not exist in '
+                "the config. Use 'set_config' to write and set your entire config at once."
```
amontanez24, line 354: This validation and eror message also exists in `update_sdtypes`. Can we pull that out into a function `_validate_column_in_config` or something like that?

amontanez24 (commented):
One comment about pulling out reused functionality but besides that looks good to me!

rdt/hyper_transformer.py
```diff
@@ -346,14 +349,9 @@ def update_transformers(self, column_name_to_transformer):
...
-        invalid_columns = list(
-            set(column_name_to_transformer).difference(set(self.field_transformers)))
-        if invalid_columns:
-            raise Error(
-                f'Invalid column names: {invalid_columns}. These columns do not exist in '
-                "the config. Use 'set_config' to write and set your entire config at once."
-            )
-
+        update_columns = column_name_to_transformer.keys()
+        config_columns = self.field_transformers.keys()
```
amontanez24, line 353: In both cases this is only used for the validate method so we can just get the `config_columns` inside of the method and then we don't have to pass it along

amontanez24 (approved):
One more comment, but after that :shipit:

PR #503, opened 2022-05-05
Add noise parameter to `LabelEncoder`

### LINKED ISSUE #500: Add noise parameter to LabelEncoder
### Problem Description

Most distributions used by `GaussianCopula` and `CopulaGAN` fit better to continuous variables. In order to optimize the fitting process we can add noise to the output of the `LabelEncoder` to makes the transformed categorical variables continuous.

### Expected behavior

- Add a new boolean parameter, `add_noise`, to `__init__` for `LabelEncoder`.
    -  If value is false, no changes should be made
    - If value is true, then
        - On the forward transform: Perform the label encoding as usual, and then add uniform noise within the interval. For example:
            1. Label=1 is noised to anything in the interval [1, 2) → 1.002, 1.3, 1.9999, ..
            2. Label=2 is noised to anything in the interval [2, 3)
            3. Label=3 is noised to anything in the interval [3, 4)
        - On the reverse transform floor the values to the nearest integer (eg. 3.9 becomes 3, 4.321 becomes 4) and then continue the normal reverse transformation.

fealho:
Resolve #500.

### DIFF
--- a/tests/unit/transformers/test_categorical.py
+++ b/tests/unit/transformers/test_categorical.py
@@ -1683,7 +1683,7 @@ class TestLabelEncoder:
         # Setup
         transformer = LabelEncoder(add_noise=True)
         transformer.values_to_categories = {0: 'a', 1: 'b', 2: 'c'}
-        data = pd.Series([0, 1, 10])
+        data = pd.Series([0.5, 1.0, 10.9])
 
         # Run
         out = transformer._reverse_transform(data)


rdt/transformers/categorical.py
```diff
@@ -472,7 +478,10 @@ def _transform(self, data):
             size=is_null.sum()
         )
 
-        return mapped.astype('int64')
+        if self.add_noise:
+            mapped = np.random.uniform(mapped, mapped + 1)
```
amontanez24, line 482: since we already use the `pandas.Series.map` function, can we do this there and just pass a function that either uses the dictionary value if `add_noise` is false or uniformly samples from the dictionary value otherwise?

amontanez24 (commented):
Looking good so far. Just one comment about potentially doing 2 things at once

rdt/transformers/categorical.py
fealho: The issue with this change are lines 476-479:
```python3
mapped[is_null] = np.random.randint(
    len(self.categories_to_values),
    size=is_null.sum()
)
```

If I move the `self.add_noise`, these lines will have to be rewritten to something like:
```python3
if self.add_noise:
    mapped[is_null] = np.random.uniform(
        0,
        len(self.categories_to_values),
        size=is_null.sum()
    )
else:
    mapped[is_null] = np.random.randint(
        len(self.categories_to_values),
        size=is_null.sum()
    )
```
I think the speed gain is negligible compared to the loss in code readability.

pvk-developer (approved):
LGTM! 👍🏻

PR #504, opened 2022-05-05
updating license

amontanez24:


LICENSE
```diff
@@ -2,6 +2,8 @@ MIT License
 
 Copyright (c) 2018, MIT Data To AI Lab
 
+Copyright © 2022 DataCebo, Inc.
```
npatki, line 5: Will the © character properly show up in the license? This is is exactly what the doc has specified, but I guess we should keep the character consistent -- use (c) in both places or © in both places

LICENSE
```diff
@@ -20,3 +22,8 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
```
npatki, line 23: Out of curiosity, did you check to see if the rest of the wording is the same between the current License and the doc? That might be good to validate.

npatki (changes requested):

LICENSE
amontanez24: they're the same

LICENSE
amontanez24: Not sure, but I've seen `(c)` in other libraries so I'll do that one

npatki (approved):
LGTM!

PR #510, opened 2022-05-11
Add order_by parameter to LabelEncoder

### LINKED ISSUE #506: Add order_by parameter to LabelEncoder
### Problem Description

As a user, it would be useful to be able to specify the type of ordering I want to use for the `LabelEncoder`. 

### Expected behavior

- Add an init parameter called `order_by`
- It should be a string with the following possible values:
    - `None` (default): Just use the order the values appear in. Same as current implementation
    - `numerical_value`: Order by the numerical value (ie. the numerically first number should be mapped to 1, so on and so forth)
    - `alphabetical`: Order strings alphabetically (ie. the alphabetically first string should be mapped to 1, so on and so forth)

### Validation
- During init, if the parameter is not `None`, `'numerical_value'` or `'alphabetical'`, raise the following error:
    - `Error: order_by must be one of the following values: None, 'numerical_value' or 'alphabetical'`
- During `fit` if the `order_by` value is `'numerical_value'` but the data is not numerical, then raise the following error:
    - `Error: The data must be numerical or able to be casted as a float if order_by is 'numerical_value'`
- During `fit`, if the `order_by` value is `'alphabetical'` but the data is not a string, raise the following error:
    - `Error: The data must be of type string if order_by is 'alphabetical'`

amontanez24:
resolves #506

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -458,15 +458,18 @@ class LabelEncoder(BaseTransformer):
             if unique_data.dtype.type in [np.str_, np.object_]:
                 try:
                     unique_data.astype(np.float)
+                    unique_data = list(unique_data)
+                    unique_data.sort(key=float)
+                    unique_data = np.array(unique_data)
+
                 except ValueError as error:
                     raise Error(
                         'The data must be numerical or able to be casted as a float if order_by '
                         "is 'numerical_value'."
                     ) from error
 
-            unique_data = list(unique_data)
-            unique_data.sort(key=float)
-            unique_data = np.array(unique_data)
+            else:
+                unique_data = np.sort(unique_data)
 
         return unique_data
 
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -457,8 +457,10 @@ class LabelEncoder(BaseTransformer):
                 raise Error("The data must be numerical if order_by is 'numerical_value'.")
 
         if self.order_by is not None:
-            unique_data = np.sort(unique_data[~pd.isna(unique_data)])
-            unique_data = np.append(unique_data, [np.nan])
+            nans = pd.isna(unique_data)
+            unique_data = np.sort(unique_data[~nans])
+            if nans.any():
+                unique_data = np.append(unique_data, [np.nan])
 
         return unique_data
 
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -452,28 +452,24 @@ class LabelEncoder(BaseTransformer):
             if unique_data.dtype.type not in [np.str_, np.object_]:
                 raise Error("The data must be of type string if order_by is 'alphabetical'.")
 
-            unique_data = np.sort(unique_data)
+            nans = pd.isna(unique_data)
+            if nans.any():
+                nan_rep = unique_data[nans][0]
+                unique_data = np.sort(unique_data[~nans])
+                unique_data = np.append(unique_data, [nan_rep])
+
+            else:
+                unique_data = np.sort(unique_data)
 
         elif self.order_by == 'numerical_value':
-            error_message = (
-                'The data must be numerical or able to be casted as a float if order_by '
-                "is 'numerical_value'."
-            )
-            if unique_data.dtype.type in [np.str_, np.object_]:
-                try:
-                    unique_data.astype(np.float)
-                    unique_data = list(unique_data)
-                    unique_data.sort(key=float)
-                    unique_data = np.array(unique_data)
-
-                except ValueError as error:
-                    raise Error(error_message) from error
-
-            elif np.issubdtype(unique_data.dtype.type, np.number):
+            if np.issubdtype(unique_data.dtype.type, np.number):
                 unique_data = np.sort(unique_data)
 
             else:
-                raise Error(error_message)
+                raise Error(
+                    'The data must be numerical or able to be casted as a float if order_by '
+                    "is 'numerical_value'."
+                )
 
         return unique_data
 
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -7,6 +7,7 @@ import pandas as pd
 import psutil
 from scipy.stats import norm
 
+from rdt.errors import Error
 from rdt.transformers.base import BaseTransformer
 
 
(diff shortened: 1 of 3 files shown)


fealho (changes requested):
Looks good! I only left a couple of comments.

rdt/transformers/categorical.py
```diff
@@ -425,8 +437,41 @@ class LabelEncoder(BaseTransformer):
...
+                    unique_data.sort(key=float)
+                    unique_data = np.array(unique_data)
+
+                except ValueError as error:
+                    raise Error(
+                        'The data must be numerical or able to be casted as a float if order_by '
+                        "is 'numerical_value'."
+                    ) from error
+
+            else:
```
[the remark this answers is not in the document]
amontanez24: the problem is that booleans can be converted to floats without a problem so it still doesn't raise the error

fealho (approved):

pvk-developer (approved):
LGTM! :shipit:

tests/unit/transformers/test_categorical.py
```diff
@@ -1516,6 +1517,177 @@ def test__reverse_transform_1d(self):
...
+        # Setup
+        transformer = LabelEncoder(order_by='alphabetical')
+        arr = np.array(['one', 'two', 'three', np.nan, 'four'], dtype='object')
+
+        # Run
+        ordered = transformer._order_categories(arr)
+
+        # Assert
+        expected = np.array(['four', 'one', 'three', 'two', np.nan], dtype='object')
+        pd.testing.assert_series_equal(pd.Series(ordered), pd.Series(expected))
```
amontanez24, line 1589: I converted to `pd.Series` here because numpy can't tell if nans are equivalent

rdt/transformers/categorical.py
```diff
@@ -425,8 +437,41 @@ class LabelEncoder(BaseTransformer):
...
+
+        self.order_by = order_by
+
+    def _order_categories(self, unique_data):
+        if self.order_by == 'alphabetical':
+            if unique_data.dtype.type not in [np.str_, np.object_]:
+                raise Error("The data must be of type string if order_by is 'alphabetical'.")
+
+            nans = pd.isna(unique_data)
+            if nans.any():
```
fealho, line 456: I don't think you need the if statement.

rdt/transformers/categorical.py
```diff
@@ -425,8 +437,41 @@ class LabelEncoder(BaseTransformer):
...
+            else:
+                unique_data = np.sort(unique_data)
+
+        elif self.order_by == 'numerical_value':
+            if np.issubdtype(unique_data.dtype.type, np.number):
+                unique_data = np.sort(unique_data)
+
+            else:
+                raise Error(
+                    'The data must be numerical or able to be casted as a float if order_by '
```
fealho, line 470: The error message should be updated, since we don't support "caste-able" to float anymore.

rdt/transformers/categorical.py
```diff
@@ -425,8 +437,41 @@ class LabelEncoder(BaseTransformer):
...
         self.add_noise = add_noise
+        if order_by not in [None, 'alphabetical', 'numerical_value']:
+            raise Error(
+                "order_by must be one of the following values: None, 'numerical_value' or "
+                "'alphabetical'"
+            )
+
+        self.order_by = order_by
+
+    def _order_categories(self, unique_data):
```
fealho, line 450: With the new logic, I think you could reorganize the method like this:

```python3
    def _order_categories(self, unique_data):
        if self.order_by == 'alphabetical':
            if unique_data.dtype.type not in [np.str_, np.object_]:
                raise Error("The data must be of type string if order_by is 'alphabetical'.")

        elif self.order_by == 'numerical_value':
            if not np.issubdtype(unique_data.dtype.type, np.number):
                raise Error(
                    'The data must be numerical or able to be casted as a float if order_by '
                    "is 'numerical_value'."
                )

        if self.order_by is not None:
            nans = pd.isna(unique_data)
            nan_rep = unique_data[nans]
            unique_data = np.sort(unique_data[~nans])
            unique_data = np.append(unique_data, nan_rep)

        return unique_data
```
So validate the data first, then do the sorting.

fealho (changes requested):

rdt/transformers/categorical.py
amontanez24: If there are no nans then there is no reason to do any of this.

rdt/transformers/categorical.py
amontanez24: The reason I only do this in the 'alphabetical' section is that `np.sort` automatically sorts nans last with floats. It just doesn't work when it has to compare strings to nan, so in the string case I removed the nans and add it back in after the sorting

fealho: Sure, I just think the code could maybe be better organized. So instead of applying only to alphabetical, I applied to both to make the code cleaner. I don't think we lose much, if any, speed by doing it this way (since the only difference is that we manually put the nans at the end instead of letting np handle it). Removing the `if nans.any():` is the same thing, I don't think it gains any speed and it just makes the code slightly more complicated, which is why I think it could be removed.

Either way works fine though, the performance and results should be the same.

pvk-developer (approved):

fealho:
@amontanez24 The way you implemented, only 1 `np.nan` is added at the end, regardless of how many were present in the data. Is that the intended behavior? I thought we wanted to put them all at the end of the array.

Also, you gotta update the error message. After that I think it's good to merge.

rdt/transformers/categorical.py
```diff
@@ -425,8 +437,33 @@ class LabelEncoder(BaseTransformer):
...
+        elif self.order_by == 'numerical_value':
+            if not np.issubdtype(unique_data.dtype.type, np.number):
+                raise Error(
+                    'The data must be numerical or able to be casted as a float if order_by '
+                    "is 'numerical_value'."
+                )
+
+        if self.order_by is not None:
+            unique_data = np.sort(unique_data[~pd.isna(unique_data)])
+            unique_data = np.append(unique_data, [np.nan])
```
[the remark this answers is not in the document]
amontanez24: I think there is a misunderstanding of what this method does. It only takes in unique values as it is ordering the keys for the `self.values_to_categories` dict. The result of this method should be a list of all unique categories sorted, and with nan appearing last. That is why previously I was only adding it if there were actually nans, but truthfully there is no harm in always adding it

fealho (approved):

rdt/transformers/categorical.py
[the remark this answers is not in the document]
amontanez24: Actually I will change it to only add if there is a nan because the spec is that any categories not seen in `fit` (including nans) should not be transformable

PR #511, opened 2022-05-12
Remove logic to round integers in `FloatFormatter`

### LINKED ISSUE #508: Only round to decimal places in FloatFormatter
### Problem Description

As a user, it is confusing that my data might automatically be rounded to the nearest tens place, hundreds place etc. Unless explicitly asked for, it should only round the decimal places.

### Expected behavior

- When `learn_rounding_scheme` = `True`, only capture the number of digits after the decimal point. Remove the logic that rounds to places greater than the ones place.

fealho:
Resolve #508.

### DIFF
--- a/tests/unit/transformers/test_numerical.py
+++ b/tests/unit/transformers/test_numerical.py
@@ -271,32 +271,6 @@ class TestFloatFormatter(TestCase):
         # Asserts
         assert transformer._rounding_digits == 4
 
-    def test__fit_learn_rounding_scheme_true_large_numbers(self):
-        """Test ``_fit`` with ``learn_rounding_scheme`` set to ``True`` on large numbers.
-
-        If the ``learn_rounding_scheme`` parameter is set to ``True`` and the data is
-        very large, ``_fit`` should learn ``_rounding_digits`` to be None.
-
-        Input:
-        - Series of data with numbers between 10^10 and 10^20
-        Side Effect:
-        - ``_rounding_digits`` is set to None
-        """
-        # Setup
-        exponents = [np.random.randint(10, 20) for i in range(10)]
-        big_numbers = [10**exponents[i] for i in range(10)]
-        data = pd.Series(big_numbers)
-
-        # Run
-        transformer = FloatFormatter(
-            missing_value_replacement='mean',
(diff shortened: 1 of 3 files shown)


tests/unit/transformers/test_numerical.py
```diff
@@ -377,19 +370,18 @@ def test__fit_learn_rounding_scheme_true_max_zero(self):
         transformer._fit(data)
 
         # Asserts
-        assert transformer._rounding_digits == 0
+        assert transformer._rounding_digits is None
 
     def test__fit_learn_rounding_scheme_true_max_negative(self):
```
amontanez24, line 375: I don't think this test is even necessary anymore

tests/unit/transformers/test_numerical.py
```diff
@@ -277,15 +273,14 @@ def test__fit_learn_rounding_scheme_true(self):
 
     def test__fit_learn_rounding_scheme_true_large_numbers(self):
```
amontanez24, line 274: This test also probably isn't necessary anymore

tests/unit/transformers/test_numerical.py
```diff
@@ -333,14 +328,12 @@ def test__fit_learn_rounding_scheme_true_inf(self):
         """Test ``_fit`` with ``learn_rounding_scheme`` set to ``True``.
 
         If the ``learn_rounding_scheme`` parameter is set to ``True``, and the data
-        contains values that are infinite, ``_fit`` should learn the ``_rounding_digits``
-        to be the min number of decimal places seen in the data with the infinite values
-        filtered out.
+        contains values that are infinite, ``_fit`` shouldn't round.
```
amontanez24, line 331: This is misleading because it should round if there are infinite values so long as there are actual decimals. This test was just making sure that when we used to round to higher places, the infinite values were ignored.

amontanez24 (commented):
Some of the tests could probably just be removed as they aren't really testing changes in the logic anymore

katxiao (approved):
lgtm after Andrew's comments are addressed

amontanez24 (approved):
:shipit:

PR #512, opened 2022-05-13
Implement RegexGenerator

### LINKED ISSUE #505: Add RegexGenerator Transformer
### Problem Description

As a user, it would be useful to have a transformer that can handle text columns that follow a certain regex pattern. 

### Expected behavior

- Add a transformer called `RegexGenerator`
- The input `sdtype` should be a new `sdtype` called `text`
- It should have the following init parameters:
   - `regex_format`: a regex string that will be used to generate regex value. By default, use the following regex: `[A-Za-z]{4}`
- On the forward transform, the original column should just be dropped.
- On reverse generate random text that follows the regex
- Add a text `DatasetGenerator` for performance testing

### Additional context

- Investigate open source libraries that can help with this
- One option: [xeger](https://github.com/crdoconnor/xeger)

pvk-developer:
Resolves #505

### DIFF
--- a/rdt/transformers/pii/anonymizer.py
+++ b/rdt/transformers/pii/anonymizer.py
@@ -162,7 +162,7 @@ class AnonymizedFaker(BaseTransformer):
 
         return None
 
-    def _reverse_transform(self, data, sample_size=None):
+    def _reverse_transform(self, data):
         """Generate new anonymized data using a ``faker.provider.function``.
 
         Args:
--- a/rdt/transformers/pii/anonymizer.py
+++ b/rdt/transformers/pii/anonymizer.py
@@ -145,7 +145,7 @@ class AnonymizedFaker(BaseTransformer):
         self.data_length = len(data)
 
     def _transform(self, data):
-        """Drop the column and return ``null`` column if ``models_missing_values``.
+        """Return ``null`` column if ``models_missing_values``.
 
         Args:
             data (pandas.Series):
@@ -162,7 +162,7 @@ class AnonymizedFaker(BaseTransformer):
 
         return None
 
-    def _reverse_transform(self, data):
+    def _reverse_transform(self, data, sample_size=None):
         """Generate new anonymized data using a ``faker.provider.function``.
 
         Args:
@@ -172,9 +172,11 @@ class AnonymizedFaker(BaseTransformer):
         Returns:
             pandas.Series
         """
+        sample_size = len(data) or self.data_length
+
         reverse_transformed = np.array([
             self._function()
-            for _ in range(self.data_length)
+            for _ in range(sample_size)
         ], dtype=object)
 
         if self.null_transformer.models_missing_values():
--- a/rdt/transformers/pii/anonymizer.py
+++ b/rdt/transformers/pii/anonymizer.py
@@ -172,7 +172,10 @@ class AnonymizedFaker(BaseTransformer):
         Returns:
             pandas.Series
         """
-        sample_size = len(data) or self.data_length
+        if data is not None and len(data):
+            sample_size = len(data)
+        else:
+            sample_size = self.data_length
 
         reverse_transformed = np.array([
             self._function()
--- a/rdt/transformers/text.py
+++ b/rdt/transformers/text.py
@@ -36,6 +36,7 @@ class RegexGenerator(BaseTransformer):
 
     def __init__(self, regex_format='[A-Za-z]{5}', missing_value_replacement=None,
                  model_missing_values=False):
+        self.data_length = None
         self.missing_value_replacement = missing_value_replacement
         self.model_missing_values = model_missing_values
         self.regex_format = regex_format
--- a/rdt/transformers/text.py
+++ b/rdt/transformers/text.py
@@ -1,4 +1,5 @@
 """Transformers for text data."""
+import warnings
 
 import numpy as np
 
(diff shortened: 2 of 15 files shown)


tests/unit/transformers/pii/test_anonymizer.py
```diff
@@ -236,7 +236,7 @@ def test_get_output_sdtypes(self):
         """Test the ``get_output_sdtypes``.
 
         Setup:
-            - initialize a ``AnonymizedFaker`` transformer which:
+            - initialize a ``AnonymizedFaker`` transformer with default values.
```
amontanez24, line 239: good catch

rdt/transformers/text.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        self.null_transformer = NullTransformer(
+            self.missing_value_replacement,
+            self.model_missing_values
+        )
+        self.null_transformer.fit(data)
+        self.data_length = len(data)
+
+    def _transform(self, data):
+        """Drop the column and return ``null`` column if ``models_missing_values``.
```
amontanez24, line 72: I would just say 'Return ``null`` column if ``model_missing_values``.', because nothing is actually being dropped

rdt/transformers/text.py
```diff
@@ -0,0 +1,118 @@
...
+                Data to transform.
+
+        Returns:
+            pandas.Series
+        """
+        generator, size = strings_from_regex(self.regex_format)
+        if size > self.data_length:
+            reverse_transformed = np.array([
+                next(generator)
+                for _ in range(self.data_length)
```
amontanez24, line 103: It seems like we are matching the number of rows specified by `self.data_length` which is set in `fit`. What happens if `reverse_transform` is called on a dataset with more or less rows than what we fitted? In the synthetic data case, the number of sampled rows is not necessarily the same as the number of rows used for modeling

tests/unit/transformers/test_text.py
```diff
@@ -0,0 +1,383 @@
...
+            'C',
+            'D',
+            'E',
+            'A',
+            'B',
+            'C',
+            'D',
+            'E',
+            'A'
+        ])
```
amontanez24, line 317: any reason not to put these in one line?

tests/unit/transformers/test_text.py
```diff
@@ -0,0 +1,383 @@
...
+            'B',
+            'C',
+            'D',
+            'E',
+            'A',
+            'B',
+            'C',
+            'D',
+            'E',
+            'A'
```
amontanez24, line 298: same question

tests/unit/transformers/test_text.py
```diff
@@ -0,0 +1,383 @@
...
+        """Test the ``_reverse_transform`` method.
+
+        Validate that the ``_reverse_transform`` method calls the ``strings_from_regex``
+        function using the ``instance.regex_format`` and then generates the
+        ``instance.data_length`` number of data.
+
+        Setup:
+            - Initialize a ``RegexGenerator`` instance.
+            - Set ``data_length`` to 4.
+            - Initialie a generator.
```
amontanez24, line 334: minor: type `initialie`

tests/unit/transformers/test_text.py
```diff
@@ -0,0 +1,383 @@
...
```
amontanez24, line 272: typo

tests/unit/transformers/test_text.py
```diff
@@ -0,0 +1,383 @@
...
```
amontanez24, line 234: typo

tests/unit/transformers/test_text.py
```diff
@@ -0,0 +1,383 @@
...
+        result = instance._reverse_transform(columns_data)
+
+        # Assert
+        expected_null_call = np.array(['A', 'B', 'C'])
+        null_call = instance.null_transformer.reverse_transform.call_args_list[0][0][0]
+        np.testing.assert_array_equal(null_call, expected_null_call)
+        np.testing.assert_array_equal(result, np.array(['A', 'B', 'C']))
+
+    @patch('rdt.transformers.text.strings_from_regex')
+    def test__reverse_transform_generator_size_smaller_than_data_legnth(self,
```
amontanez24, line 261: typo: `data_legnth`

tests/unit/transformers/test_text.py
```diff
@@ -0,0 +1,383 @@
...
+
+        # Run
+        result = instance._transform(columns_data)
+
+        # Assert
+        instance.null_transformer.transform.assert_called_once_with(columns_data)
+        np.testing.assert_array_equal(result, np.array([0, 1, 0]))
+
+    @patch('rdt.transformers.text.strings_from_regex')
+    def test__reverse_transform_generator_size_bigger_than_data_legnth(self,
```
amontanez24, line 223: typo: `data_legnth`

amontanez24 (changes requested):

rdt/transformers/pii/anonymizer.py
```diff
@@ -162,7 +162,7 @@ def _transform(self, data):
 
         return None
 
-    def _reverse_transform(self, data):
+    def _reverse_transform(self, data, sample_size=None):
```
amontanez24, line 165: we can get rid of this parameter right? we're just using the data's length

amontanez24 (approved):
One comment but after that this looks good to go!

rdt/transformers/pii/anonymizer.py
pvk-developer: Forgot to remove it from the `anonymizer`

fealho (approved):
Looks good! 🎊

PR #516, opened 2022-05-17
Add CustomLabelEncoder transformer

### LINKED ISSUE #507: Add CustomLabelEncoder transformer
### Problem Description

As a user, it would be useful to have a `LabelEncoder` that uses a custom ordering that I supply.

### Expected behavior

- Add a new transformer called `CustomLabelEncoder`
- This transformer should have the following init parameters:
    - `order`: A list of categories that will appear in the data. The order of this list determines the order of the labels.
    - `add_noise`: Adds uniform noise in the same way as the normal `LabelEncoder`
- Forward transform should work same as normal `LabelEncoder` except use the provided ordering for the labels
- Reverse transform should also work the same

### Validation
- During `fit` if the data contains categories not specified in the list provided in `order` then raise the following error:
    - `Error: Unknown category '{category}'. All possible categories must be defined in the 'order' parameter.`

amontanez24:
resolves #507 

This PR:

- Adds a new `CustomLabelEncoder` class
- Overwrites the class' `fit` method
- Adds unit tests
- Adds integration test
- Skips performance tests, integration tests and quality tests for this transformer since the `order` param cannot be automated for now

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -468,8 +468,8 @@ class LabelEncoder(BaseTransformer):
         """Fit the transformer to the data.
 
         Generate a unique integer representation for each category and
-        store them in the `categories_to_values` dict and its reverse
-        `values_to_categories`.
+        store them in the ``categories_to_values`` dict and its reverse
+        ``values_to_categories``.
 
         Args:
             data (pandas.Series):
@@ -540,7 +540,7 @@ class LabelEncoder(BaseTransformer):
 class CustomLabelEncoder(LabelEncoder):
     """Custom label encoder for categorical data.
 
-    This class works very similarly to the ``LabelEncoder``, except that is requires the ordering
+    This class works very similarly to the ``LabelEncoder``, except that it requires the ordering
     for the labels to be provided.
 
     Null values are considered just another category.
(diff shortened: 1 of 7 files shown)


amontanez24:
Still need to add integration tests

rdt/transformers/categorical.py
```diff
@@ -535,3 +535,49 @@ def _reverse_transform(self, data):
...
+
+        Generate a unique integer representation for each category and
+        store them in the `categories_to_values` dict and its reverse
+        `values_to_categories`.
+
+        Args:
+            data (pandas.Series):
+                Data to fit the transformer to.
+        """
+        missing = list(data[~data.isin(self.order)].unique())
```
fealho, line 572: This line is converting None to nan, which is breaking things when Nones are involved. Example:

```python3
import pandas as pd
import numpy as np
from rdt.transformers import CustomLabelEncoder

data = pd.DataFrame([None, 1], columns=['column_name'])
transformer = CustomLabelEncoder(order=[None, 1])
transformer.fit(data, 'column_name')
transformed = transformer.transform(data)
```

tests/integration/transformers/test_categorical.py
```diff
@@ -348,3 +348,26 @@ def test_label_encoder_order_by_alphabetical():
     expected = pd.DataFrame([1, 3, 4, 2, 0], columns=['column_name.value'])
     pd.testing.assert_frame_equal(transformed, expected)
     pd.testing.assert_frame_equal(reverse, data)
+
+
+def test_custom_label_encoder():
+    """The the CustomLabelEncoder.
```
katxiao, line 354: Should be "Test the CustomLabelEncoder"

Nit: The comment could be more descriptive

tests/unit/transformers/test_categorical.py
```diff
@@ -1856,3 +1857,70 @@ def test__reverse_transform_add_noise(self):
 
         # Assert
         pd.testing.assert_series_equal(out, pd.Series(['a', 'b', 'c']))
+
+
+class TestCustomLabelEncoder:
+
+    def test___init__(self):
+        """Passed arguments must be stored as attributes."""
```
katxiao, line 1865: nit: Should we follow the convention of starting the description with "Test the ``__init__`` method"?

rdt/transformers/categorical.py
amontanez24: I actually can't replicate this. To be safe I can cast all nan types to `np.nan` like in the normal `LabelEncoder`

fealho (approved):
👍

katxiao (approved):

PR #523, opened 2022-07-08
Add `PseudoAnonymizedFaker` transformer

### LINKED ISSUE #517: Add `PseudoAnonymizedFaker` transformer
### Problem Description

As a user, I would like the option to pseudo anonymize my data instead of a complete anonymization.

_Note: **Pseudo**-anonymization means that the anonymized values can be traced back to the original, for eg. by using a mapping._ 

### Expected behavior

Input params, warnings and error states are exactly the same as [`AnonymizedFaker`](https://github.com/sdv-dev/RDT/issues/397). The functionality is slightly different:

- **fit**: Create a mapping of real data --> fake data using the Faker provider/function pair.
- **Forward transform:** Map the data first (throw an error if there are new values that weren't seen during the fit phase). Then use label encoding + noise on each of the values, as if they are categorical. (The SDV can model them)
- **Reverse transform:** Reverse transform the noised labels back to the mapped value. Return the mapped value.

### Additional context
The user should be able to access the mapping after they have `fit` the data:
- `<transformer_object>.get_mapping()` should return a dictionary `{ original_value : mapped_value, ... }`

Note: The user should be able to use `reverse_transform` even without access to the mapping. This will become important when we integrate this with the SDV.

### Additional Error Checking

Due to the mapping, there are some additional warnings/errors:

**Saving the transformer:** Warning because you're saving PII

```python
Warning: You are saving the mapping information, which includes the original data.
Sharing this object with others will also give them access to the original data '
used with this transformer.
```

**Unable to generate enough fake values:** The faker may not be able to generate enough unique values

```python
transformer = PsuedoAnonymizedFaker(provider_name='misc', function_name='boolean')
transformer.fit(data, 'column_name')
Error: The Faker function you specified is only able to generate 2 unique values, which is not enough to create a mapping. Please use a different Faker function for this column.
```

**New values appear when transforming:** The faker forms a mapping during `fit`. Unexpected values should not be present during the `transform`.

```python
transformer = PsuedoAnonymizedFaker(provider_name='credit_card', function_name='credit_card_full')
transformer.fit(data, 'column_name')
transformer.transform(invalid_data)
Error: The data you are transforming has new, unexpected values ('a', 'b', 'c' and 5 more). Please fit the transformer again using this new data.
```

pvk-developer:
Resolves #517

### DIFF
--- a/rdt/transformers/pii/anonymizer.py
+++ b/rdt/transformers/pii/anonymizer.py
@@ -8,7 +8,6 @@ from copy import deepcopy
 
 import faker
 import numpy as np
-import pandas as pd
 
 from rdt.errors import Error
 from rdt.transformers.base import BaseTransformer
@@ -51,6 +50,7 @@ class AnonymizedFaker(BaseTransformer):
     DETERMINISTIC_TRANSFORM = False
     DETERMINISTIC_REVERSE = False
     INPUT_SDTYPE = 'pii'
+    OUTPUT_SDTYPES = {}
     null_transformer = None
 
     @staticmethod
@@ -129,7 +129,7 @@ class AnonymizedFaker(BaseTransformer):
             dict:
                 Mapping from the transformed column names to supported sdtypes.
         """
-        output_sdtypes = {}
+        output_sdtypes = self.OUTPUT_SDTYPES
         if self.null_transformer and self.null_transformer.models_missing_values():
             output_sdtypes['is_null'] = 'float'
 
@@ -236,6 +236,12 @@ class PseudoAnonymizedFaker(AnonymizedFaker):
             List of localized providers to use instead of the global provider.
     """
 
+    MAX_ATTEMPTS_FOR_UNIQUE_VALUES = 10
+    OUTPUT_SDTYPES = {'value': 'categorical'}
+    NEXT_TRANSFORMER = {
+        'value': LabelEncoder(add_noise=True)
+    }
+
     def __getstate__(self):
         """Return a dictionary representation of the instance and warn the user when pickling."""
         LOGGER.warning((
@@ -262,16 +268,6 @@ class PseudoAnonymizedFaker(AnonymizedFaker):
         """Return the mapping dictionary."""
         return deepcopy(self._mapping_dict)
 
-    def get_output_sdtypes(self):
-        """Return the output sdtypes supported by the transformer.
-
-        Returns:
-            dict:
-                Mapping from the transformed column names to supported sdtypes.
-        """
-        output_sdtypes = {'value': 'float'}
-        return self._add_prefix(output_sdtypes)
-
     def _fit(self, columns_data):
         """Fit the transformer to the data.
 
@@ -283,7 +279,6 @@ class PseudoAnonymizedFaker(AnonymizedFaker):
             data (pandas.Series):
                 Data to fit the transformer to.
         """
-        super()._fit(columns_data)
         unique_values = columns_data[columns_data.notna()].unique()
         unique_data_length = len(unique_values)
         generated_values = [self._function() for _ in range(unique_data_length)]
@@ -292,7 +287,7 @@ class PseudoAnonymizedFaker(AnonymizedFaker):
             while len(set(generated_values)) != unique_data_length:
                 generated_values.append(self._function())
                 counter += 1
-                if counter == 10:
+                if counter == self.MAX_ATTEMPTS_FOR_UNIQUE_VALUES:
                     error_msg = (
                         'The Faker function you specified is only able to generate '
                         f'{len(set(generated_values))} unique values, which is not enough to '
@@ -341,13 +336,11 @@ class PseudoAnonymizedFaker(AnonymizedFaker):
 
             raise ValueError(error_msg)
 
-        mapped_data = pd.DataFrame(columns_data.map(self._mapping_dict))
-        self._label_encoder = LabelEncoder(add_noise=True)
-        transformed = self._label_encoder.fit_transform(mapped_data, self.get_input_column())
-        return transformed[self._label_encoder.get_output_columns()]
+        mapped_data = columns_data.map(self._mapping_dict)
+        return mapped_data
 
     def _reverse_transform(self, columns_data):
-        """Convert float values back to the mapped categorical values.
+        """Return the input data.
 
         Args:
             data (pd.Series or numpy.ndarray):
@@ -356,6 +349,4 @@ class PseudoAnonymizedFaker(AnonymizedFaker):
         Returns:
             pandas.Series
         """
-        data = pd.DataFrame(columns_data)
-        columns_data = self._label_encoder.reverse_transform(data)
-        return columns_data[self.get_input_column()]
+        return columns_data
--- a/rdt/transformers/pii/anonymizer.py
+++ b/rdt/transformers/pii/anonymizer.py
@@ -2,15 +2,20 @@
 
 import importlib
 import inspect
+import logging
 import warnings
 from copy import deepcopy
 
 import faker
 import numpy as np
+import pandas as pd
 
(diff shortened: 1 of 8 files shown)


rdt/transformers/utils.py
```diff
@@ -42,7 +42,7 @@ def _max_repeat(options, max_repeat):
         max_ = max_repeat
 
     option, args = options[0]
-    generator, size = _GENERATORS[option](args, max_repeat)
+    generator, size = _GENERATORS[option](args, max_repeat)  # pylint: disable=W0612
```
pvk-developer, line 45: Pylint started complaining about this, however `generator` is being used in the `for for loop` this gets raised anyways...

rdt/transformers/pii/anonymizer.py
```diff
@@ -209,3 +214,148 @@ def __repr__(self):
...
+        super()._fit(columns_data)
+        unique_values = columns_data[columns_data.notna()].unique()
+        unique_data_length = len(unique_values)
+        generated_values = [self._function() for _ in range(unique_data_length)]
+        if len(set(generated_values)) != unique_data_length:
+            counter = 0
+            while len(set(generated_values)) != unique_data_length:
+                generated_values.append(self._function())
+                counter += 1
+                if counter == 10:
```
katxiao, line 295: Why is this capped at 10?

rdt/transformers/utils.py
katxiao, line 45: I don't know if *this* `generator` is being used, because in the return nested for loop, `generator` is a local variable that iterates through `generators`.

rdt/transformers/pii/anonymizer.py
```diff
@@ -209,3 +214,148 @@ def __repr__(self):
...
+                error_msg = (
+                    'The data you are transforming has new, unexpected values '
+                    f'({new_values} and {diff} more). Please fit the transformer again '
+                    'using this new data.'
+                )
+
+            raise ValueError(error_msg)
+
+        mapped_data = pd.DataFrame(columns_data.map(self._mapping_dict))
+        self._label_encoder = LabelEncoder(add_noise=True)
```
amontanez24, line 345: Instead of putting the `LabelEncoder` in here, can we set the `NEXT_TRANSFORMERS` attribute? This was meant to be a way for one transformer to set the next transformer to use for any of its outputs

rdt/transformers/pii/anonymizer.py
```diff
@@ -209,3 +214,148 @@ def __repr__(self):
...
+        return deepcopy(self._mapping_dict)
+
+    def get_output_sdtypes(self):
+        """Return the output sdtypes supported by the transformer.
+
+        Returns:
+            dict:
+                Mapping from the transformed column names to supported sdtypes.
+        """
+        output_sdtypes = {'value': 'float'}
```
amontanez24, line 272: Can we just set `self.OUTPUT_SDTYPES` to this dict? The base class has a method that does the next line already

rdt/transformers/pii/anonymizer.py
amontanez24: yeah, if this is an arbitrary number can we set it as a class attribute like `MAX_ATTEMPTS_FOR_UNIQUE_VALUES` or something like that

rdt/transformers/pii/anonymizer.py
```diff
@@ -209,3 +214,148 @@ def __repr__(self):
...
+
+        Generate a ``_mapping_dict`` and a ``_reverse_mapping_dict`` for each
+        value in the provided ``columns_data`` using the ``Faker`` provider and
+        ``function``.
+
+        Args:
+            data (pandas.Series):
+                Data to fit the transformer to.
+        """
+        super()._fit(columns_data)
```
amontanez24, line 286: what is the parent doing in its fit?

amontanez24 (changes requested):

rdt/transformers/pii/anonymizer.py
pvk-developer: Yes this was a little bit 'improvised'. This piece of the code could change in a different ways, and I also had different ideas.

The current one is basically adding new values to `generated_values` and tries 10 times, if it fails that amount of times (which I agree we can set a `MAX_ATTEMPTS_FOR_UNIQUE_VALUES`), the other approach is to generate the same amount of data and see if the `set` is `>= unique_data_length` and then just use `list(set(generated_values))[:unique_data_length]` . (I feel that the second one is less efficient when working with huge amounts of data that's why I decided to go with the first option).

rdt/transformers/pii/anonymizer.py
pvk-developer: Actually fits the `null_transformer` and I forgot to add it to the `reverse_transform` 🤦🏻‍♂️

rdt/transformers/pii/anonymizer.py
```diff
@@ -209,3 +214,139 @@ def __repr__(self):
...
+            while len(set(generated_values)) != unique_data_length:
+                generated_values.append(self._function())
+                counter += 1
+                if counter == self.MAX_ATTEMPTS_FOR_UNIQUE_VALUES:
+                    error_msg = (
+                        'The Faker function you specified is only able to generate '
+                        f'{len(set(generated_values))} unique values, which is not enough to '
+                        'create a mapping. Please use a different Faker function for this column.'
+                    )
+                    raise ValueError(error_msg)
```
amontanez24, line 296: It looks like faker has a way to guarantee uniqueness https://faker.readthedocs.io/en/master/#unique-values. Have we tried this? Maybe we can remove the loop with this and say if it's unable to generate unique rows using this attribute we probably can't do it

amontanez24 (commented):
I left a question about how we generate unique values

rdt/transformers/pii/anonymizer.py
```diff
@@ -209,3 +214,134 @@ def __repr__(self):
...
+    """
+
+    OUTPUT_SDTYPES = {'value': 'categorical'}
+    NEXT_TRANSFORMER = {
+        'value': LabelEncoder(add_noise=True)
+    }
+
+    def __getstate__(self):
+        """Return a dictionary representation of the instance and warn the user when pickling."""
+        LOGGER.warning((
```
amontanez24, line 246: any reason to use logger over warnings?

tests/unit/transformers/pii/test_anonymizer.py
```diff
@@ -508,3 +512,293 @@ def test___repr__custom_provider(self):
...
+            instance._transform(pd.Series([1, 2, 3]))
+
+        with pytest.raises(Error, match=error_msg_long):
+            instance._transform(pd.Series([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]))
+
+    def test__reverse_transform(self):
+        """Test the ``_reverse_transform`` method.
+
+        Test that the ``_reverse_transform``  method uses the ``instance._label_encoder`` to
+        reverse transform the input data.
```
amontanez24, line 786: update this

tests/unit/transformers/pii/test_anonymizer.py
```diff
@@ -508,3 +512,293 @@ def test___repr__custom_provider(self):
...
+        data = pd.Series(['a', 'b', 'c'])
+
+        # Run
+        instance._fit(data)
+
+        # Assert
+        assert instance._mapping_dict == {'a': 1, 'b': 2, 'c': 3}
+        assert instance._reverse_mapping_dict == {1: 'a', 2: 'b', 3: 'c'}
+
+    def test__fit_multiple_iterations_raises_an_error(self):
```
amontanez24, line 685: should this name be updated?

amontanez24 (commented):
Just a few small comments

amontanez24 (approved):
LGTM!

PR #525, opened 2022-07-09
Performance tests update

### LINKED ISSUE #524: Performance Tests update
The current profiling functions for the transformers contain extra code to support legacy transformers which are no longer present.
Since all the current transformers share an unified `API` we can remove the legacy code and make the required changes in order to support the new workflow.

pvk-developer:
Resolves #524

### DIFF
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -145,16 +145,17 @@ class BaseTransformer:
 
     @staticmethod
     def _set_columns_data(data, columns_data, columns):
-        if columns_data is None:
-            return
+        if columns_data is not None:
+            if isinstance(columns_data, (pd.DataFrame, pd.Series)):
+                columns_data.index = data.index
 
-        if isinstance(columns_data, (pd.DataFrame, pd.Series)):
-            columns_data.index = data.index
+            if len(columns_data.shape) == 1:
+                data[columns[0]] = columns_data
+            else:
+                new_data = pd.DataFrame(columns_data, columns=columns)
+                data = pd.concat([data, new_data.set_index(data.index)], axis=1)
 
-        if len(columns_data.shape) == 1:
-            data[columns[0]] = columns_data
-        else:
-            data[columns] = columns_data
+        return data
 
     def _build_output_columns(self, data):
         self.column_prefix = '#'.join(self.columns)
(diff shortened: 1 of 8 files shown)


rdt/transformers/base.py
```diff
@@ -145,16 +145,17 @@ def _get_columns_data(data, columns):
...
+                data[columns[0]] = columns_data
+            else:
+                new_data = pd.DataFrame(columns_data, columns=columns)
+                data = pd.concat([data, new_data.set_index(data.index)], axis=1)
 
-        if len(columns_data.shape) == 1:
-            data[columns[0]] = columns_data
-        else:
-            data[columns] = columns_data
+        return data
```
pvk-developer, line 158: This change is motivated by the following [logs](https://github.com/sdv-dev/RDT/runs/7263153947?check_suite_focus=true)

rdt/transformers/base.py
```diff
@@ -145,16 +145,17 @@ def _get_columns_data(data, columns):
 
     @staticmethod
     def _set_columns_data(data, columns_data, columns):
```
amontanez24, line 147: minor: can we rename this to `add_columns_to_data` and change the parameter names to `data, columns, column_names`?

amontanez24 (commented):
This is looking good, but I'm not sure I understand why we have to bump all the memory and time values. Is there a good reason for this?

pvk-developer:
> This is looking good, but I'm not sure I understand why we have to bump all the memory and time values. Is there a good reason for this?

Before we were using `pd.Series` but now we are using `pd.DataFrame`, this lead to the memory increase. Also, now we are going over the `base.py` which has some extra steps and that of course brings some extra time. 
I just went and bumped most of the values in order to be 'covered' in the future.

rdt/transformers/base.py
```diff
@@ -144,17 +144,31 @@ def _get_columns_data(data, columns):
         return data[columns].copy()
 
     @staticmethod
-    def _set_columns_data(data, columns_data, columns):
-        if columns_data is None:
-            return
+    def add_columns_to_data(data, columns, column_names):
```
amontanez24, line 147: minor: can we keep it private (add the `_` back)? Sorry I didn't put that in my first comment

amontanez24 (approved):
Just one small comment and then :shipit:

rdt/transformers/base.py
pvk-developer: :shipit:

PR #531, opened 2022-08-10
Add support for multiple `sdtypes` by a transformer.

### LINKED ISSUE #527: Boolean columns should be able to use any of the categorical transformers
### Problem Description

Right now, I can only use the `BinaryEncoder` for boolean columns. Meanwhile, categorical columns have many options such as `LabelEncoder` (with and without noise), `OneHotEncoder` and more. These options are nice because they can improve the synthetic data quality.

### Expected behavior

We can think of boolean data as a special case of categorical data where number of categories = 2. Any categorical transformer should be able to work on boolean data.
 
We should allow any of the categorical transformers to be used on boolean data. The HyperTransformer should not throw an error if I try to apply a categorical transformer to the boolean sdtype.

pvk-developer:
Resolves #527

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -34,6 +34,7 @@ class FrequencyEncoder(BaseTransformer):
     """
 
     INPUT_SDTYPE = 'categorical'
+    SUPPORTED_SDTYPES = ['categorical', 'boolean', 'numerical', 'datetime']
     OUTPUT_SDTYPES = {'value': 'float'}
     DETERMINISTIC_REVERSE = True
     COMPOSITION_IS_IDENTITY = True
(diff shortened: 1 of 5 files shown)


rdt/transformers/categorical.py
```diff
@@ -34,6 +34,7 @@ class FrequencyEncoder(BaseTransformer):
     """
 
     INPUT_SDTYPE = 'categorical'
+    SUPPORTED_SDTYPES = ['categorical', 'boolean', 'numerical', 'datetime']
```
amontanez24, line 37: let's keep it to `categorical` and `boolean`

amontanez24 (approved):
One small comment but after that :shipit:

fealho (approved):
👍

PR #534, opened 2022-08-11
Ensure reversibility `FrequencyEncoder`

### LINKED ISSUE #528: Unpredictable results for `FrequencyEncoder(add_noise=True)`
### Environment Details

* RDT version: 1.1.0
* Python version: 3.9
* Operating System: Colab Notebook

### Error Description

When there is added noise, the observed composition identity of this transformer is `False` even though it is [listed as `True`](https://github.com/sdv-dev/RDT/blob/4487d00f33e8625e529158a034a6a6136065fc12/rdt/transformers/categorical.py#L39). This is causing some problems with [conditional sampling in the SDV](https://github.com/sdv-dev/SDV/issues/910)

There seem to be 2 related issues:
1. The forward transform can noise the values outside of the allowable range for a category, and
2. Some the reverse transformed values are not following the intervals

### Steps to reproduce

To replicate, download and use the `student_placements` dataset.

```python
import numpy as np
import pandas as np
import rdt

data = pd.read_csv('student_placements.csv')

ht = rdt.HyperTransformer()
ht.update_transformers(column_name_to_transformer={
    'gender': rdt.transformers.categorical.FrequencyEncoder(add_noise=True),
    'high_spec': rdt.transformers.categorical.FrequencyEncoder(add_noise=True)
})

np.random.seed(seed=33)

```
Observe that the original data and the reverse transformed data do not have the same values for two of the rows

```python
correct_rows = data['gender'] == reversed['gender']
correct_rows.value_counts()

True     213
False      2
Name: gender, dtype: int64
```

**Forward Transform**

Observe that `M` is supposed to be mapped to the interval (0, 0.6465116279069767). Sometimes the forward transform is mapping outside that range (eg. 0.655336)

```python
data[~correct_rows]['gender']
105    M
187    M
Name: gender, dtype: object

transformed[~correct_rows]['gender.value']
105    0.655336
187    0.606273
Name: gender.value, dtype: float64

ht._transformers_tree['gender']['transformer'].intervals
{'F': (0.6465116279069767, 1.0, 0.8232558139534883, 0.05891472868217054),
 'M': (0, 0.6465116279069767, 0.32325581395348835, 0.10775193798449612)}
```

**Reverse Transform**
Observe that `M` is supposed to be mapped to the interval (0, 0.6465116279069767). But some values inside it -- like 0.606273 -- are reversed transformed to `F`.

```python
transformed[~correct_rows]['gender.value']
105    0.655336
187    0.606273
Name: gender.value, dtype: float64

reversed[~correct_rows]['gender']
105    F
187    F
Name: gender, dtype: object

ht._transformers_tree['gender']['transformer'].intervals
{'F': (0.6465116279069767, 1.0, 0.8232558139534883, 0.05891472868217054),
 'M': (0, 0.6465116279069767, 0.32325581395348835, 0.10775193798449612)}
```

This is likely due to [this line](https://github.com/sdv-dev/RDT/blob/v0.6.3/rdt/transformers/categorical.py#L205) -- we are taking the diff of the value with the _average from that category_ and choosing the min distance. This doesn't make sense when it's noised. We should instead be checking to see if each value is within the correct interval.

fealho:
Resolve #528.

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -127,13 +127,24 @@ class FrequencyEncoder(BaseTransformer):
         self.dtype = data.dtype
         self.intervals, self.means, self.starts = self._get_intervals(data)
 
+    def _clip_noised_transform(self, result, start, end):
+        """Clip transformed values.
+
+        Used to ensure the noise added to transformed values doesn't make it
+        go out of the bounds of a given category.
+
+        The upper bound must be slightly lower than ``end``
+        so it doesn't get treated as the next category.
+        """
+        return result.clip(start, end - 1e-9)
+
     def _transform_by_category(self, data):
         """Transform the data by iterating over the different categories."""
         result = np.empty(shape=(len(data), ), dtype=float)
 
         # loop over categories
         for category, values in self.intervals.items():
-            mean, std = values[2:]
+            start, end, mean, std = values
             if category is np.nan:
                 mask = data.isna()
             else:
@@ -141,6 +152,7 @@ class FrequencyEncoder(BaseTransformer):
 
             if self.add_noise:
                 result[mask] = norm.rvs(mean, std, size=mask.sum())
+                result[mask] = self._clip_noised_transform(result[mask], start, end)
             else:
                 result[mask] = mean
 
@@ -151,10 +163,11 @@ class FrequencyEncoder(BaseTransformer):
         if pd.isna(category):
             category = np.nan
 
-        mean, std = self.intervals[category][2:]
+        start, end, mean, std = self.intervals[category]
 
         if self.add_noise:
-            return norm.rvs(mean, std)
+            result = norm.rvs(mean, std)
+            return self._clip_noised_transform(result, start, end)
 
         return mean
 
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -136,7 +136,7 @@ class FrequencyEncoder(BaseTransformer):
         The upper bound must be slightly lower than ``end``
         so it doesn't get treated as the next category.
         """
-        return result.clip(start, end - 1e-9)
+        return np.clip(result, start, end - 1e-9)
 
     def _transform_by_category(self, data):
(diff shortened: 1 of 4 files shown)


pvk-developer (approved):
LGTM :shipit:

rdt/transformers/utils.py
```diff
@@ -1,10 +1,10 @@
 """Tools to generate strings from regular expressions."""
 
 import re
-import sre_parse
 import string
 
 import numpy as np
+import sre_parse
```
amontanez24: I think this causes builds to fail depending on the machine. I think on sdv we just ended up ignoring it

rdt/transformers/categorical.py
```diff
@@ -195,15 +194,15 @@ def _transform(self, data):
...
-        num_categories = len(self.means)
+        num_categories = len(self.starts)
 
         data = np.broadcast_to(data, (num_categories, num_rows)).T
-        means = np.broadcast_to(self.means, (num_rows, num_categories))
-        diffs = np.abs(data - means)
-        indexes = np.argmin(diffs, axis=1)
+        starts = np.broadcast_to(self.starts.index, (num_rows, num_categories))
+        diffs = (data >= starts)[:, ::-1]
+        indexes = num_categories - np.argmax(diffs, axis=1) - 1
```
amontanez24, line 206: What exactly are these two like doing?

amontanez24 (commented):
I just have a small comment and a question to better understand the operations being done. Besides that this looks good!

rdt/transformers/utils.py
fealho: I'll fix it in my other RDT PR.

rdt/transformers/categorical.py
amontanez24: cool, can we rename diffs to `interval_starts_data_is_greater_than` or something like that? and `indexes` to `interval_indexes`?

amontanez24 (approved):
Thanks for addressing the comment! :shipit:

PR #535, opened 2022-08-11
Fix Datetime formatting and inconsistency after reverse transform

### LINKED ISSUE #515: Inconsistency in date format after reverse transform
### Environment details

If you are already running RDT, please indicate the following details about the environment in
which you are running it:

* RDT version: 1.0.0
* Python version: 3.7
* Operating System: google colab

### Problem description

For date columns, after transforming it and then reversing the transform, if the entry that used to be null is filled, the formatting is off and contains time. It’s probably the average time in consideration to all the other values 24 / 4 = 6. In the scenario where the third row is reversed into NaT, we don’t see this issue. See screen shots below

**Expected format**
In this case, row 3 (the original one with NaT value) is reversed to NaT so we don't see the problem.
Screen Shot 2022-05-16 at 6 13 11 PM

**Inconsistent format**
In this case, row 3 is reversed to an average value so we see inconsistency with the output where time is now part of the format.
Screen Shot 2022-05-16 at 6 12 18 PM

### Reproducing the problem
Link to colab notebook ([link](https://colab.research.google.com/drive/1BZ2wgMFYSwUwPxMQhS7GahWlR5xZJpcs?usp=sharing))

```python3
from rdt import get_demo, HyperTransformer

data = get_demo()

ht = HyperTransformer()
ht.detect_initial_config(data=data)

transformed_data = ht.fit_transform(data)
ht.reverse_transform(transformed_data)
```

`HyperTransformer` config
```
Detecting a new config from the data ... SUCCESS
Setting the new config ... SUCCESS
Config:
{
    "sdtypes": {
        "last_login": "datetime",
        "email_optin": "boolean",
        "credit_card": "categorical",
        "age": "numerical",
        "dollars_spent": "numerical"
    },
    "transformers": {
        "last_login": UnixTimestampEncoder(missing_value_replacement='mean'),
        "email_optin": BinaryEncoder(missing_value_replacement='mode'),
        "credit_card": FrequencyEncoder(),
        "age": FloatFormatter(missing_value_replacement='mean'),
        "dollars_spent": FloatFormatter(missing_value_replacement='mean')
    }
}
```

pvk-developer:
Resolves #515

### DIFF
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -43,6 +43,7 @@ class UnixTimestampEncoder(BaseTransformer):
         self.missing_value_replacement = missing_value_replacement
         self.model_missing_values = model_missing_values
         self.datetime_format = datetime_format
+        self._dtype = None
 
     def is_composition_identity(self):
         """Return whether composition of transform and reverse transform produces the input data.
@@ -117,7 +118,8 @@ class UnixTimestampEncoder(BaseTransformer):
             data (pandas.Series):
                 Data to fit the transformer to.
         """
-        if self.datetime_format is None and data.dtype == 'object':
+        self._dtype = data.dtype
+        if self.datetime_format is None:
             self.datetime_format = _guess_datetime_format_for_array(data.to_numpy())
 
         transformed = self._transform_helper(data)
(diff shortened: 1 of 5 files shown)


rdt/transformers/datetime.py
```diff
@@ -156,7 +158,10 @@ def _reverse_transform(self, data):
             datetime_data = pd.Series(datetime_data)
 
         if self.datetime_format:
-            datetime_data = datetime_data.dt.strftime(self.datetime_format)
+            if self._dtype == 'object':
+                datetime_data = datetime_data.dt.strftime(self.datetime_format)
+            elif 'datetime' in self._dtype.name:
```
amontanez24, line 163: is there a way to check that a dtype is a datetime dtype other than checking the string?

amontanez24 (approved):
One question about making sure we correctly detect the dtypes. Besides that looks good!

fealho (approved):

rdt/transformers/datetime.py
pvk-developer: Yes we can use `is_datetime64_dtype`  from `pandas` api. 👍🏻

PR #536, opened 2022-08-12
Add `computer_representation` parameter

### LINKED ISSUE #521: Update `FloatFormatter` with parameters for the computer representation
### Problem Description

As a user, I want to make sure the min/max values in the reverse transform can be represented by the machine.

### Expected behavior

Add the following parameter to `FloatFormatter`:
- `computer_type`: Default (`'Float'`). Accepts: `'Int8'`, `'Int16'`, `'Int32'`, `'Int64'`, `'UInt8'`, `'UInt16'`, `'UInt32'`, `'UInt64'`, `'Float'`

**Functionality Changes**:
- During `fit`, store the original `dtype` of the column
- During the `transform`, convert everything to `Float` for machine learning purposes.
- During the `reverse_transform`:Cast back to the original dtype. As an extra measure: Clip the values to the min and max machine limits for the given `computer_representation`, and round back to whole numbers if needed. (This is a no-op for `Float`.) Note if: `learn_min_max_bounds=True`, then use the learned values instead.

_Note: The `dtype` may be different than the computer representation. For example, pandas might have read in a column as `Int64` by default but the user might be telling us it's supposed to be `UInt8`. Always defer to the parameter, not the dtype._

**Errors**
During `fit` or `transform`: Throw an error if the data is out of bounds according to the computer representation. _Note: It does not matter what the actual pandas dtype is, only what the computer representation parameter is._
```python
# limits correspond to uint8
transformer = FloatFormatter(computer_representation='UInt8')
transformer.fit(data, column='test')
Error: The minimum value in column 'test' is -5. All values represented by 'UInt8' must be in the range [0, 255].

transformer.transform(data)
Error: The minimum value in column 'test' is -5. All values represented by 'UInt8' must be in the range [0, 255].
```

### Additional context

- Info about pandas dtypes [here](https://pbpython.com/pandas_dtypes.html) and [here](https://pandas.pydata.org/docs/user_guide/basics.html#basics-dtypes)
- A reference for standard min/max values: [here](https://crystal-lang.org/reference/1.4/syntax_and_semantics/literals/integers.html#integers)

See #518 as an example for where this fails today.

fealho:
Resolve #521.

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -117,14 +117,14 @@ class FloatFormatter(BaseTransformer):
                         return decimal
 
         return None
-    
+
     def _raise_out_of_bounds_error(self, values, bound_type, min_bound, max_bound):
         column = values.index[0]
         value = values[column]
         raise ValueError(
             f"The {bound_type} value in column '{column}' is {value}."
             f" All values represented by '{self.computer_representation}'"
-            f" must be in the range [{min_bound}, {max_bound}]."
+            f' must be in the range [{min_bound}, {max_bound}].'
         )
 
     def _validate_values_within_bounds(self, data):
@@ -198,8 +198,10 @@ class FloatFormatter(BaseTransformer):
 
         if self.enforce_min_max_values:
             data = data.clip(self._min_value, self._max_value)
-        else:
-            data = data
+
+        if self.computer_representation != 'Float':
+            min_bound, max_bound = INTEGER_BOUNDS[self.computer_representation]
+            data = data.clip(min_bound, max_bound)
 
         is_integer = np.dtype(self._dtype).kind == 'i'
         if self.learn_rounding_scheme or is_integer:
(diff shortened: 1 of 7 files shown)


rdt/transformers/numerical.py
```diff
@@ -157,6 +204,10 @@ def _reverse_transform(self, data):
         if self.enforce_min_max_values:
             data = data.clip(self._min_value, self._max_value)
 
+        if self.computer_representation != 'Float':
+            min_bound, max_bound = INTEGER_BOUNDS[self.computer_representation]
+            data = data.clip(min_bound, max_bound)
```
amontanez24, line 209: we only need to clip once, so we could just get the strictest bounds from the parameters and clip with that right?

amontanez24 (commented):
This looks mostly ready! Just one small comment

rdt/transformers/numerical.py
fealho: I did it like this because I thought the added complexity of the code didn't justify the minimal speed gains. I'm committing the fix below, so you can compare.

amontanez24 (approved):
LGTM!

PR #544, opened 2022-08-30
Inconsistency in datetime format after reverse transform

pvk-developer:


rdt/transformers/datetime.py
```diff
@@ -117,8 +119,10 @@ def _fit(self, data):
             data (pandas.Series):
                 Data to fit the transformer to.
         """
-        if self.datetime_format is None and data.dtype == 'object':
-            self.datetime_format = _guess_datetime_format_for_array(data.to_numpy())
+        self._dtype = data.dtype
+        if self.datetime_format is None:
+            datetime_array = data.astype(str).to_numpy()
```
pvk-developer, line 124: This way we are able to properly capture the `str` representation of the `dataframe`.
If we have a representation of a `datetime[ns]` as `2022-04-05`, if we convert it directly to `numpy` it will be represented as `2022-04-05 00:00:00000000` which is not the intended output format.

tests/integration/test_transformers.py
```diff
@@ -21,10 +21,12 @@
     },
     'UnixTimestampEncoder': {
         'missing_value_replacement': 'mean',
+        'datetime_format': False,
```
pvk-developer, line 24: Setting `datetime_format` to `False` does not learn the `datetimeformat` and helps us not having ot change all the integration tests.

amontanez24: is there a way to instead figure out what the format string for the datetimes of the integration data would be and pass that in here instead? This is a strange workaround

amontanez24 (changes requested):

amontanez24 (approved):

katxiao (approved):
Nice, lgtm!

PR #547, opened 2022-08-31
Increase pylint score to `10.0`

pvk-developer:


fealho (approved):

fealho:
@pvk-developer The title implies you have to make some change in the setup.py no?

pvk-developer:
> @pvk-developer The title implies you have to make some change in the setup.py no?

Increase the score 👍🏻  just changed that!

PR #548, opened 2022-09-07
Add `enforce_uniqueness` to `AnonymizedFaker` and `RegexGenerator`

### LINKED ISSUE #542: Add parameter to `AnonymizedFaker()` and `RegexGenerator()` to generate only unique values
### Problem Description

To effectively use `AnonymizedFaker` and `RegexGenerator` on data such as primary keys, I need to ensure that the anonymized data created during the `reverse_transform` stage is unique.

### Expected behavior

**Parameters**
Add parameter `enforce_uniqueness` that controls whether the data should be unique.
- (default) `False`: No guarantees that the anonymized data is unique
- `True`: Guarantee that the anonymized data will be unique

**Functionality**
If `enforce_uniqueness=True` then:
- During `reverse_transform`: Ensure that the new, anonymized data is all unique. If it isn't possible to create the requested number of rows, then raise an error

```
Error: The Faker function you specified is not able to generate 1000 unique values. Please use a different Faker function for column ('user_id').

Error: The regex is not able to generate 1000 unique values. Please use a different regex for column ('user_id').
```

### Backwards Compatibility

This change is backwards compatible.

The `enforce_uniqueness` parameter is optional and assumed to be `False` if it's not provided. This is the current functionality.

pvk-developer:
Resolves #542

### DIFF
--- a/rdt/transformers/pii/anonymizer.py
+++ b/rdt/transformers/pii/anonymizer.py
@@ -94,8 +94,10 @@ class AnonymizedFaker(BaseTransformer):
             )
 
     def __init__(self, provider_name=None, function_name=None, function_kwargs=None,
-                 locales=None, missing_value_replacement=None, model_missing_values=False):
+                 locales=None, missing_value_replacement=None,
+                 model_missing_values=False, enforce_uniqueness=False):
         self.data_length = None
+        self.enforce_uniqueness = enforce_uniqueness
         self.provider_name = provider_name if provider_name else 'BaseProvider'
         if self.provider_name != 'BaseProvider' and function_name is None:
             raise Error(
@@ -117,6 +119,9 @@ class AnonymizedFaker(BaseTransformer):
 
     def _function(self):
         """Return a callable ``faker`` function."""
+        if self.enforce_uniqueness:
+            return getattr(self.faker.unique, self.function_name)(**self.function_kwargs)
+
         return getattr(self.faker, self.function_name)(**self.function_kwargs)
 
     def get_output_sdtypes(self):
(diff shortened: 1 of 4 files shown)


tests/unit/transformers/pii/test_anonymizer.py
```diff
@@ -102,9 +107,49 @@ def test__function(self):
         result = AnonymizedFaker._function(instance)
 
         # Assert
+        unique_function.assert_not_called()
         function.assert_called_once_with(type='int')
         assert result == 1
 
+    def test__function_enforce_uniqueness_true(self):
+        """Test that ``_function``.
```
fealho, line 115: Add description, or remove the `that`. Do the same on line 77.

tests/unit/transformers/pii/test_anonymizer.py
```diff
@@ -476,6 +524,35 @@ def test__reverse_transform_models_missing_values(self):
...
+        # Setup
+        instance = AnonymizedFaker('misc', 'boolean', enforce_uniqueness=True)
+        data = pd.Series(['a', 'b', 'c', 'd'])
+        instance.columns = ['a']
+
+        # Run / Assert
+        error_msg = re.escape(
+            'The Faker function you specified is not able to generate 4 unique '
+            'values. Please use a different Faker function for column '
+            "('a')."
```
fealho, line 551: No need for this new line, just put it in the line above.

fealho (approved):
Looks good!

tests/unit/transformers/pii/test_anonymizer.py
pvk-developer: 👍🏻

rdt/transformers/pii/anonymizer.py
```diff
@@ -234,7 +250,7 @@ class PseudoAnonymizedFaker(AnonymizedFaker):
     """
 
     OUTPUT_SDTYPES = {'value': 'categorical'}
-    NEXT_TRANSFORMER = {
+    NEXT_TRANSFORMERS = {
```
amontanez24, line 253: nice!

rdt/transformers/text.py
```diff
@@ -103,14 +109,20 @@ def _reverse_transform(self, data):
...
+                f'The regex is not able to generate {sample_size} unique values. '
+                f"Please use a different regex for column ('{self.get_input_column()}')."
+            )
+
         if sample_size > size:
             warnings.warn(
                 f"The data has {sample_size} rows but the regex for '{self.get_input_column()}' "
                 f'can only create {size} unique values. Some values in '
                 f"'{self.get_input_column()}' may be repeated."
             )
```
amontanez24, line 123: minor but can we combine these if statements? like 
```python
if sample_size > size:
    if self.enforce_uniqueness:
        raise error
    
    warnings.warn(...
```

amontanez24 (approved):
One small comment, but this looks good to me! Let's hold off on merging though until after the release

PR #549, opened 2022-09-09
Extra anonymized columns are created during `reverse_transform_subset`

### LINKED ISSUE #545: Extra anonymized columns are created during `reverse_transform_subset`
### Environment Details

* RDT version: 1.2.0

### Error Description

The `reverse_transform_subset` method is meant to only reverse transform the columns that are inputted. However, it appears to be creating additional columns from `AnonymizedFaker` and `RegexGenerator` transformers.

This is likely happening because these two transformers drop columns on the forward transform.

### Steps to reproduce

```python
from rdt import get_demo

from rdt.transformers.text import RegexGenerator

# create some data
customers = get_demo()
customers['id'] = ['ID_a', 'ID_b', 'ID_c', 'ID_d', 'ID_e']

# create a config
ht = HyperTransformer()
ht.detect_initial_config(customers)

# credit_card and id are pii and text columns
ht.update_sdtypes({
    'credit_card': 'pii',
    'id': 'text'
})

ht.update_transformers({
    'credit_card': AnonymizedFaker(),
    'id': RegexGenerator(regex_format='id_[a-z]')
})

# transform the data
# this will drop the credit card and id columns
ht.fit(customers)
transformed = ht.transform(customers)

# try to reverse transform only the login column
ht.reverse_transform_subset(transformed[['last_login.value']])
```

### Output
The `credit_card` and `id` columns are recreated during `reverse_transform_subset` even though I only asked to reverse transform the `last_login` column.

image

I expect that _only_ the `last_login` column will be created.

pvk-developer:
Resolves #545

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -798,24 +798,30 @@ class HyperTransformer:
             raise NotFittedError(self._NOT_FIT_MESSAGE)
 
         unknown_columns = self._subset(data.columns, self._output_columns, not_in=True)
+        if unknown_columns:
+            raise Error(
+                'There are unexpected column names in the data you are trying to transform. '
+                f'A reverse transform is not defined for {unknown_columns}.'
+            )
+
         if prevent_subset:
             contained = all(column in self._output_columns for column in data.columns)
             is_subset = contained and len(data.columns) < len(self._output_columns)
-            if unknown_columns or is_subset:
+            if is_subset:
                 raise Error(
                     'There are unexpected columns in the data you are trying to transform. '
                     'You must provide a transformed dataset with all the columns from the '
                     'original data.'
                 )
 
-        elif unknown_columns:
-            raise Error(
-                'There are unexpected column names in the data you are trying to transform. '
-                f'A reverse transform is not defined for {unknown_columns}.'
-            )
+            for transformer in reversed(self._transformers_sequence):
+                data = transformer.reverse_transform(data, drop=False)
 
-        for transformer in reversed(self._transformers_sequence):
-            data = transformer.reverse_transform(data, drop=False)
+        else:
+            for transformer in reversed(self._transformers_sequence):
+                output_columns = transformer.get_output_columns()
+                if output_columns and set(output_columns).issubset(data.columns):
+                    data = transformer.reverse_transform(data, drop=False)
 
         reversed_columns = self._subset(self._input_columns, data.columns)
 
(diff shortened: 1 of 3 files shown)


rdt/hyper_transformer.py
```diff
@@ -798,24 +798,30 @@ def _reverse_transform(self, data, prevent_subset):
...
-            )
+            for transformer in reversed(self._transformers_sequence):
+                data = transformer.reverse_transform(data, drop=False)
 
-        for transformer in reversed(self._transformers_sequence):
-            data = transformer.reverse_transform(data, drop=False)
+        else:
+            for transformer in reversed(self._transformers_sequence):
+                output_columns = transformer.get_output_columns()
+                if output_columns and set(output_columns).issubset(data.columns):
```
fealho, line 823: Why is it necessary to check that output_columns is not an empty list? Instead of just:
`if set(output_columns).issubset(data.columns):`

fealho (approved):

rdt/transformers/datetime.py
```diff
@@ -106,7 +106,7 @@ def _reverse_transform_helper(self, data):
         if not isinstance(data, np.ndarray):
             data = data.to_numpy()
 
-        if self.missing_value_replacement is not None:
+        if self.model_missing_values or self.missing_value_replacement is not None:
```
amontanez24, line 109: I think if you rebase this will go away

rdt/hyper_transformer.py
```diff
@@ -798,24 +798,30 @@ def _reverse_transform(self, data, prevent_subset):
...
+                f'A reverse transform is not defined for {unknown_columns}.'
+            )
+
         if prevent_subset:
             contained = all(column in self._output_columns for column in data.columns)
             is_subset = contained and len(data.columns) < len(self._output_columns)
-            if unknown_columns or is_subset:
+            if is_subset:
                 raise Error(
                     'There are unexpected columns in the data you are trying to transform. '
```
amontanez24, line 812: with the change you are making, idk if this error message makes sense anymore. There will no longer be unexpected columns here right? Just missing ones?

rdt/hyper_transformer.py
pvk-developer: Because some transformers, generators, return an empty list, and that conditions ends up being `True` if so:
```python
In [1]: set(list()).issubset(['a'])
Out[1]: True
```

amontanez24 (approved):
Thanks for addressing :shipit:

PR #554, opened 2022-09-22
Add `transformer_name`, `transformer_parameters` parameters to `update_transformers_by_sdtype`

### LINKED ISSUE #540: Update API for `update_transformers_by_sdtype` to be more explicit about instances vs. copies
### Problem Description

As a user, it is confusing to pass in instance to the method `update_transformers_by_sdtype` when in reality, the RDT is actually creating _copies_ of the instance for actual assignment.

Eg. 
```python
hyper_transformer.update_transformers_by_sdtype(
  sdtype='numerical',
  transformer=FloatFormatter(model_missing_values=True) # this exact instance is never used
)
```

### Expected behavior

We can more explicitly communicate intent by allowing users to input a description of the transformer rather than an instance of the transformer.

**New Parameters:**
- (required) `transformer_name`: A string with the class name of the transformers
- `transformer_parameters`: An optional dictionary containing the keyword args for the transformer

Based on this, we can create multiple instances of the transformer.

```python
hyper_transformer.update_transformers_by_sdtype(
  sdtype='numerical',
  transformer_name='FloatFormatter',
  transformer_parameters={'model_missing_values': True}
)
```

### Backwards Compatibility

We should still allow the older usage without crashing. However there should be a deprecation warning whenever it's used.

```python
hyper_transformer.update_transformers_by_sdtype(
  sdtype='numerical',
  transformer=FloatFormatter(model_missing_values=True)
)

Deprecation Warning: The 'transformer' parameter will no longer be supported in future versions of the RDT.
Please use the 'transformer_name' and 'transformer_parameters' parameters instead.
```

### Error Cases

**All parameters are used**: A user may pass in `transformer` as well as the new `transformer_name` and `transformer_parameters`. In this case, we can throw a modified deprecation warning and prioritize the info in the new parameters instead.
```python
hyper_transformer.update_transformers_by_sdtype(
  sdtype='categorical',
  transformer=FrequencyEncoder(),
  transformer_name='LabelEncoder'
)

Deprecation Warning: The 'transformer' parameter will no longer be supported in future versions of the RDT.
Using the 'transformer_name' parameter instead.
```

**No transformer or `transformer_name`** is given.
```python
hyper_transformer.update_transformers_by_sdtype(
   sdtype='numerical',
   transformer_parameters={'model_missing_values': True}
)
Error: Missing required parameter 'transformer_name'.
```

**Invalid transformer name**: If the transformer name is invalid for the given sdtype.
```python
hyper_transformer.update_transformers_by_sdtype(
   sdtype='numerical',
   transformer_name='FFloatFormater'
)
Error: Invalid transformer name 'FFloatFormater' for the 'numerical' sdtype.

# Same if it doesn't match up to the sdtype 
hyper_transformer.update_transformers_by_sdtype(
   sdtype='numerical',
   transformer_name='LabelEncoder'
)
Error: Invalid transformer name 'LabelEncoder' for the 'numerical' sdtype.
```

**Invalid keyword args**: If the the arguments are not valid for the given transformer name
```python
hyper_transformer.update_transformers_by_sdtype(
   sdtype='numerical',
   transformer_name='FloatFormatter',
   transformer_parameters={ 'order': ['a', 'b', 'c'] }
)
Error: Invalid parameters ('order') for the 'FloatFormatter'.
```

fealho:
Resolve #540.

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -11,7 +11,8 @@ import yaml
 
 from rdt.errors import Error, NotFittedError
 from rdt.transformers import (
-    BaseTransformer, get_default_transformer, get_transformer_instance, get_transformers_by_type)
+    BaseTransformer, get_class_by_transformer_name, get_default_transformer,
+    get_transformer_instance, get_transformers_by_type)
 
 
 class Config(dict):
@@ -96,9 +97,6 @@ class HyperTransformer:
         'The HyperTransformer is not ready to use. Please fit your data first using '
         "'fit' or 'fit_transform'."
     )
-    _TRANSFORMERS_TO_CLASS = {
-        class_.__name__: class_ for class_ in BaseTransformer.get_subclasses()
-    }
 
     @staticmethod
     def _user_message(text, prefix=None):
@@ -314,7 +312,7 @@ class HyperTransformer:
 
         if transformer_name is None:
             if transformer is None:
-                raise ValueError("Missing required parameter 'transformer_name'.")
+                raise Error("Missing required parameter 'transformer_name'.")
 
             if not isinstance(transformer, BaseTransformer):
                 raise Error('Invalid transformer. Please input an rdt transformer object.')
@@ -323,22 +321,22 @@ class HyperTransformer:
                 raise Error("The transformer you've assigned is incompatible with the sdtype.")
 
         else:
-            if transformer_name not in self._TRANSFORMERS_TO_CLASS or sdtype not in \
-                    self._TRANSFORMERS_TO_CLASS[transformer_name].get_supported_sdtypes():
-                raise ValueError(
+            if transformer_name not in get_class_by_transformer_name() or sdtype not in \
+                    get_class_by_transformer_name()[transformer_name].get_supported_sdtypes():
+                raise Error(
                     f"Invalid transformer name '{transformer_name}' for the '{sdtype}' sdtype.")
 
             if transformer_parameters is not None:
-                transformer = self._TRANSFORMERS_TO_CLASS[transformer_name]
+                transformer = get_class_by_transformer_name()[transformer_name]
                 valid = inspect.signature(transformer).parameters
                 invalid_parameters = {arg for arg in transformer_parameters if arg not in valid}
                 if invalid_parameters:
-                    raise ValueError(
+                    raise Error(
                         f'Invalid parameters {tuple(sorted(invalid_parameters))} '
                         f"for the '{transformer_name}'."
                     )
 
-    def _warnings_update_transformers_by_sdtype(self, transformer, transformer_name):
+    def _warn_update_transformers_by_sdtype(self, transformer, transformer_name):
         if self._fitted:
             warnings.warn(self._REFIT_MESSAGE)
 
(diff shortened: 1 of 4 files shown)


rdt/transformers/categorical.py
```diff
@@ -207,7 +207,7 @@ def _reverse_transform_by_matrix(self, data):
         is_data_greater_than_starts = (data >= starts)[:, ::-1]
         interval_indexes = num_categories - np.argmax(is_data_greater_than_starts, axis=1) - 1
 
-        get_category_from_index = list(self.starts['category']).__getitem__
+        get_category_from_index = list(self.starts['category']).__getitem__  # pylint:disable=E1136
```
fealho, line 210: I have no idea why these started to fail lint. I'm assuming some library just got released which makes them fail?

rdt/transformers/categorical.py
```diff
@@ -329,7 +329,7 @@ def _fit(self, data):
         data = self._prepare_data(data)
 
         null = pd.isna(data)
-        self._uniques = list(pd.unique(data[~null]))
+        self._uniques = list(pd.unique(data[~null]))  # pylint: disable=E1130
```
fealho, line 332: Same

rdt/transformers/categorical.py
```diff
@@ -465,7 +465,7 @@ def _order_categories(self, unique_data):
 
         if self.order_by is not None:
             nans = pd.isna(unique_data)
-            unique_data = np.sort(unique_data[~nans])
+            unique_data = np.sort(unique_data[~nans])  # pylint: disable=E1130
```
fealho, line 468: Same

rdt/transformers/categorical.py
pvk-developer: This is due to recent updates on `pylint` my older version was `pylint==2.12.2` but after updating it started failing. However instead of ignore those I would propose to fix them which is a minor change:

```python
null = pd.isna(data).to_numpy()
```

rdt/transformers/categorical.py
pvk-developer, line 210: This one can be fixed by using
```python
category_starts = self.starts.category.to_list()
get_category_from_index = category_starts.__getitem__
```

pvk-developer (changes requested):

rdt/hyper_transformer.py
```diff
@@ -95,6 +96,9 @@ class HyperTransformer:
         'The HyperTransformer is not ready to use. Please fit your data first using '
         "'fit' or 'fit_transform'."
     )
+    _TRANSFORMERS_TO_CLASS = {
+        class_.__name__: class_ for class_ in BaseTransformer.get_subclasses()
+    }
```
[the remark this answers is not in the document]
amontanez24: I think there is already a method that is cached that we can use in place of `BaseTransformer.get_subclasses()`

rdt/hyper_transformer.py
```diff
@@ -300,7 +304,61 @@ def set_config(self, config):
...
+                transformer = self._TRANSFORMERS_TO_CLASS[transformer_name]
+                valid = inspect.signature(transformer).parameters
+                invalid_parameters = {arg for arg in transformer_parameters if arg not in valid}
+                if invalid_parameters:
+                    raise ValueError(
+                        f'Invalid parameters {tuple(sorted(invalid_parameters))} '
+                        f"for the '{transformer_name}'."
+                    )
+
+    def _warnings_update_transformers_by_sdtype(self, transformer, transformer_name):
```
amontanez24, line 341: minor: I think we should try to make all methods verbs. so `_check_warnings...` or `_warn..`

rdt/hyper_transformer.py
```diff
@@ -300,7 +304,61 @@ def set_config(self, config):
...
+        if self._fitted:
+            warnings.warn(self._REFIT_MESSAGE)
+
+        if transformer_name is not None:
+            if transformer is not None:
```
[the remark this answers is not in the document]
fealho: But then the `else` wouldn't work, I want the `else` to be with regards to the `transformer_name is not None`, not the whole thing.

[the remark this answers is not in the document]
pvk-developer: 🤔 Yes i see your point I got miss-leaded (I read it right on the line 316 but here I went with my vision I guess).

pvk-developer (approved):

rdt/transformers/__init__.py
```diff
@@ -94,6 +94,21 @@ def get_transformer_name(transformer):
 }
 
 
+@lru_cache()
```
fealho, line 97: @pvk-developer @amontanez24 wrote this method to address your comment regarding the `TRANSFORMERS_TO_CLASS` dict.

It feels weird to use cache with only 1 key, but I didn't want to import the dictionary itself 🤷

amontanez24 (approved):
Thanks for making the changes! :shipit:

PR #562, opened 2022-10-06
Fix when supported sdtype was set using `set_config`.

### LINKED ISSUE #560: Setting config on `HyperTransformer` does not read `supported_sdtypes`
### Error Description

Currently, when setting a config to the `HyperTransformer` if the `sdtype` is not the same as the result of  `get_input_sdtype`, it fails. We should use `get_supported_sdtypes` instead.

### Steps to reproduce
```python
from rdt.transformers import LabelEncoder

config = {
    'transformers': {
        'boolean_col': LabelEncoder(add_noise=True),
    },
    'sdtypes': {
        'boolean_col': 'boolean'
    }
}

ht = HyperTransformer()
ht.set_config(config)
Error: Some transformers you've assigned are not compatible with the sdtypes. Please change the following columns: ['boolean_col']
```

pvk-developer:
Resolves #560

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -39,37 +39,6 @@ class HyperTransformer:
 
     The ``HyperTransformer`` class contains a collection of ``transformers`` that can be
     used to transform and reverse transform one or more columns at once.
-
-    Example:
-        Create a simple ``HyperTransformer`` instance that will decide which transformers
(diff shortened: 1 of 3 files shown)


amontanez24 (approved):
LGTM :shipit:

fealho (approved):

PR #563, opened 2022-10-07
Update `update_transformers` validation

fealho:
Update `update_transformers` so an error is raised when the sdtype is not supported by the transformer, instead of raising an error.

### DIFF
--- a/rdt/errors.py
+++ b/rdt/errors.py
@@ -9,5 +9,5 @@ class Error(Exception):
     """Error to raise when ``HyperTransformer`` produces a controlled error message."""
 
 
-class InvalidSdtypeForTransformerError(Exception):
-    """Error to raise when the sdtype is not supported by the transformer."""
+class SynthesizerInputError(Exception):
+    """Error to raise when invalid parameters are used to set up the synthesizer."""
--- a/rdt/errors.py
+++ b/rdt/errors.py
@@ -9,5 +9,5 @@ class Error(Exception):
     """Error to raise when ``HyperTransformer`` produces a controlled error message."""
 
 
-class SynthesizerInputError(Exception):
-    """Error to raise when invalid parameters are used to set up the synthesizer."""
+class TransformerInputError(Exception):
+    """Error to raise when ``HyperTransformer`` receives incorrect input."""
--- a/rdt/errors.py
+++ b/rdt/errors.py
@@ -7,3 +7,7 @@ class NotFittedError(Exception):
 
 class Error(Exception):
     """Error to raise when ``HyperTransformer`` produces a controlled error message."""
+
+
+class InvalidSdtypeForTransformerError(Exception):
+    """Error to raise when the sdtype is not supported by the transformer."""
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -9,7 +9,7 @@ from copy import deepcopy
 import pandas as pd
 import yaml
 
-from rdt.errors import Error, InvalidSdtypeForTransformerError, NotFittedError
+from rdt.errors import Error, SynthesizerInputError, NotFittedError
 from rdt.transformers import (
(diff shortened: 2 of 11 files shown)


rdt/hyper_transformer.py
```diff
@@ -442,21 +442,17 @@ def update_transformers(self, column_name_to_transformer):
...
 
-        incompatible_sdtypes = []
         for column_name, transformer in column_name_to_transformer.items():
             if transformer is not None:
                 current_sdtype = self.field_sdtypes.get(column_name)
                 if current_sdtype and current_sdtype not in transformer.get_supported_sdtypes():
-                    incompatible_sdtypes.append(column_name)
+                    raise InvalidSdtypeForTransformerError(
+                        f"Column '{column_name}' is a {current_sdtype} column, which is "
+                        f"incompatible with the '{type(transformer).__name__}' transformer."
```
pvk-developer, line 451: 🤔 Should we add a `get_name` method to the `BaseTransformer` and use that whenever we want to get the `__name__` ?

tests/integration/test_transformers.py
```diff
@@ -255,7 +256,11 @@ def _test_transformer_with_hypertransformer(transformer_class, input_data, steps
         }
 
     hypertransformer.detect_initial_config(input_data)
-    hypertransformer.update_transformers(field_transformers)
+    try:
+        hypertransformer.update_transformers(field_transformers)
+    except InvalidSdtypeForTransformerError:
+        pass
```
amontanez24, line 262: I think instead we could use the `set_config` method and avoid having this error get raised altogether

rdt/errors.py
```diff
@@ -7,3 +7,7 @@ class NotFittedError(Exception):
 
 class Error(Exception):
     """Error to raise when ``HyperTransformer`` produces a controlled error message."""
+
+
+class InvalidSdtypeForTransformerError(Exception):
```
amontanez24, line 12: I think we should change this error name to match the ones we created in the doc. You won't need to catch this error anymore in the integration tests if you do the other change

amontanez24 (changes requested):

tests/integration/test_transformers.py
fealho: It will still run into an error, since we have a version of it validating set_config as well:

https://github.com/sdv-dev/RDT/blob/1c2c8211fd5697e4945bbdd28f2cccb353735455/rdt/hyper_transformer.py#L240-L244

rdt/errors.py
```diff
@@ -9,5 +9,5 @@ class Error(Exception):
     """Error to raise when ``HyperTransformer`` produces a controlled error message."""
 
 
-class InvalidSdtypeForTransformerError(Exception):
-    """Error to raise when the sdtype is not supported by the transformer."""
+class SynthesizerInputError(Exception):
```
amontanez24, line 12: This should be `TransformerInputError`. The synthesizer prefixes will only be for `SDV`

tests/integration/test_transformers.py
amontanez24: When we set the config can't we force the sdtype to match?

amontanez24 (changes requested):

tests/integration/test_transformers.py
```diff
@@ -255,12 +254,16 @@ def _test_transformer_with_hypertransformer(transformer_class, input_data, steps
...
-        pass
+    sdtypes = {}
+    for field, transformer in field_transformers.items():
+        sdtypes[field] = transformer.get_supported_sdtypes()[0]
 
+    config = {
+        'sdtypes': sdtypes,
+        'transformers': field_transformers
+    }
+    hypertransformer.detect_initial_config(input_data)
```
amontanez24, line 265: you can just delete this line now

amontanez24 (approved):
One last comment, but then it should be good to go!

PR #565, opened 2022-10-11
Add utility functions

### LINKED ISSUE #561: Add utility functions to `AnonymizedFaker`
### Problem Description

In order to work directly with an `sdtype` (eg: 'job' as sdtype), it would be usefull to have the following two utility functions in `AnonymizedFaker`:
* `is_faker_function`: A function that returns `True` or `False` whether or not the provided `sdtype` is considered a `faker` function.
* `get_provider_name`: A function that returns the `provider_name` for the given `sdtype`.

With the following functions we would be able to support wider range of `sdtypes`.

### Expected behavior

```python
from rdt.transformers.pii.utilis import get_provider_name, is_faker_function

is_faker_function('job')
True

get_provider_name('job')
job

get_provider_name('lexify'):
BaseProvider
```

### Additional context

By adding this two features we will be able to remove `faker` dependency from https://github.com/sdv-dev/SDV

pvk-developer:
Resolves #561

### DIFF
--- a/rdt/transformers/pii/utils.py
+++ b/rdt/transformers/pii/utils.py
@@ -1,0 +1,43 @@
+"""Utility functions for anonymization."""
+
+import inspect
+
+from faker import Faker
+
+
+def is_faker_function(function_name):
+    """Return whether or not the function name is a valid Faker function.
+
+    Args:
+        function_name (str):
+            String representing a ``faker`` function.
+
+    Returns:
+        True if the ``function_name`` is know to ``Faker``, otherwise False.
+    """
+    try:
+        getattr(Faker(), function_name)
+    except AttributeError:
+        return False
+
+    return True
+
+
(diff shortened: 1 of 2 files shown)


amontanez24 (approved):
I think you mentioned something about potentially using these functions within the transformers in `RDT`. I'm not sure if that was something you still wanted to do. Otherwise this looks good!

pvk-developer:
> I think you mentioned something about potentially using these functions within the transformers in `RDT`. I'm not sure if that was something you still wanted to do. Otherwise this looks good!

After I reviewed the current implementation, I don't think we would like to change anything without previous discussions. I think this two functions can stay and if in the future we decide to go with a less restrictive `AnonymizedFaker` we can add those changes.

katxiao (approved):
lgtm

PR #566, opened 2022-10-12
Update `next_transformer` logic

### LINKED ISSUE #557: Change the NEXT_TRANSFORMERS logic
### Problem Description

_As a user, it would be helpful if we could debug transformer chaining by traversing a current transformer to see the transformers used of the outputs._

In order to get this functionality, a few changes should be made.
1. The `NEXT_TRANSFORMERS` attribute should be made an instance attribute instead of class attribute.
2. Any transformer that uses the `NEXT_TRANSFORMERS` attribute should point to an actual transformer instance.
3. If not already present, we should add a `_get_next_transformers` method that returns the actual dictionary, NOT a copy.
4. We should change the logic in these lines to just use the transformer defined in `NEXT_TRANSFORMERS`
https://github.com/sdv-dev/RDT/blob/52ed8a08f52e8707aae06d5ec1e863f70c7e2bc4/rdt/hyper_transformer.py#L668-L679

### Expected behavior

```python
phone_number_transformer._get_next_transformers()
{
    'country': LabelEncoder(),
    'region': LabelEncoder()
}

float_formatter._get_next_transformers()
{
    None: None,
    'is_null': None
}
```

### Additional context

- For now the output columns all have names (so `None` is not a key like in the example above). That will change in #533, so this issue just needs to return a dict mapping the output name to the transformer instance.

fealho:
Resolve #557.


tests/unit/test_hyper_transformer.py
```diff
@@ -565,72 +433,6 @@ def test__fit_field_transformer(self, get_transformer_instance_mock):
             'a.out1': {'transformer': transformer2, 'outputs': ['a.out1.value']}
         }
 
-    @patch('rdt.hyper_transformer.get_transformer_instance')
-    def test__fit_field_transformer_multi_column_field_not_ready(
```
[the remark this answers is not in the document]
fealho: `get_next_transformers` would need to support multicolumn fields, so basically I'd have to make the following change:
```
(delete) transformer1.get_next_transformers.return_value = None
(add) transformer1.get_next_transformers.return_value = {('a.out1', 'b.out1'): transformer2}
```
I don't think the user can do this anymore, so this test doesn't quite make sense anymore.

rdt/transformers/base.py
```diff
@@ -20,11 +20,11 @@ class BaseTransformer:
     DETERMINISTIC_REVERSE = None
     COMPOSITION_IS_IDENTITY = None
     IS_GENERATOR = None
-    NEXT_TRANSFORMERS = None
 
     columns = None
     column_prefix = None
     output_columns = None
+    _next_transformers = None
```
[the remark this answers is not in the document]
fealho: We can, but then we don't have the functionality you mentioned in the issue:
```python3
float_formatter._get_next_transformers()
{
    None: None,
    'is_null': None
}
```

pvk-developer (changes requested):
With the comment I left, if we agree on performing that change, I think most of the changes that happen in the `transformers` may need to be undone.

rdt/transformers/base.py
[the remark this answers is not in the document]
fealho: That's a good idea. My question then is whether we want to store the next transformers or not? Because if we use the method above, there's no point in creating the ``self._next_transformers`` attribute, we can get the next transformers by calling the function.

rdt/transformers/pii/anonymizer.py
```diff
@@ -223,6 +218,9 @@ def __init__(self, provider_name=None, function_name=None, function_kwargs=None,
         )
         self._mapping_dict = {}
         self._reverse_mapping_dict = {}
+        self.output_properties = {
+            'value': {'sdtype': 'categorical', 'next_transformer': LabelEncoder(add_noise=True)}
```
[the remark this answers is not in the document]
fealho: This will be changed in the `.value` PR.

rdt/transformers/base.py
```diff
@@ -73,13 +74,16 @@ def get_supported_sdtypes(cls):
...
-        if not dictionary:
-            return {}
-
+    def _get_output_to_property(self, property_):
         output = {}
-        for output_columns, output_sdtype in dictionary.items():
-            output[f'{self.column_prefix}.{output_columns}'] = output_sdtype
+        for output_column, properties in self.output_properties.items():
+            # if 'sdtype' is not in the dict, ignore the column
+            if property_ not in properties:
```
fealho, line 81: **Note**: this is required for the special case where transformers don't pass an 'sdtype'. In such cases we don't want to add the output_column at all (instead of adding the `column_prefix: None` pair to the dict) because `_build_output_columns` adds a # to the end of generated column which already exit in the data.

rdt/transformers/base.py
```diff
@@ -74,16 +74,16 @@ def get_supported_sdtypes(cls):
         """
         return cls.SUPPORTED_SDTYPES or [cls.INPUT_SDTYPE]
 
-    def _add_prefix(self, sdtype_or_transformer):
+    def _get_output_to_property(self, property_):
```
amontanez24, line 77: is the underscore because property is a shadowed name?

rdt/transformers/base.py
amontanez24: what happens if `property_` is set to `'next_transformer'`? Do we still want to skip if the dict doesn't contain it?

fealho: That never happens (and should never happen). If you mess up and forget to add a `next_transformer` to `output_properties`, skipping it is the best we can do.

rdt/transformers/base.py
fealho: yep

PR #569, opened 2022-10-22
Remove transformers tree

fealho:


rdt/hyper_transformer.py
```diff
@@ -653,25 +548,25 @@ def _fit_field_transformer(self, data, field, transformer):
...
                 next_transformer = next_transformers[output_field]
-                if next_transformer and self._field_in_data(output_field, data):
-                    self._fit_field_transformer(data, output_field, next_transformer)
+
+                # If the column is part of a multi-column field, and at least one column
+                # isn't present in the data, then it should not fit the next transformer
+                if self._field_in_data(output_field, data):
+                    data = self._fit_field_transformer(data, output_field, next_transformer)
 
         return data
```
fealho, line 571: @pvk-developer @amontanez24 This method changes the data inplace, so there's no need to return the data. I ran into a problem with the mocks when trying to make this change though, so I ended up undoing it instead, since I didn't care enough to bother figuring out what the problem is.

If either of you thinks this would indeed be an improvement, let me know and I'll fix it, otherwise I'll just leave it as is.

rdt/hyper_transformer.py
```diff
@@ -831,7 +719,7 @@ def create_anonymized_columns(self, num_rows, column_names):
 
         transformers = []
         for column_name in column_names:
-            transformer = self._transformers_tree.get(column_name, {}).get('transformer')
+            transformer = self.field_transformers.get(column_name, {})
```
amontanez24, line 722: I don't think we should return a dictionary as a backup here. The returned value should be a transformer, so if it doesn't exist shouldn't we let it be `None`?

rdt/hyper_transformer.py
```diff
@@ -653,25 +548,25 @@ def _fit_field_transformer(self, data, field, transformer):
         """
         if transformer is None:
             self._add_field_to_set(field, self._fitted_fields)
-            self._transformers_tree[field]['transformer'] = None
-            self._transformers_tree[field]['outputs'] = [field]
+            if field not in self._output_columns:
+                self._output_columns.append(field)
```
amontanez24, line 552: why do we need this?

amontanez24 (changes requested):

tests/integration/test_hyper_transformer.py
```diff
@@ -199,19 +199,6 @@ def test_hypertransformer_default_inputs():
...
-    assert ht._transformers_tree['float']['outputs'] == ['float.value']
-    assert isinstance(ht._transformers_tree['categorical']['transformer'], FrequencyEncoder)
-    assert ht._transformers_tree['categorical']['outputs'] == ['categorical.value']
-    assert isinstance(ht._transformers_tree['bool']['transformer'], BinaryEncoder)
-    assert ht._transformers_tree['bool']['outputs'] == ['bool.value']
-    assert isinstance(ht._transformers_tree['datetime']['transformer'], UnixTimestampEncoder)
-    assert ht._transformers_tree['datetime']['outputs'] == ['datetime.value']
-    assert isinstance(ht._transformers_tree['names']['transformer'], FrequencyEncoder)
-    assert ht._transformers_tree['names']['outputs'] == ['names.value']
-
```
pvk-developer, line 214: Maybe just leave this as `ht.field_transformers`

tests/integration/test_hyper_transformer.py
```diff
@@ -199,6 +200,19 @@ def test_hypertransformer_default_inputs():
             actual = reverse_transformed.iloc[row, column]
             assert pd.isna(actual) or expected == actual
 
+    assert isinstance(ht.field_transformers['integer'], FloatFormatter)
+    # assert ht.field_transformers['integer'].get_output_columns() == ['integer.value']  noqa
```
fealho, line 204: @pvk-developer  I can't test the code being commented out yet, because it needs code from the other PR's I'm writing. Leaving it here so I can uncomment it after rebasing in the other PR.

amontanez24 (approved):
LGTM!

rdt/hyper_transformer.py
amontanez24: I realized I misread this code, so it would have been fine to leave this in although I don't think it will ever see the same output column more than once

PR #572, opened 2022-10-31
Remove `.value`

### LINKED ISSUE #533: Remove '.value' suffix
### Problem Description

As a user, it is extra work to figure out which output column directly maps to the original since the name is modified. In cases where the transformation creates a new column or completely changes the input column, this makes sense, but in cases like the `FloatFormatter`, it would be nice if the formatted column kept the same name.

### Expected behavior

- Get rid of the `'.value'` suffix on output column names

### Additional context

- we can probably modify the `OUTPUT_SDTYPES` attribute everywhere to accept None or an empty string and then modify the following lines to not add a `.`:
https://github.com/sdv-dev/RDT/blob/2c9e53627bd9c626ee937fb770624b8432a41e9e/rdt/transformers/base.py#L65-L73
- There are a few methods that depend on the `.value` suffix because they use the `_transformer_tree`. After #557 , we can remove all of these methods and the `_transformer_tree` completely.

fealho:
Resolve #533.


rdt/transformers/base.py
```diff
@@ -210,10 +213,11 @@ def _build_output_columns(self, data):
         self.output_columns = self.get_output_columns()
 
         # make sure none of the generated `output_columns` exists in the data
-        data_columns = set(data.columns)
-        while data_columns & set(self.output_columns):
+        output_columns = set(self.output_columns) - set(self.columns)
+        while set(output_columns) & set(data.columns):
             self.column_prefix += '#'
             self.output_columns = self.get_output_columns()
+            output_columns = set(self.output_columns) - set(self.columns)
```
amontanez24, line 218: should this be `self.output_columns`?

rdt/transformers/base.py
```diff
@@ -193,15 +194,17 @@ def _add_columns_to_data(data, columns, column_names):
...
+        if isinstance(transformed_data, (pd.Series, np.ndarray)):
+            transformed_data = pd.DataFrame(transformed_data, columns=transformed_names)
 
-            if len(columns.shape) == 1:
-                data[column_names[0]] = columns
-            else:
-                new_data = pd.DataFrame(columns, columns=column_names)
-                data = pd.concat([data, new_data.set_index(data.index)], axis=1)
+        if transformed_names:
+            transformed_data.index = data.index
```
amontanez24: should we copy `transformed_data` since we are modifying its index or does it not matter?

rdt/transformers/base.py
```diff
@@ -179,7 +180,7 @@ def _get_columns_data(data, columns):
         return data[columns].copy()
 
     @staticmethod
-    def _add_columns_to_data(data, columns, column_names):
+    def _update_data(data, transformed_data, transformed_names):
         """Add new columns to a ``pandas.DataFrame``.
```
amontanez24, line 184: can we update the docstring too?

amontanez24 (commented):
I thought we were going to raise an error in the `HyperTransformer` if we ever saw the same output column twice

rdt/transformers/base.py
fealho: no, this variable control the while loop

rdt/transformers/base.py
fealho: shouldn't matter. I simplified this part of the code either way

tests/unit/test_hyper_transformer.py
```diff
@@ -1022,6 +1005,34 @@ def test_fit(self):
         ht._validate_detect_config_called.assert_called_once()
         ht._unfit.assert_called_once()
 
+    def test_fit_crashes(self):
```
fealho, line 1008: This test won't pass until the instances PR gets merged (which is why the current build fails).

pvk-developer (changes requested):
Just left a minor things to be changed, but after that this should be good to go!

tests/unit/test_hyper_transformer.py
```diff
@@ -1005,32 +1005,72 @@ def test_fit(self):
...
+            "The output columns {'col.is_null'} generated by the DummyTransformer2 "
+            'transformer already exist in the data (or they have already been generated '
+            "by some other transformer). Appending a '#' to the column name to distinguish "
+            'between them.'
+        )
+        with pytest.warns(UserWarning, match=warn_msg):
+            ht.fit(data)
+
+    def test_fit_warns_columns_in_data(self):
+        """Test it warns when transformers generat columns which already exist in the data."""
```
amontanez24, line 1059: typo: generat

amontanez24 (approved):
Some minor comments but should be good after that!

rdt/transformers/base.py
```diff
@@ -212,10 +213,18 @@ def _build_output_columns(self, data):
         # make sure none of the generated `output_columns` exists in the data,
         # except when a column generates another with the same name
         output_columns = set(self.output_columns) - set(self.columns)
-        while set(output_columns) & set(data.columns):
+        repeated_columns = set(output_columns) & set(data.columns)
+        while repeated_columns:
```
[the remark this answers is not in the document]
fealho: @amontanez24 It seemed more annoying, which is why I didn't bother. It should also never really happen (I mean, you have to have the columns `colum_name.is_null` AND `column_name#.is_null` in your data, which seems a little bit of a stretch), as you mentioned.

Also, it's not totally unreasonable to warn every time, since the "error" is occurring because of different columns, so it makes some sense to first tell you that the `colum_name.is_null` is suspicious, and if you also happen to have `column_name#.is_null` in your data, it should raise a separate warning telling you that this other, independent column is also suspicious.

pvk-developer (approved):

PR #577, opened 2022-11-02
Update `missing_value_replacement` default

### LINKED ISSUE #576: Impute missing values by default
### Problem Description
Right now, any transformer that can impute missing values does nothing by default. This causes issues when reverse transforming the data, as missing values can be modeled or randomized separately.

### Expected behavior

1. By default, the transformers should impute missing values: `BinaryEncoder` should use `'mode'`. All other transformers should use `'mean'`.
2. If a user explicitly passes in `missing_value_replacement=None`, then throw a warning:

For other `BinaryEncoder`:
```
Warning: Setting the 'missing_value_replacement' to 'None' is no longer supported. Imputing with the 'mode' instead.  
```

For other transformers:
```
Warning: Setting the 'missing_value_replacement' to 'None' is no longer supported. Imputing with the 'mean' instead.  
```

fealho:
Resolve #576.

### DIFF
--- a/rdt/performance/performance.py
+++ b/rdt/performance/performance.py
@@ -13,24 +13,13 @@ TRANSFORMER_ARGS = {
         'missing_value_replacement': -1,
         'model_missing_values': True
     },
-    'DatetimeTransformer': {
-        'missing_value_replacement': 'mean',
+    'UnixTimestampEncoder': {
         'model_missing_values': True
     },
-    'DatetimeRoundedTransformer': {
-        'missing_value_replacement': 'mean',
+    'OptimizedTimestampEncoder': {
         'model_missing_values': True
     },
     'FloatFormatter': {
-        'missing_value_replacement': 'mean',
-        'model_missing_values': True
-    },
-    'NumericalRoundedBoundedTransformer': {
-        'missing_value_replacement': 'mean',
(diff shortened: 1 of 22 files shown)


rdt/performance/performance.py
```diff
@@ -13,24 +13,13 @@
         'missing_value_replacement': -1,
         'model_missing_values': True
     },
-    'DatetimeTransformer': {
-        'missing_value_replacement': 'mean',
+    'UnixTimestampEncoder': {
```
fealho, line 16: Unrelated change: I think we just forgot to remove these when we renamed the transformers.

pvk-developer (approved):
LGTM!

tests/unit/transformers/test_base.py
```diff
@@ -362,7 +290,7 @@ def test_is_generator(self):
             ``IS_GENERATOR`` is set to ``True``.
 
         Output:
-            - the boolean value stored in ``COMPOSITION_IS_IDENTITY``.
+            - the boolean value stored in ``IS_GENERATOR``.
```
amontanez24, line 293: good catch

amontanez24 (approved):
This LGTM :shipit:

PR #578, opened 2022-11-03
Remove support for classes and strings in `field_transformers`. Only actual instances allowed.

### LINKED ISSUE #541: Don't copy instances of tranformer
### Problem Description

As a user, it is confusing that the HyperTransformer makes internal copies of transformer instances. 
- The transformer instances in the HyperTransformer config are not the actual ones used by the HyperTransformer. I cannot easily access any methods or attributes that may be available from the transformer.
- Any transformer instances I input are the not the actual objects used

eg. 
```python
>>> hyper_transformer.fit(data)
>>> transformer = hyper_transformer.get_config()['transformers']['address'] # assume this is a PseudoAnonymizedFaker
>>> transformer.get_mapping() # expected: an actual mapping 
{} 
# actual: a blank dictionary bc this transformer a copy; it's not the one that was actually fit
```

### Expected behavior

1. All the transformers available through the config should be the actual ones that are used. 
2. All instances that are passed in through `set_config` and `update_transformers` should be added to the config as-is (no copying)

### Additional Context
Note that there the `update_transformers_by_sdtype` method cannot use the actual instance because it needs to apply the transformer to multiple columns. See #540 for more details.

fealho:
Resolve #541.

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -555,6 +555,8 @@ class HyperTransformer:
             transformer.fit(data, field)
             self._transformers_sequence.append(transformer)
             data = transformer.transform(data)
+            if field in self._input_columns:
+                self.field_transformers[field] = transformer
 
             output_columns = transformer.get_output_columns()
             next_transformers = transformer.get_next_transformers()
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -358,6 +358,8 @@ class HyperTransformer:
     def update_sdtypes(self, column_name_to_sdtype):
         """Update the ``sdtypes`` for each specified column name.
 
+        The method may also update ``field_transformers`` to match the new sdtypes.
+
         Args:
             column_name_to_sdtype(dict):
                 Dict mapping column names to ``sdtypes`` for that column.
@@ -367,7 +369,6 @@ class HyperTransformer:
 
         update_columns = column_name_to_sdtype.keys()
         self._validate_update_columns(update_columns)
-
         self._validate_sdtypes(column_name_to_sdtype)
 
         transformers_to_update = {}
@@ -379,7 +380,7 @@ class HyperTransformer:
                     supported_sdtypes = current_transformer.get_supported_sdtypes()
 
                 if sdtype not in supported_sdtypes:
-                    transformers_to_update[column] = get_default_transformer(sdtype)
+                    transformers_to_update[column] = deepcopy(get_default_transformer(sdtype))
 
         self.field_sdtypes.update(column_name_to_sdtype)
         self.field_transformers.update(transformers_to_update)
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -1,9 +1,9 @@
 """Hyper transformer module."""
 
-from copy import deepcopy
 import inspect
 import json
 import warnings
+from copy import deepcopy
 
 import pandas as pd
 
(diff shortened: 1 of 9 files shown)


rdt/hyper_transformer.py
```diff
@@ -381,7 +380,7 @@ def update_sdtypes(self, column_name_to_sdtype):
                     supported_sdtypes = current_transformer.get_supported_sdtypes()
 
                 if sdtype not in supported_sdtypes:
-                    transformers_to_update[column] = get_default_transformer(sdtype)
+                    transformers_to_update[column] = deepcopy(get_default_transformer(sdtype))
```
fealho, line 383: get_default_transformer caches the values, so it always returns  references to the same instances, so we gotta copy them.

rdt/hyper_transformer.py
```diff
@@ -494,9 +493,10 @@ def _learn_config(self, data):
             if field not in self.field_transformers:
                 sdtype = self.field_sdtypes[field]
                 if sdtype in self._default_sdtype_transformers:
-                    self.field_transformers[field] = self._default_sdtype_transformers[sdtype]
+                    self.field_transformers[field] = deepcopy(
```
fealho, line 496: same

rdt/hyper_transformer.py
```diff
@@ -551,7 +551,6 @@ def _fit_field_transformer(self, data, field, transformer):
             self._output_columns.append(field)
 
         else:
-            transformer = get_transformer_instance(transformer)
```
fealho, line 554: We are removing this method all together.

rdt/hyper_transformer.py
```diff
@@ -351,7 +351,7 @@ def update_transformers_by_sdtype(
 
         for field, field_sdtype in self.field_sdtypes.items():
             if field_sdtype == sdtype:
-                self.field_transformers[field] = transformer_instance
+                self.field_transformers[field] = deepcopy(transformer_instance)
```
fealho, line 354: This makes so we don't used the passed transformer, but since that's gonna be deprecated, I don't think we should bother writing code for it.

amontanez24: what do you mean by writing code for it?

amontanez24 (approved):
I just had one small question but I think this looks good!

rdt/hyper_transformer.py
fealho: The motivation behind this PR was that we should directly use the instance the user passes us. By deepcopying it, we are not quite doing that (so the user may be confused why the instance they passed was not affected).

However, since we plan do deprecate the scenario where a user can pass an instance directly to `update_transformers_by_sdtype`, I don't think we should bother adding extra code + test cases for this scenario.

tests/unit/test_hyper_transformer.py
```diff
@@ -280,13 +275,12 @@ def test__learn_config(self, get_default_transformer_mock):
...
-            'categorical': categorical_transformer,
-            'bool': bool_transformer,
-            'datetime': datetime_transformer
-        }
+
+        assert isinstance(ht.field_transformers['integer'], FloatFormatter)
+        assert isinstance(ht.field_transformers['float'], ClusterBasedNormalizer)
+        assert isinstance(ht.field_transformers['categorical'], FrequencyEncoder)
+        assert isinstance(ht.field_transformers['bool'], BinaryEncoder)
+        assert isinstance(ht.field_transformers['datetime'], UnixTimestampEncoder)
```
pvk-developer, line 283: Just minor but I think that here we could ensure that `ht.field_transformers['integer'] == instance_integer`.
I would probably define the transformers before setting them to the field transformers and here ensure that they match the same id rather than just being an instance of those.

pvk-developer (approved):
LGTM! I just left a small suggestion, I think you already check for it on a later stage so it's not relevant to be merged!

tests/unit/test_hyper_transformer.py
fealho: `learn_config` uses o copy of the transformers, not the transformers themselves.

PR #579, opened 2022-11-09
Make `RegexGenerator` not to reset

### LINKED ISSUE #558: Make `RegexGenerator` not to reset itself
### Problem Description

Currently our `RegexGenerator` does not maintain its status, when we use `reverse_transform` the generator starts over again. It would be ideal to continue from where it was left the last time.

### Expected behavior

This functionality is essential when using `enforce_uniqueness`. The generator should generate new, unique values each time when called, if the generator can't generate new values an error should be raised.

```python
from rdt.transformers import RegexGenerator

data = pd.DataFrame({'a': np.arange(10)})
generator = RegexGenerator(enforce_uniqueness=True)

tr_data = generator.fit_transform(data, 'a')
first_try = generator.reverse_transform(tr_data)
second_try = generator.reverse_transform(tr_data)

pd.testing.assert_frame_equal(first_try, second_try)  # -> True
```
The two dataframes should be different or raise an error if there are no more items to be generated from the current generator.

### Additional Context

The reason why this is not working currently it's because the `generator` is being created on the `_reverse_transform` rather than being stored in the `instance` itself.

pvk-developer:
Resolves #558

### DIFF
--- a/rdt/transformers/text.py
+++ b/rdt/transformers/text.py
@@ -27,10 +27,18 @@ class RegexGenerator(BaseTransformer):
     INPUT_SDTYPE = 'text'
 
     def __init__(self, regex_format='[A-Za-z]{5}', enforce_uniqueness=False):
+        self.output_properties = {None: {'next_transformer': None}}
         self.enforce_uniqueness = enforce_uniqueness
         self.regex_format = regex_format
         self.data_length = None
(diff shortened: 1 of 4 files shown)


amontanez24 (approved):
LGTM! Good job!

fealho (approved):

PR #589, opened 2022-11-30
Add ability to control randomness

### LINKED ISSUE #584: Add ability to control randomness
### Problem Description

_As a user, there are times where it would be useful to obtain the same outputs across the same or different class instances when using RDT to reverse transform data. There are multiple places where randomness occurs in RDT, and ideally I'd be able to control all of them._

In the case of synthetic data, it is sometimes useful to be able to sample the same data. Right now, that won't work because RDT randomizes the null values, the faker values, adds noise during some transform operations and has an element of randomness in the `ClusterBasedNormalizer`. This makes it impossible to control sampling to a degree where the same results can be obtained reliably.

### Expected behavior

```python
ht = HyperTransformer()
ht.detect_config(data)

ht2 = HyperTransformer()
ht2.detect_config(data)
transformed2 = ht2.fit_transform(data)
reversed2 = ht.reverse_transform(transformed2)
```
In the case above, because the two `HyperTransformer`s are fit using the same data, the outputs should be exactly the same. That is, `transformed == transformed2` and `reversed == reversed2`.

#### Reverse transforming
```python
ht = HyperTransformer()
ht.detect_config(data)
ht.fit(data)

# later
synthetic_data = ht.reverse_transform(input)

ht.reset_randomization()
synthetic_data_2 = ht.reverse_transform(input)
```
In this case, `synthetic_data` and `synthetic_data_2` should be exactly the same. That is, even any generated PII values, regexes, etc. are the same. Note that if you do not reset randomization, then the data may be different as in the following case:

```python
synthetic_data = ht.reverse_transform(input)
synthetic_data_2 = ht.reverse_transform(input)
```

### Additional context

- To make this happen we want to be able to set a random seed for each portion of the workflow (ie. fit, transform, reverse_transform). Additionally, there are two different seeds that need to be controlled: the numpy random seed and the seed used by Faker.
- We can use this [PR](https://github.com/sdv-dev/CTGAN/pull/204/files) as inspiration. The main difference is that instead of handling torch, we need to handle faker, and that fit, transform and reverse_transform should all use different seeds. This way if a user samples, resets and samples again, the seed won't be affected by whether or not they refitted.
- The `reset_anonymization` method should be renamed to `reset_randomization` and should set each of the seeds back to their initial state.
- Should also resolve this issue #574

amontanez24:
resolves #584

This PR creates decorators that are used to set the numpy random seed before calling the method in question. It then sets the seed back to whatever it was before upon completing the method and updates the state of the random seeds for that method accordingly.

There is a dictionary created for each transformer instance that stores the random state for each method (`fit`, `transform` and `reverse_transform`).

There is a `reset_randomization` method that resets the state for `reverse_transform` back to it's initial hardcoded state. Additionally, the generators have to be able to reset. For Faker, this means hardcoding an initial seed as well.


rdt/transformers/categorical.py
```diff
@@ -69,6 +69,24 @@ def _get_intervals(data):
...
+                return data_is_na.loc[data_is_na == 1].index[0]
+
+            return data_new_index.loc[data_new_index == element].index[0]
+
+        augmented_frequencies[sortable_column_name] = frequencies.index.map(tie_breaker)
+        augmented_frequencies = augmented_frequencies.sort_values(
+            [column_name, sortable_column_name],
+            ascending=[False, True]
+        )
+        sorted_frequencies = augmented_frequencies[column_name]
```
amontanez24, line 89: The problem here is that `value_counts` is not deterministic. There fore, sometimes different values are assigned to different categories if they both occur the same amount. To fix this, I added a tie breaker based on the index of the first occurrence of that category. I couldn't use the category itself as a tie breaker because we allow for categories of different types, which can't be compared

rdt/transformers/base.py
```diff
@@ -1,12 +1,57 @@
...
+    """Context manager for managing the random state.
+
+    Args:
+        random_state (int or tuple):
+            The random seed or a tuple of (numpy.random.RandomState, torch.Generator).
+        set_model_random_state (function):
+            Function to set the random state on the model.
+    """
+    original_np_state = np.random.get_state()
+    random_np_state = random_states[method_name]
```
pvk-developer, line 23: is `random_states` a `dict` or `int or tuple` ? Looks like a `dict` to me

rdt/transformers/base.py
```diff
@@ -26,6 +74,31 @@ class BaseTransformer:
...
+
+        Args:
+            state (numpy.random.RandomState):
+                The numpy random state to set.
+            method_name (str):
+                The method to set it for.
+        """
+        if method_name not in self.random_states:
+            raise ValueError(
+                "'method_name' must be one of 'fit', 'transform' or 'reverse_transform'"
```
pvk-developer, line 94: Missing a `.` at the end.

rdt/transformers/base.py
```diff
@@ -1,12 +1,57 @@
...
 import numpy as np
 import pandas as pd
 
 
+@contextlib.contextmanager
+def set_random_states(random_states, method_name, set_model_random_state):
+    """Context manager for managing the random state.
+
+    Args:
+        random_state (int or tuple):
```
pvk-developer, line 17: typo - `random_state` -> `random_states`

pvk-developer (approved):
LGTM! :shipit: 
Great job!

rdt/transformers/base.py
amontanez24: it's a dict

fealho (commented):
Looks good!

fealho (approved):

PR #591, opened 2022-12-05
Fix `learn_rounding_scheme` for more than 14 digits

### LINKED ISSUE #556: Float formatter `learn_rounding_scheme` doesn't work on all digits
### Environment Details

* RDT version: 1.2.1

### Error Description

When using a FloatFormatter with `learn_rounding_scheme=True`, we expect the transformer to learn the maximum # of significant digits. In practice, we see the following:

1. If data has 0-14 digits, then the transformer learns rounding scheme [Working as intended] 
3. If the data has 15+ digits, then the transformer learns 0 digits, producing whole numbers instead [Bug] 

We expect that case 2 to work. Or as a fallback, at least stop enforcing the rounding if there are already a large number of digits.

### Steps to reproduce

```python
import pandas as pd
from rdt.transformers.numerical import FloatFormatter

# create test data with 16 digits
    'column': [1.1234567890123456]
})

ht = HyperTransformer()
ht.set_config({
    'sdtypes': { 'column': 'numerical' },
    'transformers': { 'column': FloatFormatter(learn_rounding_scheme=True)}  
})

t = ht.fit_transform(test_data)
ht.reverse_transform(t)
```

Output: 1.0 (no digits learned)

fealho:
Resolve #556.

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -78,12 +78,17 @@ class FloatFormatter(BaseTransformer):
     @staticmethod
     def _learn_rounding_digits(data):
         # check if data has any decimals
+        name = data.name
         data = np.array(data)
         roundable_data = data[~(np.isinf(data) | pd.isna(data))]
 
+        # Doesn't contain numbers
+        if len(roundable_data) == 0:
+            return None
+
         # Doesn't contain decimal digits
         if ((roundable_data % 1) == 0).all():
-            return None
+            return 0
 
         # Try to round to fewer digits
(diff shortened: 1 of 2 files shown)


amontanez24 (approved):
Good catch! I believe we used to think that returning None meant we didn't round at all, but now we have the line of code that says `data.round(self._rounding_digits or 0)`. I think we agreed to round to the `MAX_DECIMALS` but I think it would also be ok to just not round in this case.

pvk-developer (approved):

PR #602, opened 2022-12-23
Update errors

### LINKED ISSUE #599: Update errors
### Problem Description

__As a developer, it would be helpful to have more descriptive error types, so that different types of crashes could be handled separately.__

### Expected behavior

- Remove the [Error](https://github.com/sdv-dev/RDT/blob/master/rdt/errors.py#L8) class and replace all instances with the appropriate of the following. Add the new error type if not already created.
    - `InvalidConfigError`: Something is incorrect about the config. Use in any method that sets or updates the config:
        - column names in 'transformers' are different than 'sdtypes'
        - incorrect 'sdtypes'
        - incorrect transformer names
    - `InvalidDataError`: The data is ill-formed in some way. Use any time data is passed in and doesn't match the expected format
        - data in fit doesn't match config
        - data in transform doesn't match config
        - data in reverse_transform doesn't have the right columns"
    - `TransformerNotFitError`: Trying to use the HyperTransformer before fitting it
        - Transform (transform_subset) before fit
        - Reverse transform (reverse_transform_subset) before fit
        - create_anonymized_columns before fit
    - `TransformerInputError`: The input to the hypertransformer is incorrect
        - assigned transformer doesn't match the sdtype
    - `TransformerProcessingError`: Catch all for anything that doesn't fit into the above. **NOTE** - Only add this one if needed.

amontanez24:
resolves #599

### DIFF
--- a/rdt/errors.py
+++ b/rdt/errors.py
@@ -22,4 +22,4 @@ class TransformerInputError(Exception):
 
 
 class TransformerProcessingError(Exception):
-    """Error to raise when ."""
+    """Error to raise when transformer fails to complete some process (ie. anonymization)."""
--- a/rdt/errors.py
+++ b/rdt/errors.py
@@ -1,13 +1,25 @@
 """RDT Exceptions."""
 
 
+class ConfigNotSetError(Exception):
+    """Error to use when no config has been set or detected."""
+
+
+class InvalidConfigError(Exception):
+    """Error to raise when something is incorrect about the config."""
+
+
+class InvalidDataError(Exception):
+    """Error to raise when the data is ill-formed in some way."""
+
+
 class NotFittedError(Exception):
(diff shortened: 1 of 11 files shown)


rdt/errors.py
```diff
@@ -1,13 +1,25 @@
...
+class NotFittedError(Exception):
+    """Error to raise when ``transform`` or ``reverse_transform`` are used before fitting."""
 
 
 class TransformerInputError(Exception):
     """Error to raise when ``HyperTransformer`` receives an incorrect input."""
+
+
+class TransformerProcessingError(Exception):
+    """Error to raise when ."""
```
pvk-developer, line 25: Missing when the error is raised.

pvk-developer (changes requested):

pvk-developer (approved):
:shipit: LGTM!

PR #603, opened 2022-12-26
Add Python 3.10-3.11 and drop Python 3.6 support

### LINKED ISSUE #593: Support Python 3.10
Future versions of RDT should support Python 3.10 (and ideally 3.11).

We can also drop Python 3.6 since it has reached [end of life](https://devguide.python.org/versions/#versions).

pvk-developer:
Resolves #593


rdt/performance/datasets/datetime.py
```diff
@@ -67,7 +67,7 @@ def get_performance_thresholds():
             },
             'transform': {
                 'time': 5e-05,
-                'memory': 300.0
```
amontanez24, line 70: Should we boost it more just in case or has this passed consistently?

setup.py
```diff
@@ -15,18 +15,20 @@
...
+    "scikit-learn>=1.1.3,<2;python_version>='3.10'",
     'psutil>=5.7,<6',
-    'scikit-learn>=0.24,<2',
-    'pyyaml>=5.4.1,<6',
     'Faker>=10',
 ]
 
 copulas_requires = [
-    'copulas>=0.6.0,<0.7',
+    'copulas>=0.7.1.dev0,<0.8',
```
amontanez24, line 31: let's remember to change this before merging

tasks.py
```diff
@@ -39,20 +45,26 @@ def integration(c):
...
+    for python_version_match in re.finditer(r"python_version(<=?|>=?|==)\'(\d\.?)+\'", line):
         python_version = python_version_match.group(0)
-        comparison = re.search(r'(>=?|<=?)', python_version).group(0)
+        comparison = re.search(r'(>=?|<=?|==)', python_version).group(0)
         version_number = python_version.split(comparison)[-1].replace("'", "")
         comparison_function = COMPARISONS[comparison]
-        return comparison_function(platform.python_version(), version_number)
+        is_valid = is_valid and comparison_function(
+            pkg_resources.parse_version(platform.python_version()),
+            pkg_resources.parse_version(version_number),
```
amontanez24, line 63: what's going on here?

rdt/performance/datasets/datetime.py
amontanez24: @pvk-developer any thoughts on this?

amontanez24 (approved):
Once builds pass :shipit:

PR #610, opened 2023-01-12
The 'reset_randomization' should also apply to 'fit' and 'transform'

### LINKED ISSUE #608: The `reset_randomization` should also apply to `fit` and `transform`
### Environment Details

* RDT version: 1.3.0 (dev)
* Python version: Any
* Operating System: Any

### Description

The `reset_randomization` method is being used to control the determinism of the reverse transform only. It would make sense to also use this on the fit and forward transform.

**Fit**: Some statistical transformers such as `GaussianNormalizer` use random starting points to fit a univariate distribution. In the example below, we should expect that states 1 and 3 are exactly the same.

```python
ht.fit(data) # state 1
ht.fit(data) # state 2

ht.reset_randomization()
ht.fit(data) # state 3
```

**Forward Transform**: Randomization can appear due to adding noise. In the example below, we should expect that `transformed1` and `transformed3` are exactly the same.
```python
ht.fit(data)
transformed1 = ht.transform(data)
transformed2 = ht.transform(data)

ht.reset_randomization()
transformed3 = ht.transform(data)
```

amontanez24:
resolves #608

### DIFF
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -100,6 +100,8 @@ class BaseTransformer:
 
     def reset_randomization(self):
         """Reset the random state for ``reverse_transform``."""
+        self.set_random_state(self.INITIAL_FIT_STATE, 'fit')
+        self.set_random_state(self.INITIAL_TRANSFORM_STATE, 'transform')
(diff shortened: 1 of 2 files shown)


pvk-developer (approved):
LGTM 🚢

PR #622, opened 2023-03-16
Issue 619 faker seed

### LINKED ISSUE #619: Choose a different seed for each transformer
### Problem Description
If I have two columns that are both assigned to `AnonymizedFaker` (eg. for `uuid4`), then the reverse transform always produces the same exact values for both columns. This is unexpected, as I'd want each uuid column to have different values.

image

See [SDV issue 1303](https://github.com/sdv-dev/SDV/issues/1303).

### Expected behavior
We should change the behavior for the `AnonymizedFaker` and `PseudoAnonymizedFaker` transformers.

1. When first initializing the HyperTransformer, we should set a _different_ seed for each of these transformers. (The seed values should be deterministic. Eg. The transformer for `column1` is always set to seed=37.) As a result, these transformers will have a different seed. Each transformer should separately **store its seed**.
2. When [resetting the HyperTransformer](https://docs.sdv.dev/rdt/usage/hypertransformer/transformation#reset_randomization), these transformer should set their own individual seeds to what was stored in step (1)

fealho:
Resolve #619.


tests/unit/transformers/test_base.py
```diff
@@ -935,7 +935,7 @@ def test_fit(self):
             'b': [4, 5, 6],
             'c': [7, 8, 9]
         })
-        column = 'a'
+        column = ['a']
```
fealho, line 938: Need to add a test which passes a list of one value to get 100% coverage.

rdt/transformers/categorical.py
```diff
@@ -147,7 +147,7 @@ def _transform_by_category(self, data):
             if category is np.nan:
                 mask = data.isna()
             else:
-                mask = (data.to_numpy() == category)
+                mask = data.to_numpy() == category
```
fealho, line 150: My lint complained about these for some reason

rdt/transformers/pii/anonymizer.py
```diff
@@ -126,6 +124,7 @@ def _fit(self, data):
             data (pandas.Series):
                 Data to fit to.
         """
+        self.faker.seed_instance(self._random_seed)
```
fealho, line 127: This is the main change: we set the seed during `fit` instead of during initialization.

rdt/transformers/base.py
```diff
@@ -328,6 +330,11 @@ def fit(self, data, column):
             column (str):
                 Column name. Must be present in the data.
         """
+        if isinstance(column, list):
+            column = column[0]
+
+        hash_int = int(hashlib.sha256(column.encode('utf-8')).hexdigest(), 16)
+        self._random_seed = hash_int % ((2 ** 32) - 1)  # maximum value for a seed
```
amontanez24, line 337: Should we change `self.reset_randomization` to use this seed for the reverse_transform as well? Like
```python
self.set_random_state(self._random_seed, 'reverse_transform')
```

If this is only going to apply to generator type transformers then instead maybe we call this `_generator_random_seed`

fealho: This implementation simply generates a unique `_random_seed` for each transformer. Besides the `AnonymizedFaker`, no transformer currently uses it. The idea would be that, if we add any more random transformers - _which don't depend on the data_ - they can use this value to solve the issue of every column generating the same results.

If I understand you correctly, you are proposing the more general solution of changing the `INITIAL_X_STATE` seed. As I mentioned in the meeting yesterday, I had some trouble getting this implementation to work (particularly because `AnonymizedFaker` sets its seed directly, so changing the random state doesn't do anything). It's definitely doable, it just seemed unnecessary to make the seed of every state different for distinct transformers when we don't currently have any transformers which would benefit from it (besides possibly the `nan` substitution issue).

amontanez24: Can we rename it then. Rn it seems like this attribute is being used for all transformers but it isn't

rdt/transformers/base.py
```diff
@@ -328,6 +330,11 @@ def fit(self, data, column):
             column (str):
                 Column name. Must be present in the data.
         """
+        if isinstance(column, list):
+            column = column[0]
+
+        hash_int = int(hashlib.sha256(column.encode('utf-8')).hexdigest(), 16)
+        self._faker_random_seed = hash_int % ((2 ** 32) - 1)  # maximum value for a seed
```
pvk-developer, line 337: Move to `AnonymizedFaker`

rdt/transformers/pii/anonymizer.py
```diff
@@ -244,6 +248,7 @@ def _fit(self, columns_data):
             data (pandas.Series):
                 Data to fit the transformer to.
         """
+        self.faker.seed_instance(self._faker_random_seed)
```
pvk-developer, line 251: I think that we should compute `self._faker_random_seed` here as well, since this is not calling `super()`

pvk-developer (changes requested):
Just a minor change and after that it should be good to go!

pvk-developer (approved):
LGTM!

amontanez24 (approved):
:shipit:

PR #634, opened 2023-04-21
Remove pandas upper bound

### LINKED ISSUE #633: Remove upper bound for pandas
### Problem Description

With the newly released [pandas 2.0](https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html) it would nice to support it across the sdv-dev libraries.
 
### Expected behavior

The newly expected version for pandas should be only `>=` than the minimum supported without upper bound. After the changes are done, both minimum and newest version should work properly.

pvk-developer:
Resolves #633

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -80,14 +80,10 @@ def install_minimum(c):
                 continue
 
             line = line.strip()
-            if _validate_python_version(line):
+            if _validate_python_version(line) and 'copulas' not in line:
                 requirement = re.match(r'[^>]*', line).group(0)
                 requirement = re.sub(r"""['",]""", '', requirement)
-                if 'copulas' in line:
-                    version = 'git+https://github.com/sdv-dev/Copulas@issue-349-remove-pandas-upper-bound'
-
-                else:
-                    version = re.search(r'>=?[^(,|#)]*', line).group(0)
+                version = re.search(r'>=?[^(,|#)]*', line).group(0)
                 if version:
                     version = re.sub(r'>=?', '==', version)
                     version = re.sub(r"""['",]""", '', version)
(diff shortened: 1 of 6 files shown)


tasks.py
```diff
@@ -80,10 +80,10 @@ def install_minimum(c):
                 continue
 
             line = line.strip()
-            if _validate_python_version(line):
+            if _validate_python_version(line) and 'copulas' not in line:
```
frances-h, line 83: I think this should also be reverted?

tasks.py
```diff
@@ -96,7 +96,7 @@ def install_minimum(c):
             started = True
 
 
-    c.run(f'python -m pip install {" ".join(versions)}')
+    c.run(f'python -m pip install {" ".join(versions)} copulas')
```
frances-h, line 99: ditto

tests/integration/test_hyper_transformer.py
```diff
@@ -167,7 +167,7 @@ def test_hypertransformer_default_inputs():
     pd.testing.assert_frame_equal(transformed, expected_transformed)
 
     reversed_datetimes = pd.to_datetime([
-        '2010-01-09 20:34:17.142857216',
```
pvk-developer, line 170: This was actually a bug, in `python 3.7`, the format wasn't being learned, this change fixed it: `datetime_array = data[data.notna()].astype(str).to_numpy()`

tests/quality/test_quality.py
```diff
@@ -108,7 +109,7 @@ def get_transformer_regression_scores(data, sdtype, dataset_name, transformers,
                 'column': column,
                 'score': score
             })
-            scores = scores.append(row, ignore_index=True)
+            scores = scores._append(row, ignore_index=True)
```
frances-h, line 112: Can we use `pd.concat` here instead of `_append`? I think `pd.concat([scores, row], axis=1])` and then returning `scores.T` at the end should achieve the same result.

amontanez24 (approved):
:shipit: Thanks for fixing the datetime bug!

PR #636, opened 2023-04-21
Fix a very large regex combination to end in a endless loop

### LINKED ISSUE #635: `RegexGenerator` continues to have problems if there are too many possibilities
### Environment Details

* RDT version: 1.4.0 (latest)
* Python version: Any
* Operating System: Any

### Error Description
In the latest version of RDT, we fixed #623 and #624 so that the `RegexGenerator` could work with larger regexes. However, it still only supported Regex with less than 2^64 possible combinations.

If there are more combinations than that, then the `RegexGenerator` still fails to create those values.

### Steps to reproduce
Try the same code with the following regex:

```
regex_format = '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}'
```

Observe that the same issues are found.

### Solutions
We are using an `int64` today. Can we use a representation with a larger # of bits?

pvk-developer:
Resolves #635

### DIFF
--- a/rdt/transformers/utils.py
+++ b/rdt/transformers/utils.py
@@ -149,4 +149,4 @@ def strings_from_regex(regex, max_repeat=16):
             generators.append((generator, option, args))
             sizes.append(size)
 
-    return _from_generators(generators, max_repeat), np.prod(sizes)
+    return _from_generators(generators, max_repeat), np.prod(sizes, dtype=np.complex128)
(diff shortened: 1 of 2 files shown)


amontanez24 (approved):
LGTM!

R-Palazzo (approved):
LGTM :)

PR #645, opened 2023-05-12
Randomize nulls

### LINKED ISSUE #639: Fix randomization when creating `null` values
* RDT version: 1.4.0 (latest)

### Error Description
Currently, some tansformers have a default strategy of randomly assigning null values (in the same proportion as observed during fit). However, since each of these transformers use the same seed, the null values are assigned in exactly the same places.  This leads to some odd results, as it appears that the nulls are fixed to particular rows.

Affected transformers are the numerical and datetime ones:
- Numerical: [FloatFormatter](https://docs.sdv.dev/rdt/transformers-glossary/browse-transformers/floatformatter), [ClusterBasedNormalizer](https://docs.sdv.dev/rdt/transformers-glossary/browse-transformers/clusterbasednormalizer), [GaussianNormalizer](https://docs.sdv.dev/rdt/transformers-glossary/browse-transformers/gaussiannormalizer)
- Datetime: [UnixTimestampEncoder](https://docs.sdv.dev/rdt/transformers-glossary/browse-transformers/unixtimestampencoder), [OptimizedTimestampEncoder](https://docs.sdv.dev/rdt/transformers-glossary/browse-transformers/optimizedtimestampencoder)

### Fix
Choose a random seed based on the column name and x values of the data itself.

We made a similar fix for #619, but I believe we only did it for the PII transformers.

### Steps to reproduce

Reproduce this issue when using the RDT transformers through the SDV.

```python
import pandas as pd
import numpy as np
from sdv.metadata import SingleTableMetadata
from sdv.single_table import GaussianCopulaSynthesizer

data = pd.DataFrame(data={
    'low':    [1, 4, np.nan, 0,      4,      np.nan, np.nan, 5,      np.nan],
    'middle': [2, 5, 3,      np.nan, 5,      np.nan, 5,      np.nan, np.nan],
    'high':   [3, 7, 8,      4,      np.nan, 9,      np.nan, np.nan, np.nan]
})

metadata_dict = {
    'columns': {
        'low': { 'sdtype': 'numerical' },
        'middle': { 'sdtype': 'numerical'},
        'high': { 'sdtype': 'numerical'}
    }
}

metadata = SingleTableMetadata.load_from_dict(metadata_dict)
synth = GaussianCopulaSynthesizer(metadata)
synth.fit(data)
synth.sample(10)
```

image

fealho:
Resolve #639.

It also changes the behavior of the randomization. The `fit` random state is set during initialization, while the `transform` and `reverse_transform` states are only set after seeing the data provided to `fit`. `reset_randomization` returns the random states to the one seen during `fit`.

### DIFF
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -1,6 +1,7 @@
 """BaseTransformer module."""
 import abc
 import contextlib
+import hashlib
 import inspect
 import warnings
 from functools import wraps
@@ -318,6 +319,19 @@ class BaseTransformer:
         """
         raise NotImplementedError()
 
+    def _set_seed(self, data):
+        hash_value = self.get_input_column()
+        for value in data.head(5):
+            hash_value += str(value)
+
+        hash_value = int(hashlib.sha256(hash_value.encode('utf-8')).hexdigest(), 16)
+        random_seed = hash_value % ((2 ** 32) - 1)  # maximum value for a seed
+        self.random_states = {
+            'fit': np.random.RandomState(seed=random_seed),
+            'transform': np.random.RandomState(seed=random_seed + 1),
+            'reverse_transform': np.random.RandomState(seed=random_seed + 2)
+        }
+
     @random_state
     def fit(self, data, column):
         """Fit the transformer to a ``column`` of the ``data``.
@@ -329,6 +343,7 @@ class BaseTransformer:
                 Column name. Must be present in the data.
         """
         self._store_columns(column, data)
+        self._set_seed(data)
         columns_data = self._get_columns_data(data, self.columns)
         self._fit(columns_data)
         self._build_output_columns(data)
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -78,8 +78,8 @@ class BaseTransformer:
         self.output_properties = {None: {'sdtype': 'float', 'next_transformer': None}}
         self.random_states = {
             'fit': self.INITIAL_FIT_STATE,
-            'transform': np.random.RandomState(42),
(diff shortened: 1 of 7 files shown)


rdt/transformers/base.py
```diff
@@ -65,21 +66,20 @@ class BaseTransformer:
...
+    random_seed = 42
 
     def __init__(self):
         self.output_properties = {None: {'sdtype': 'float', 'next_transformer': None}}
         self.random_states = {
             'fit': self.INITIAL_FIT_STATE,
-            'transform': self.INITIAL_TRANSFORM_STATE,
-            'reverse_transform': self.INITIAL_REVERSE_TRANSFORM_STATE
+            'transform': np.random.RandomState(42),
+            'reverse_transform': np.random.RandomState(42)
```
amontanez24, line 82: Can we just set these values to None to start?

rdt/transformers/base.py
```diff
@@ -318,6 +320,19 @@ def _fit(self, columns_data):
         """
         raise NotImplementedError()
 
+    def _set_seed(self, data):
+        hash_value = self.get_input_column()
+        for value in data.head(5):
+            hash_value += str(value)
+
+        hash_value = int(hashlib.sha256(hash_value.encode('utf-8')).hexdigest(), 16)
+        self.random_seed = hash_value % ((2 ** 32) - 1)  # maximum value for a seed
```
amontanez24, line 329: Since we store the random seed on the object, maybe we should just use a uuid and then we could even get the seed before fitting

pvk-developer (approved):
LGTM!

PR #647, opened 2023-05-17
Add add-on detection for transformers

### LINKED ISSUE #646: Add add-on detection for premium transformers
### Problem Description

_As a user, it would be useful to be able to import enterprise transformers in the same way I import public ones._

### Expected behavior

- Add add-on detection for transformers. There will be two types we need to support:
    1. Transformers to be added to already existing modules. An example of this is when we want to add premium transformers to the numerical module so that a user can do
    ```python
    from rdt.transformers.numerical import PremiumNumericalTransformer
    ```
    2. Transformers in an entirely new module. An example could be if we wanted to add email transformers so a user could do
    ```python
    from rdt.transformers.email import PremiumEmailTransformer
    ``` 

### Additional context

- When creating new transformer add-ons, the entrypoint group name we use should indicate which case it falls under. A proposal could be to do something like `rdt.new_transformers_module` vs `rdt.existing_transformers_module`. In the former case we would have to add the module to rdt.transformers and in the latter case we would have to add all the new module in the add on to the existing sdtype transformer module (ie. rdt.numerical)
- We can use something similar to the add-on logic [here](https://github.com/sdv-dev/SDV/blob/master/sdv/_addons.py#L7)

frances-h:
Resolve #646


rdt/__init__.py
```diff
@@ -92,3 +93,81 @@ def get_demo(num_rows=5):
...
+
+    Args:
+        group (str):
+            The name of the entry points group to load.
+        parent_globals (dict):
+            The caller's global scope. Modules will be added to the parent's global scope through
+            their name.
+        add_all (bool):
+            Whether to also add everything in the add-on's ``module.__all__`` to the parent's
+            global scope. Defaults to ``False``.
```
pvk-developer, line 109: Not needed anymore right?

rdt/__init__.py
```diff
@@ -92,3 +93,81 @@ def get_demo(num_rows=5):
...
+
+        if module_path[0] != __name__:
+            msg = (
+                f"Failed to load '{entry_point.name}'. Expected base module to be '{__name__}'"
+                f", found '{module_path[0]}'."
+            )
+            warnings.warn(msg)
+            continue
+
+        base_module = modules[__name__]
```
pvk-developer, line 131: Could `modules` be `sys.modules` ? Took me a while to understand where this variable magically appears from.

rdt/__init__.py
```diff
@@ -92,3 +93,81 @@ def get_demo(num_rows=5):
...
+                    base_object = attrgetter('.'.join(split_object[:-1]))(base_module)
+
+                setattr(base_object, object_path[-1], addon)
+            except AttributeError:
+                msg = (
+                    f"Failed to load '{entry_point.name}'. Cannot find "
+                    f"'{'.'.join(split_object[:-1])}' in submodule '{'.'.join(module_path)}'."
+                )
+                warnings.warn(msg)
+                continue
```
pvk-developer, line 170: I would suggest to move all this to a `process_entry_point` function. And maybe try to even split that into multiple steps if possible.

pvk-developer (changes requested):

tests/unit/test___init__.py
```diff
@@ -38,3 +40,149 @@ def test_get_demo_many_rows():
...
+    entry_point.name = 'rdt.submodule.entry_name'
+    entry_point.load.return_value = 'entry_point'
+    entry_points_mock.return_value = [entry_point]
+
+    # Run
+    _find_addons()
+
+    # Assert
+    entry_points_mock.assert_called_once_with(group='rdt_modules')
+    modules_mock['rdt'].submodule.entry_name == 'entry_point'
```
fealho, line 60: assert

tests/unit/test___init__.py
```diff
@@ -38,3 +40,149 @@ def test_get_demo_many_rows():
...
+    entry_point.name = 'rdt.submodule:entry_object.entry_method'
+    entry_point.load.return_value = 'new_method'
+    entry_points_mock.return_value = [entry_point]
+
+    # Run
+    _find_addons()
+
+    # Assert
+    entry_points_mock.assert_called_once_with(group='rdt_modules')
+    modules_mock['rdt'].submodule.entry_name.entry_object.entry_method == 'new_method'
```
fealho, line 78: assert

pvk-developer (approved):
LGTM! Thanks for addressing the comments!

fealho (approved):
After you fix lint, :shipit:

PR #649, opened 2023-05-24
Allow no op handling strategy for missing values

### LINKED ISSUE #644: Allow a no-op handling strategy for missing values (nulls)
### Problem Description
In certain use cases, it makes sense not to do anything with missing values. That is, I would like the `fit`, `transform` and `reverse_transform` to largely ignore missing values and just pass them through.

Currently this is not possible for many of the transformers: (a) on the forward transform, you have to specify a strategy to fill in missing values and (b) on the reverse transform, the missing values are created either randomly or based on another column.

### Expected behavior

1. The `missing_value_replacement` parameter should accept a value of `None`. If selected, then do nothing with missing values on the forward transform (i.e. do not replace them with anything)
2. Deprecate the `model_missing_values` parameter. If a user adds this parameter, show a warning asking them to use the new parameter instead (described below)
3. Add a parameter called `missing_value_generation` that accepts 3 options:
    - (default) `'random'`: Randomly generates missing values based on the % of missing values observed during fit. This is the equivalent of today's `model_missing_values=False`
    - `'from_column'`: Creates a binary column that describes whether the original value was missing. Then uses it to recreate missing values. This is the equivalent of today's `model_missing_values=True`
    - `None` (**new**): Do nothing with missing values on the reverse transform. Simply pass in whatever data we get through (nulls and all).

**Which transformers does this apply to?** At minimum, this should apply to [FloatFormatter](https://docs.sdv.dev/rdt/transformers-glossary/browse-transformers/floatformatter), [UnixTimestampEncoder](https://docs.sdv.dev/rdt/transformers-glossary/browse-transformers/unixtimestampencoder) and [OptimizedTimestampEncoder](https://docs.sdv.dev/rdt/transformers-glossary/browse-transformers/optimizedtimestampencoder). Optionally, we may also apply it to any of the other numerical transformers that can handle missing values (GaussianNormalizer, etc.).

### Additional context

Maintain backwards compatibility. If the user has provided a `model_missing_values` parameter, then we can convert it to the equivalent `missing_value_generation` parameter. Always show a `FutureWarning` if this parameter is included.

```python
transformer = FloatFormatter(model_missing_values=True)

FutureWarning: Future versions of RDT will not support the 'model_missing_values' parameter.
Please switch to using the 'missing_value_generation' parameter to select your strategy.
```

pvk-developer:
Resolves #644

### DIFF
--- a/rdt/transformers/boolean.py
+++ b/rdt/transformers/boolean.py
@@ -20,20 +20,33 @@ class BinaryEncoder(BaseTransformer):
             Indicate what to replace the null values with. If the string ``'mode'`` is given,
             replace them with the most common value.
             Defaults to ``mode``.
+        missing_value_generation (str or None):
+            The way missing values are being handled. There are three strategies:
+
+                * ``RANDOM``: Randomly generates missing values based on the percentage of
+                  missing values.
+                * ``FROM_COLUMN``: Creates a binary column that describes whether the original
+                  value was missing. Then use it to recreate missing values.
+                * ``None``: Do nothing with the missing values on the reverse transform. Simply
+                  pass whatever data we get through.
+
         model_missing_values (bool):
-            Whether to create a new column to indicate which values were null or not. The column
-            will be created only if there are null values. If ``True``, create the new column if
-            there are null values. If ``False``, do not create the new column even if there
-            are null values. Defaults to ``False``.
+            **DEPRECATED** Whether to create a new column to indicate which values were null or
+            not. The column will be created only if there are null values. If ``True``, create
+            the new column if there are null values. If ``False``, do not create the new column
+            even if there are null values. Defaults to ``False``.
     """
 
     INPUT_SDTYPE = 'boolean'
     null_transformer = None
 
-    def __init__(self, missing_value_replacement='mode', model_missing_values=False):
+    def __init__(self, missing_value_replacement='mode', missing_value_generation='RANDOM',
+                 model_missing_values=None):
         super().__init__()
+        self.missing_value_generation = missing_value_generation
         self._set_missing_value_replacement('mode', missing_value_replacement)
-        self.model_missing_values = model_missing_values
+        if model_missing_values is not None:
+            self._set_model_missing_values(model_missing_values)
 
     def _fit(self, data):
         """Fit the transformer to the data.
@@ -44,7 +57,7 @@ class BinaryEncoder(BaseTransformer):
         """
         self.null_transformer = NullTransformer(
             self.missing_value_replacement,
-            self.model_missing_values
+            self.missing_value_generation
         )
         self.null_transformer.fit(data)
         if self.null_transformer.models_missing_values():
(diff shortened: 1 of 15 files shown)


amontanez24 (commented):
This is looking good! My main concern is about the order of the arguments

rdt/transformers/boolean.py
```diff
@@ -20,20 +20,33 @@ class BinaryEncoder(BaseTransformer):
...
     """
 
     INPUT_SDTYPE = 'boolean'
     null_transformer = None
 
-    def __init__(self, missing_value_replacement='mode', model_missing_values=False):
+    def __init__(self, missing_value_replacement='mode', missing_value_generation='RANDOM',
+                 model_missing_values=None):
         super().__init__()
+        self.missing_value_generation = missing_value_generation
```
R-Palazzo, line 46: Should we move
`self.missing_value_generation = missing_value_generation` after :
```
if model_missing_values is not None:
            self._set_model_missing_values(model_missing_values)
```
To avoid any wrong re-definition of `missing_value_generation` if the user also uses `model_missing_values`.

rdt/transformers/null.py
R-Palazzo, line 1: Update docstring in the `NullTransformer` definition (L20- L24)

R-Palazzo (changes requested):
Great job! Just let one question, thanks for the cleaning in the test ;)

frances-h (commented):
Overall looks good! Should we add a check somewhere to validate that whatever gets passed in for `missing_value_generation` is one of `None`, `'FROM_COLUMN'`, or `'RANDOM'`? I think it just silently defaults to using `None` right now, which could be confusing.

pvk-developer:
> Overall looks good! Should we add a check somewhere to validate that whatever gets passed in for `missing_value_generation` is one of `None`, `'FROM_COLUMN'`, or `'RANDOM'`? I think it just silently defaults to using `None` right now, which could be confusing.

I had the same concerns while implementing it, maybe raise a value error for it ? 🤔

rdt/transformers/boolean.py
pvk-developer: The reason why first we set `missing_value_generation` is to store (usually) the default value, if a user specifies `model_missing_values` I expect that to dominate over the `default` or `missing_value_generation` and if the user is confused and providing both of them, only the `model_missing_values` will be considered.

rdt/transformers/null.py
pvk-developer: Good catch 👍🏻

rdt/transformers/datetime.py
```diff
@@ -44,8 +44,8 @@ class UnixTimestampEncoder(BaseTransformer):
     INPUT_SDTYPE = 'datetime'
     null_transformer = None
 
-    def __init__(self, missing_value_replacement='mean', missing_value_generation='RANDOM',
-                 model_missing_values=None, datetime_format=None):
+    def __init__(self, missing_value_replacement='mean', model_missing_values=None,
```
amontanez24, line 47: we should switch the order in the docstrings as well

amontanez24 (approved):
LGTM! Thanks for addressing the comments

R-Palazzo (approved):
LGTM!

PR #652, opened 2023-06-01
Fix deprecated `model_missing_values`

pvk-developer:


rdt/transformers/datetime.py
```diff
@@ -168,11 +162,9 @@ class OptimizedTimestampEncoder(UnixTimestampEncoder):
             Indicate what to replace the null values with. If the strings ``'mean'`` or ``'mode'``
             are given, replace them with the corresponding aggregation.
             Defaults to ``mean``.
-        model_missing_values (bool):
```
fealho, line 171: Shouldn't we remove the docstrings when the parameter is removed? Keeping it there with the `DEPRECATED` message makes it clear what's going on, while removing it just makes it seem we forgot the docstring of the extra parameter.

amontanez24: I agree that we shouldn't get rid of the docstring

rdt/transformers/base.py
```diff
@@ -325,7 +335,7 @@ def __repr__(self):
         instanced = {
             key: getattr(self, key)
             for key in keys
-            if hasattr(self, key)
+            if key != 'model_missing_values'  # Remove after deprecation
```
amontanez24, line 338: Do we not need to do the `hasattr` check anymore?

rdt/transformers/base.py
```diff
@@ -109,6 +109,16 @@ def reset_randomization(self):
             'reverse_transform': np.random.RandomState(self.random_seed + 1)
         }
 
+    @property
+    def model_missing_values(self):
```
amontanez24, line 113: the categorical transformers don't have this property but would get it from this right? Should we just put it in the ones that don't have it? I think it would crash on the `FrequencyEncoder` since there is no `self.missing_value_generation`

amontanez24 (changes requested):

amontanez24 (approved):
Nice work! 🎉 :shipit:

PR #654, opened 2023-06-14
Add add-on modules to sys.modules

### LINKED ISSUE #653: Add add-on modules to sys.modules
### Problem Description

_As a user, it would be useful if importing for premium features worked the same as for public features._

### Expected behavior

For premium transformers, we sometimes add entirely new modules for a new sdtype (ie. `phone_number`). In those cases, we want to be able to do
```python
from rdt.transformers.new_sdtype import NewTransformer
```
Currently this is not supported. To access the new transformers you have to do
```python
from rdt.transformers import new_sdtype

new_sdtype.NewTransformer
```
To fix this, we need to add the new modules to sys.modules, since that is the dict python uses for imports.

amontanez24:
resolves #653

### DIFF
--- a/rdt/__init__.py
+++ b/rdt/__init__.py
@@ -164,8 +164,10 @@ def _find_addons():
             warnings.warn(msg)
             continue
 
-        if isinstance(addon, ModuleType) and addon.__name__ not in sys.modules:
-            sys.modules[addon.__name__] = addon
+        if isinstance(addon, ModuleType):
+            addon_module_name = f'{addon_target.__name__}.{addon_name}'
+            if addon_module_name not in sys.modules:
+                sys.modules[addon_module_name] = addon
 
         setattr(addon_target, addon_name, addon)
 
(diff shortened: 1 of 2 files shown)


rdt/__init__.py
```diff
@@ -157,6 +164,9 @@ def _find_addons():
             warnings.warn(msg)
             continue
 
+        if isinstance(addon, ModuleType) and addon.__name__ not in sys.modules:
+            sys.modules[addon.__name__] = addon
```
pvk-developer, line 168: This will not work as expected because the `addon.__name__` is just the base name of the addon being added, it has to include the `addon_target` as well.
```python
target_sys_name = f'{addon_target}.{addon.__name__}'
sys.modules[target_sys_name] = addon
```

amontanez24: The `addon.__name__` is actually the whole entry point name, but you're right it needs to be the target name added to the addon_name

pvk-developer (approved):
:shipit: LGTM!

frances-h (approved):
🚢

PR #661, opened 2023-06-27
Fix handling of datetime dtypes

### LINKED ISSUE #657: The datetime transformers don't give me back the same `dtype` sometimes
### Environment Details

* RDT version: 1.5.0
* Python version: Any
* Operating System: Any

### Error Description
In some rare cases, I may have a datetime column that is represented as a float or integer dtype. When this happens, I will always have a **datetime_format** string that specifies how to read this.

Example:
- A column of all nan values
- A column with format `'%Y%m%d'` such as `[20120212, 20100321, 20091102, ...]`

If this happens, then the output dtype (on the `reverse_transform`) is not the same dtype as the input.

### Proposed Fix
On the **fit**:
- Store the original `_dtype` of the input column

On the **transform**:
1. Convert it to a string (`astype("string")` to make sure the NaNs are properly converted)
2. Parse to `pd.datetime` to using the provided `datetime_format`
3. Convert from `pd.datetime` -> unix timestamp

On the **reverse_transform**:
1. Convert from unix timestamp -> `pd.datetime`
2. Convert it to a string (using `datetime_format`)
3. Convert back to the original `_dtype`

frances-h:
Resolve #657

### DIFF
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -1,7 +1,7 @@
 """Transformer for datetime data."""
 import numpy as np
 import pandas as pd
-from pandas.api.types import is_datetime64_dtype
+from pandas.api.types import is_datetime64_dtype, is_numeric_dtype
 from pandas.core.tools.datetimes import _guess_datetime_format_for_array
 
 from rdt.transformers.base import BaseTransformer
@@ -55,7 +55,7 @@ class UnixTimestampEncoder(BaseTransformer):
         self._dtype = None
 
     def _convert_to_datetime(self, data):
-        if data.dtype == 'object':
+        if self.datetime_format or not is_numeric_dtype(data):
             try:
                 pandas_datetime_format = None
                 if self.datetime_format:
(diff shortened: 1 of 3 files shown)


rdt/transformers/datetime.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, missing_value_replacement='mean', model_missing_values=None,
         self._dtype = None
 
     def _convert_to_datetime(self, data):
```
amontanez24, line 57: Can we actually add a docstring here explaining that we don't convert if the dtype is numeric and there is no format because we assume it's a timestamp in that case?

tests/unit/transformers/test_datetime.py
```diff
@@ -430,6 +430,39 @@ def test__reverse_transform_datetime_format_with_strftime_formats(self):
...
+        ute._dtype = 'object'
+        ute.null_transformer = NullTransformer('mean')
+
+        # Run
+        output = ute._reverse_transform(transformed)
+
+        # Assert
+        expected = pd.Series(['Jan 1, 2020', 'Feb 1, 2020', np.nan])
+        if 'windows' not in platform.system().lower():
+            pd.testing.assert_series_equal(output, expected)
```
amontanez24, line 448: I see we do this elsewhere but do you know why we skip for windows?

amontanez24 (approved):
Just a minor comment and question but looks good! 🚢 🇮🇹

tests/unit/transformers/test_datetime.py
frances-h: Not sure. I can't test locally since I don't have a windows machine, but I can try removing it and seeing what happens to the job

fealho (approved):

PR #669, opened 2023-07-07
`AnonymizedFaker` should be able to model and generate missing values

### LINKED ISSUE #660: `AnonymizedFaker` should be able to model and generate missing values
### Problem Description
Sometimes I may have a PII column with missing values. For example a `secondary_address` represents the 2nd line of an address (such as an apartment # or suite #). It is not always needed so I expect it to be blank a lot of the time.

However, if I apply the `AnonymizedFaker` to this data, then the new, anonymized data is all completely filled in. The transformer does not learn anything about missing values. 

### Expected behavior
Add a `missing_value_generation` argument to the AnonymizedFaker that works similar to the [FloatFormatter](https://docs.sdv.dev/rdt/transformers-glossary/numerical/floatformatter).
- (default) `'random'`: During fit, learn the % of missing values. Then during reverse transform, randomly assign missing values in the same proportion as the original
- ~~`'from_column'`: During fit, create a new column that stores whether the original values were missing or not. Then during reverse transform, use this column to determine whether to create anonymous values or keep them missing.~~ _This won't be supported yet. It will cause issues with creating anonymized data from scratch._
- `None` (status quo): Don't learn anything during fit. Then during reverse transform, don't create any missing values.

### Additional context
- There is no need to add a `missing_value_replacement` parameter for this transformer, as the original column will always be dropped (nothing to replace)

R-Palazzo:
Resolve #660

### DIFF
--- a/rdt/transformers/pii/anonymizer.py
+++ b/rdt/transformers/pii/anonymizer.py
@@ -190,11 +190,9 @@ class AnonymizedFaker(BaseTransformer):
             ) from exception
 
         if self.missing_value_generation == 'random':
-            nan_mask = np.random.choice(
-                [True, False], size=sample_size,
-                p=[self._nan_frequency, 1 - self._nan_frequency]
-            )
-            reverse_transformed[nan_mask] = np.nan
+            num_nans = int(self._nan_frequency * sample_size)
+            nan_indices = np.random.choice(sample_size, num_nans, replace=False)
+            reverse_transformed[nan_indices] = np.nan
 
         return reverse_transformed
 
--- a/rdt/transformers/pii/anonymizer.py
+++ b/rdt/transformers/pii/anonymizer.py
@@ -85,7 +85,7 @@ class AnonymizedFaker(BaseTransformer):
             )
 
     def __init__(self, provider_name=None, function_name=None, function_kwargs=None,
-                 locales=None, enforce_uniqueness=False):
+                 locales=None, enforce_uniqueness=False, missing_value_generation='random'):
         super().__init__()
         self.data_length = None
         self.enforce_uniqueness = enforce_uniqueness
@@ -106,6 +106,15 @@ class AnonymizedFaker(BaseTransformer):
         self.faker = faker.Faker(self.locales)
         if self.locales:
             self._check_locales()
+
+        if missing_value_generation not in ['random', None]:
+            raise TransformerInputError(
+                f"Missing value generation '{missing_value_generation}' is not supported "
+                "for AnonymizedFaker. Please use either 'random' or None."
+            )
+
+        self.missing_value_generation = missing_value_generation
+        self.nan_frequency = 0.0
 
     def reset_randomization(self):
         """Create a new ``Faker`` instance."""
@@ -138,6 +147,8 @@ class AnonymizedFaker(BaseTransformer):
         """
         self._set_faker_seed(data)
         self.data_length = len(data)
+        if self.missing_value_generation == 'random':
+            self.nan_frequency = data.isna().sum() / len(data)
 
     def _transform(self, _data):
         """Drop the input column by returning ``None``."""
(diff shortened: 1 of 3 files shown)


rdt/transformers/pii/anonymizer.py
```diff
@@ -85,7 +85,7 @@ def _check_locales(self):
             )
 
     def __init__(self, provider_name=None, function_name=None, function_kwargs=None,
-                 locales=None, enforce_uniqueness=False):
+                 locales=None, enforce_uniqueness=False, missing_value_generation='random'):
```
amontanez24, line 88: can we add this to the docstring as well?

amontanez24 (changes requested):

rdt/transformers/pii/anonymizer.py
```diff
@@ -107,6 +107,15 @@ def __init__(self, provider_name=None, function_name=None, function_kwargs=None,
...
             self._check_locales()
 
+        if missing_value_generation not in ['random', None]:
+            raise TransformerInputError(
+                f"Missing value generation '{missing_value_generation}' is not supported "
+                "for AnonymizedFaker. Please use either 'random' or None."
+            )
+
+        self.missing_value_generation = missing_value_generation
+        self.nan_frequency = 0.0
```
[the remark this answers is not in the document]
R-Palazzo: Yes done in [b597b31](https://github.com/sdv-dev/RDT/pull/669/commits/b597b315a4fd720bbb8a8a9ce6dd58191112b50d)

rdt/transformers/pii/anonymizer.py
R-Palazzo: Yes, done in [b597b31](https://github.com/sdv-dev/RDT/pull/669/commits/b597b315a4fd720bbb8a8a9ce6dd58191112b50d)

pvk-developer (approved):
LGTM!

amontanez24 (approved):
LGTM!

rdt/transformers/pii/anonymizer.py
```diff
@@ -170,6 +189,13 @@ def _reverse_transform(self, data):
                 f"('{self.get_input_column()}')."
             ) from exception
 
+        if self.missing_value_generation == 'random':
+            nan_mask = np.random.choice(
+                [True, False], size=sample_size,
+                p=[self._nan_frequency, 1 - self._nan_frequency]
+            )
+            reverse_transformed[nan_mask] = np.nan
```
amontanez24, line 197: we should add a unit test for this part

tests/integration/transformers/pii/test_anonymizer.py
```diff
@@ -69,6 +69,26 @@ def test_anonymizedfaker_with_nans():
...
 
+    pd.testing.assert_frame_equal(transformed, expected_transformed)
+    assert len(reverse_transform['username']) == 5
+    assert reverse_transform['username'].isna().sum() == 2
+
+
+def test_anonymizedfaker_with_nans_missing_value_generation_none():
+    """End to end test settings missing_value_generation=None."""
+    data = pd.DataFrame({
+        'id': [1, 2, 3, 4, 5],
```
amontanez24, line 80: Can we add a test for when it's random as well?

amontanez24 (changes requested):

rdt/transformers/pii/anonymizer.py
R-Palazzo: Yes, done in [c8823f1](https://github.com/sdv-dev/RDT/pull/669/commits/c8823f125084af19baaae4a0e66ddac99f6d0acc)

tests/integration/transformers/pii/test_anonymizer.py
R-Palazzo: I think it's the test `test_anonymizedfaker_with_nans()` just before because it's random by default.

PR #670, opened 2023-07-11
Update the default transformers that HyperTransformer assigns to each sdtype

### LINKED ISSUE #664: Update the default transformers that HyperTransformer assigns to each sdtype
### Problem Description
Currently, the default transformers used by HyperTransformer are different than the ones used by SDV. Additionally, they may also be stale (eg. FrequencyEncoder is a candidate for deprecation).

### Expected behavior
The HyperTransformer's defaults should be the same/similar to SDVs.

```python
_DEFAULT_TRANSFORMERS_BY_SDTYPE = {
        'numerical': rdt.transformers.FloatFormatter(
            learn_rounding_scheme=False,
            enforce_min_max_values=False,
            missing_value_replacement='mean',
            missing_value_generation='random',
        ),
        'categorical': rdt.transformers.LabelEncoder(add_noise=True),
        'boolean': rdt.transformers.LabelEncoder(add_noise=True),
        'datetime': rdt.transformers.UnixTimestampEncoder(
            missing_value_replacement='mean',
            missing_value_generation='random',
        ),
        'id': rdt.transformers.RegexGenerator()
    }
```

Additionally, the RDT has a `'pii'` sdtype. We can consider defaulting that to `AnonymizedFaker` (no params necessary).

R-Palazzo:
Resolve #664 

Few things about this PR.
- I removed the `_default_sdtype_transformers` of the `HyperTransformer` because it was never used. Maybe we want to define it properly and use it.
- I added the `dtype` attribute similar to the one of the `FrequencyEncoder`, so now the `LabelEncoder` can work with pandas `category` type. This also ensures that boolean columns are reverse-transformed to booleans as well.

### DIFF
--- a/rdt/transformers/__init__.py
+++ b/rdt/transformers/__init__.py
@@ -99,6 +99,7 @@ DEFAULT_TRANSFORMERS = {
         missing_value_replacement='mean',
         missing_value_generation='random',
     ),
+    'id': RegexGenerator(),
     'pii': AnonymizedFaker(),
 }
 
--- a/rdt/transformers/__init__.py
+++ b/rdt/transformers/__init__.py
@@ -87,19 +87,11 @@ TRANSFORMERS = {
 
 
 DEFAULT_TRANSFORMERS = {
-    'numerical': FloatFormatter(
-        learn_rounding_scheme=False,
-        enforce_min_max_values=False,
-        missing_value_replacement='mean',
-        missing_value_generation='random',
-    ),
+    'numerical': FloatFormatter(),
     'categorical': LabelEncoder(add_noise=True),
     'boolean': LabelEncoder(add_noise=True),
-    'datetime': UnixTimestampEncoder(
-        missing_value_replacement='mean',
-        missing_value_generation='random',
-    ),
-    'id': RegexGenerator(),
+    'datetime': UnixTimestampEncoder(),
+    'text': RegexGenerator(),
     'pii': AnonymizedFaker(),
 }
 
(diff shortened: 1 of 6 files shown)


rdt/transformers/__init__.py
```diff
@@ -87,10 +87,20 @@ def get_transformer_name(transformer):
...
+        missing_value_replacement='mean',
+        missing_value_generation='random',
+    ),
+    'categorical': LabelEncoder(add_noise=True),
+    'boolean': LabelEncoder(add_noise=True),
+    'datetime': UnixTimestampEncoder(
+        missing_value_replacement='mean',
+        missing_value_generation='random',
+    ),
+    'id': RegexGenerator(),
```
amontanez24, line 102: I don't think we have the id sdtype in RDT. The `RegexGenerator` supports `text`, so this would crash. We can just excluse it or set the default for text instead here

rdt/transformers/__init__.py
```diff
@@ -87,10 +87,20 @@ def get_transformer_name(transformer):
 
 
 DEFAULT_TRANSFORMERS = {
-    'numerical': FloatFormatter(),
-    'categorical': FrequencyEncoder(),
-    'boolean': BinaryEncoder(),
-    'datetime': UnixTimestampEncoder(),
+    'numerical': FloatFormatter(
+        learn_rounding_scheme=False,
+        enforce_min_max_values=False,
```
amontanez24, line 92: I think these are the defaults so you might not need to specify

rdt/transformers/__init__.py
R-Palazzo: I see alright, done in [1df78ab](https://github.com/sdv-dev/RDT/pull/670/commits/1df78abe7f5a8222dceea2bf2e9ea9bf9520f3a1)

pvk-developer (approved):

amontanez24 (approved):
:shipit:

PR #680, opened 2023-08-08
Create IDGenerator transformer

### LINKED ISSUE #675: Create IDGenerator transformer
### Problem Description

_As a user, I'd like to be able to generate id columns that don't all start at 0 and follow the same pattern so that my data could be more realistic. It would also be helpful to be able to customize the ids a little._

### Expected behavior

- Add `IDGenerator` transformer.

R-Palazzo:
Resolve #675


tests/integration/test_transformers.py
```diff
@@ -67,7 +67,9 @@ def _validate_helper(validator_function, args, steps):
 
 def _is_valid_transformer(transformer_name):
     """Determine if transformer should be tested or not."""
-    invalid_names = ['IdentityTransformer', 'Dummy', 'OrderedLabelEncoder', 'CustomLabelEncoder']
+    invalid_names = [
```
R-Palazzo, line 70: I don't know if we want this. I think if yes, we need an appropriate dataset with ID columns.

amontanez24: How does the `AnonymizedFaker` get tested? Wouldn't that have similar data?

rdt/transformers/id.py
```diff
@@ -0,0 +1,62 @@
...
+        self.starting_value = starting_value
+        self.suffix = suffix
+        self._counter = 0
+        self.output_properties = {None: {'next_transformer': None}}
+
+    def reset_sampling(self):
+        """Reset the sampling _counter."""
+        self._counter = 0
+
+    def _fit(self, data):
```
amontanez24, line 38: should we learn the dtype of the data and cast it back if possible? Like if the data is ints, should we output ints @npatki

frances-h (changes requested):
We also need to set the `IS_GENERATOR=True` attribute, similar to how we do with `AnonymizedFaker`

rdt/transformers/id.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __init__(self, prefix=None, starting_value=0, suffix=None):
+        super().__init__()
+        self.prefix = prefix
+        self.starting_value = starting_value
+        self.suffix = suffix
+        self._counter = 0
+        self.output_properties = {None: {'next_transformer': None}}
+
+    def reset_sampling(self):
```
frances-h, line 34: Should this be `reset_randomization` instead? Otherwise it won't get called automatically on re-fit.

tests/integration/test_transformers.py
R-Palazzo: It does not here because it has `INPUT_SDTYPE='pii'`. Following the discussion in the `EngMeeting` to not support `id sdtype` on `RDT`, should I set `INPUT_SDTYPE='pii'` to the IDGenerator rather than `INPUT_SDTYPE='id'` (as mentioned in the issue) @amontanez24?

rdt/transformers/id.py
R-Palazzo: Yes good catch thanks! Done in [33d4deb](https://github.com/sdv-dev/RDT/pull/680/commits/33d4deb7971abfdb186744e517981c2399f88d43)

R-Palazzo:
Thanks for your review @frances-h. I addressed the comments in [33d4deb](https://github.com/sdv-dev/RDT/pull/680/commits/33d4deb7971abfdb186744e517981c2399f88d43). I'm just not sure which `INPUT_SDTYPE` we should set for this transformer (`id` or `pii`)

rdt/transformers/pii/anonymizer.py
```diff
@@ -132,7 +132,9 @@ def get_supported_sdtypes(cls):
             list:
                 Accepted input sdtypes of the transformer.
         """
-        unsupported_sdtypes = {'numerical', 'datetime', 'categorical', 'boolean', 'text', None}
+        unsupported_sdtypes = {
+            'numerical', 'datetime', 'categorical', 'boolean', 'text', None, 'id'
```
frances-h, line 136: Do we need to add `id` here? My understanding from the discussion was that `RDT` doesn't support the `id` sdtype.

frances-h (commented):
One last question, but otherwise looking good!

rdt/transformers/pii/anonymizer.py
R-Palazzo: Yes good catch thank you, I forgot about it, done in [f3638cc](https://github.com/sdv-dev/RDT/pull/680/commits/f3638cc1b93f24fd7450f2ddd8fa85677ad1fc0a)

frances-h (approved):
Thanks for addressing!

amontanez24 (approved):
LGTM!

PR #681, opened 2023-08-10
Add UniformEncoder (and its ordered version)

### LINKED ISSUE #678: Add UniformEncoder (and its ordered version)
### Problem Description
The [UniformEncoder](https://docs.sdv.dev/rdt/transformers-glossary/categorical/uniformencoder) has been proven to yield better data quality than our existing FrequencyEncoder and LabelEncoder methods.

- **Marginal distributions**: Drastically improve for categorical columns, especially those with rare categories.
- **Modeling**: We have shown that after transforming a categorical column into a uniform, numerical column, the Beta distribution is able to correctly fit the resulting shape.
- **Correlations**: Generally remain unaffected.

Given the importance of categorical variables to our users, we should release these methods to the public.

### Expected behavior

Make both the [UniformEncoder](https://docs.sdv.dev/rdt/transformers-glossary/categorical/uniformencoder) and [OrderedUniformEncoder](https://docs.sdv.dev/rdt/transformers-glossary/categorical/ordereduniformencoder) available to the public by adding them to the public RDT library.

R-Palazzo:
Resolve #678
Compared to the other version, a few changes were necessary in order to get 100% coverage and make the minimum version works.
In this PR, the `UniformEncoder` is set to be the default transformer for `categorical` and `boolean` data.

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -264,7 +264,10 @@ class OrderedUniformEncoder(UniformEncoder):
         else:
             freq = data.value_counts(normalize=True, dropna=False)
 
+        nan_value = freq[np.nan] if np.nan in freq.index else None
         freq = freq.reindex(self.order).array
+        freq[np.isnan(freq)] = nan_value
+
         self.frequencies, self.intervals = self._compute_frequencies_intervals(self.order, freq)
 
     def _transform(self, data):
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -163,7 +163,7 @@ class UniformEncoder(BaseTransformer):
         def map_labels(label):
             return np.random.uniform(self.intervals[label][0], self.intervals[label][1])
 
-        return data_with_none.map(map_labels)
+        return data_with_none.map(map_labels).astype(float)
 
     def _reverse_transform(self, data):
         """Convert float values back to the original categorical values.
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -126,7 +126,11 @@ class UniformEncoder(BaseTransformer):
         data = fill_nan_with_none(data)
         labels = pd.unique(data)
         labels = self._order_categories(labels)
-        freq = data.value_counts(normalize=True, dropna=False).reindex(labels).array
+        freq = data.value_counts(normalize=True, dropna=False)
+        nan_value = freq[np.nan] if np.nan in freq.index else None
+        freq = freq.reindex(labels).array
+        freq[np.isnan(freq)] = nan_value
+
         self.frequencies, self.intervals = self._compute_frequencies_intervals(labels, freq)
 
     def _transform(self, data):
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -1,5 +1,6 @@
 """Transformers for categorical data."""
 
+import logging
 import warnings
 
 import numpy as np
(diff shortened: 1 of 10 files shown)


tests/performance/test_performance.py
```diff
@@ -8,10 +8,13 @@
...
 from rdt.performance.profiling import profile_transformer
 from rdt.transformers import get_transformers_by_type
-from rdt.transformers.categorical import CustomLabelEncoder, OrderedLabelEncoder
+from rdt.transformers.categorical import (
+    CustomLabelEncoder, OrderedLabelEncoder, OrderedUniformEncoder)
 from rdt.transformers.numerical import ClusterBasedNormalizer
 
-SANDBOX_TRANSFORMERS = [ClusterBasedNormalizer, OrderedLabelEncoder, CustomLabelEncoder]
+SANDBOX_TRANSFORMERS = [
+    ClusterBasedNormalizer, OrderedLabelEncoder, CustomLabelEncoder, OrderedUniformEncoder
```
fealho, line 16: Why does `OrderedUniformEncoder` need to be sandboxed?

tests/unit/transformers/test_categorical.py
```diff
@@ -458,16 +458,20 @@ def test__transform_error(self):
         If the data being transformed is not in ``self.order`` an error should be raised.
         """
         # Setup
-        data = pd.Series([1, 2, 3, 2, 1, 4])
+        data_error = pd.Series([1, 2, 3, 2, 1, 4])
+        data = pd.Series([1, 2, 1, 2, 1, 1])
         transformer = OrderedUniformEncoder(order=[2, 1])
 
         # Run / Assert
+        transformer._fit(data)
```
fealho, line 454: I think it makes sense to move the _fit/_transform into their own test and leave this one as it was, otherwise this one gets confusing.

tests/unit/transformers/test_categorical.py
```diff
@@ -359,38 +359,26 @@ def test___repr___default(self):
...
-            - a pandas series.
-        Side effects:
-            - set the ``values_to_categories`` dictionary to the appropriate value.
-            - set ``categories_to_values`` dictionary to the appropriate value.
-        """
+        """Test the ``_fit`` method."""
         # Setup
-        data = pd.Series([1, 2, 3, 2, np.nan, 1, 1])
-        transformer = OrderedUniformEncoder(order=[2, 3, np.nan, 1])
+        data = pd.Series([1, 2, 3, 2, None, 1, 1])
```
fealho, line 364: Why did you change this to None instead of np.nan?

rdt/transformers/categorical.py
```diff
@@ -264,7 +264,10 @@ def _fit(self, data):
         else:
             freq = data.value_counts(normalize=True, dropna=False)
 
+        nan_value = freq[np.nan] if np.nan in freq.index else None
```
fealho, line 271: I'm not sure this works with other types of nans, like float('nan')

fealho (commented):
Looking good, a few comments/questions.

rdt/transformers/categorical.py
R-Palazzo: Maybe but this should not happen because freq is defined by `data.value_counts()`, no?

tests/unit/transformers/test_categorical.py
R-Palazzo: Yes that was not necessary at the end haha good catch, reverted in [df0ad68](https://github.com/sdv-dev/RDT/pull/681/commits/df0ad68efa2fd0b78ff118934e25ef03f8a3a926)

tests/unit/transformers/test_categorical.py
R-Palazzo: Yes I agree, reverted the change in [df0ad68](https://github.com/sdv-dev/RDT/pull/681/commits/df0ad68efa2fd0b78ff118934e25ef03f8a3a926).

R-Palazzo:
Thanks for your review @fealho ;)

rdt/transformers/categorical.py
fealho: I was concerned with the freq.index having nans in there which are not np.nan. I'm not sure if it's impossible for that to happen.

rdt/transformers/categorical.py
```diff
@@ -264,7 +264,10 @@ def _fit(self, data):
...
         freq = freq.reindex(self.order).array
```
amontanez24, line 272: When this reindexes, won't add None to the index as well since None should be in `self.order`?

R-Palazzo: No because if there is `None` or `np.nan` in the data,  `data.value_counts` will return only `np.NaN` as index

amontanez24: None is is self.order

fealho (approved):
Thanks for addressing 👍

PR #690, opened 2023-08-21
Deprecate get_input_sdtype

### LINKED ISSUE #682: Deprecate get_input_sdtype
### Problem Description

_As a user, I sometimes accidently use `get_input_sdtype` instead of `get_supported_sdtypes`, which can cause errors down the line._

Since transformers can support multiple input sdtypes, we added [get_supported_sdtypes](https://github.com/sdv-dev/RDT/blob/a957b36c67816b6b0e049b39c4237b0169a6c1a2/rdt/transformers/base.py#L190C9-L190C30). The problem is we still have [get_input_sdtype](https://github.com/sdv-dev/RDT/blob/a957b36c67816b6b0e049b39c4237b0169a6c1a2/rdt/transformers/base.py#L180), and it's not clear which to use.

### Expected behavior

- When calling `get_input_sdtype` a `FutureWarning` should be raised, telling the user that this attribute is deprecated and `get_supported_sdtypes` is preferred.

### Additional context

- Make sure we aren't accidently using `get_input_sdtypes` anywhere important.

R-Palazzo:
Resolve #682
I supposed that every call of `get_input_sdtype()` could be changed by `get_supported_sdtypes()[0]`, is it fine?

### DIFF
--- a/tests/unit/test_hyper_transformer.py
+++ b/tests/unit/test_hyper_transformer.py
@@ -2246,11 +2246,10 @@ class TestHyperTransformer(TestCase):
         instance = HyperTransformer()
         instance._fitted = False
         mock_transformer = Mock()
-        mock_transformer.get_supported_sdtypes.return_value = ['datetime']
         column_name_to_transformer = {
             'my_column': mock_transformer
         }
-
+        expected_config = instance.get_config()
         # Run
         expected_msg = (
             'Nothing to update. Use the `detect_initial_config` method to pre-populate '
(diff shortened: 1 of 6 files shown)


tests/unit/test_hyper_transformer.py
```diff
@@ -2246,7 +2246,7 @@ def test_update_transformers_no_field_transformers(self):
         instance = HyperTransformer()
         instance._fitted = False
         mock_transformer = Mock()
-        mock_transformer.get_supported_sdtype.return_value = ['datetime']
```
amontanez24, line 2249: It seems like this mock isn't doing anything since it was spelled wrong and the test still passed. We can probably just remove these lines and assert at the end that the config is unchanged

amontanez24 (commented):
One small comment not really even related to this PR but outside of that looks good!

amontanez24 (approved):
:shipit:

PR #692, opened 2023-08-22
Support multi-column transformers

### LINKED ISSUE #683: Support multi-column transformers
### Problem Description

_As a user, sometimes I have multiple columns in a table used to represent one combined field. It would be helpful if RDT supported this and transformed/reverse transformed these columns together._

### Expected behavior

To add support for multi -column transformers, a couple of changes need to be made. To make it work, we can create a new class called `BaseMultiColumnTransformer` that inherits from the `BaseTransformer` and makes the following changes:

1. There should be a `get_input_columns` method instead of `get_input_column`
2. The `fit` method should change the signature to be `fit(self, data, columns_to_sdtypes)`. Here, `columns_to_sdtypes` is a dict mapping the column names in the field to the sdtypes they have. This is necessary information for a multi-column transformer.
3. `get_input_column` should just raise an error since it doesn't make sense here, and point to `get_input_columns`.

#### HyperTransformer changes

The `HyperTransformer` will also need to be changed to work with tuples as a field. There is already some support for this ([see here](https://github.com/sdv-dev/RDT/blob/a957b36c67816b6b0e049b39c4237b0169a6c1a2/rdt/hyper_transformer.py#L95)), but the `fit` method still needs to be changed.

In the lines below, we loop through single columns at a time and fit the transformers.
https://github.com/sdv-dev/RDT/blob/a957b36c67816b6b0e049b39c4237b0169a6c1a2/rdt/hyper_transformer.py#L606-L607

This logic needs to change to check if the column is part of a tuple in the config (ie. part of a multi-column field). If it is, the transformer needs to be fitted on all of those columns, not just one of them.

### Additional context

- Any multi-column transformer will need to know which columns being passed to fit belong to which sdtype. There are a few ways to do this, but the proposal is to pass the sdtypes and column names to the fit of the transformer. For example, if you have the following multi-column transformer
```python
class Address(BaseMultiColumnTransformer):
    def fit(self, data, columns_to_sdtypes):
        pass
```
Then the `HyperTransformer` can get that info and pass it in
```
columns_to_sdtypes = {column: self.config['sdtypes'][column] for column in multi_column_field}
tranformer.fit(data, columns_to_sdtypes)
```

R-Palazzo:
Resolve #683


rdt/transformers/base.py
```diff
@@ -255,16 +255,25 @@ def get_output_columns(self):
...
+        """Store the column names to be transformed.
 
-        missing = set(columns) - set(data.columns)
+        Args:
+            columns (str, list, tuple):
+                Column name(s) to be transformed.
+            data (pandas.DataFrame):
+                The entire table.
+        """
+        column_names = columns
```
amontanez24, line 266: what's the point of this assignment? Why not just change the name of the variable

rdt/hyper_transformer.py
```diff
@@ -603,8 +603,8 @@ def fit(self, data):
         self._validate_detect_config_called(data)
         self._unfit()
         self._input_columns = list(data.columns)
-        for field in self._input_columns:
-            data = self._fit_field_transformer(data, field, self.field_transformers[field])
+        for field_column, field_transformer in self.field_transformers.items():
+            data = self._fit_field_transformer(data, field_column, field_transformer)
```
amontanez24, line 607: I don't think this is going to work alone. The `_fit_field_transformer` method doesn't know to pass the columns as a dict in the multi column case

R-Palazzo: In the multi column case we pass the columns as a tuple or as a dict? I supposed it was a tuple and so it was good for me based on the docstring of `_fit_field_transformer(self, data, field, transformer)`:
```
field (str or tuple):
                Name of column or tuple of columns in data that will be transformed
```

rdt/transformers/base.py
R-Palazzo: Yes first I was not sure if we really wanted to modify the column parameter, but I removed my change, it was not necessary

rdt/hyper_transformer.py
amontanez24: In the base class you made fit takes in data and a dictionary. It needs to be a dict otherwise the values will always have to be passed in a specific order

amontanez24: @R-Palazzo I think this is still the case. The [issue](https://github.com/sdv-dev/RDT/issues/683) states that the `HyperTransformer` should run with a multi-column transformer too.

tests/integration/transformers/test_base.py
```diff
@@ -129,3 +129,190 @@ def _reverse_transform(self, data):
     })
     pd.testing.assert_frame_equal(expected_transform, transformed)
     pd.testing.assert_frame_equal(reverse, data)
+
+
+def test_multi_column_transformer_same_number_of_columns_input_output():
+    """Test a multi-column transformer when the same of input and output columns."""
+    # Setup
+    class AdditionTransformer(BaseMultiColumnTransformer):
```
amontanez24, line 137: I think we should actually test this example from the `HyperTransformer` and make sure it really works end to end

rdt/transformers/base.py
```diff
@@ -479,3 +479,122 @@ def reverse_transform(self, data):
...
+        output = {}
+        for output_column, properties in self.output_properties.items():
+            # if 'sdtype' is not in the dict, ignore the column
+            if property_ not in properties:
+                continue
+
+            if is_prefix_dict:
+                prefix = self.prefixes[output_column]
+            else:
+                prefix = self.prefixes
```
amontanez24, line 543: Why is this allowing multiple types for the prefixes? I think it could be complicated if developers have to set custom prefixes and output names in 2 places. Maybe it would be simpler to say that if you want to have custom prefixes for each output name, then just use custom output names and set the prefix to None.

In other words, `_get_prefix` always returns None or a string. And then you can just do the lines below

rdt/transformers/base.py
```diff
@@ -479,3 +479,122 @@ def reverse_transform(self, data):
...
+                output[f'{prefix}.{output_column}'] = properties[property_]
+
+        return output
+
+    def _validate_columns_to_sdtypes(self, data, columns_to_sdtypes):
+        """Check that all the columns in ``columns_to_sdtypes`` are present in the data."""
+        missing = set(columns_to_sdtypes.keys()) - set(data.columns)
+        if missing:
+            missing_to_print = ', '.join(missing)
+            raise KeyError(f'Columns ({missing_to_print}) are not present in the data.')
```
amontanez24, line 557: I don't think this should be a `KeyError`. I think that's really only for if the accessed object doesn't have the key. I think maybe a `ValueError` makes sense since `columns_to_sdtypes` has the wrong values

tests/unit/transformers/test_base.py
```diff
@@ -1272,3 +1272,224 @@ def _reverse_transform(self, data):
...
+        expected_output = {
+            'col_1': 'numerical',
+            'col_2': 'categorical',
+            'col_3': 'boolean',
+        }
+        assert output == expected_output
+        transformer._get_prefix.assert_called_once_with()
+
+    def test__validate_columns_to_sdtypes(self):
+        """Test the ``_validate_columns_to_sdtypes`` method."""
```
amontanez24, line 1400: nitpick: for these descriptions I think it's more useful to say what's being tested. eg
```
Test that an error with the missing column names is raised.
```

amontanez24 (changes requested):

rdt/hyper_transformer.py
```diff
@@ -519,6 +535,18 @@ def detect_initial_config(self, data):
...
+        """Generate the ``columns_to_sdtypes`` dict for the given field.
+
+        Args:
+            field (tuple):
+                Names of the column for the multi column trnasformer.
+        """
+        columns_to_sdtypes = {}
+        for column in field:
+            columns_to_sdtypes[column] = self.field_sdtypes[column]
+        return columns_to_sdtypes
```
pvk-developer, line 548: Break line above would be appreciated.

amontanez24 (approved):
LGTM! Thanks for addressing

PR #694, opened 2023-08-24
Fix wrong mocks

fealho:


tests/unit/test_hyper_transformer.py
```diff
@@ -2302,7 +2302,6 @@ def test_update_transformers_missmatch_sdtypes(self, mock_warnings):
         with pytest.raises(InvalidConfigError, match=err_msg):
             instance.update_transformers(column_name_to_transformer)
 
-        assert mock_warnings.called_once_with(err_msg)
```
frances-h, line 2305: Should we be deleting this or can we change it to `assert_called_once_with`?

fealho: It doesn't work. It's checking if `print` is called, but there's no `print` anywhere in the file which it's mocking. I'm not sure what the intent of this was.

frances-h: Bizarre. We may as well just get rid of the patch and mock then, since they're not doing anything.

frances-h (approved):
👍 Looks good, thanks for fixing!

PR #695, opened 2023-08-24
Improve user warnings and logic for update_transformers and update_transformers_by_sdtype

### LINKED ISSUE #685: Improve user warnings and logic for update_transformers and update_transformers_by_sdtype
### Problem Description

_As a user, I'd like to be able to update the transformers to use on multiple columns at once if they are part of the same concept (eg. an address.) When doing this, the config should update correctly._

#### update_transformers

- A column cannot be assigned to multiple transformers at once. If you try to assign a new transformer to a column, remove the column from any multi-column transformers that it's already a part of.
```python
# before
>>> ht.get_config()
{
  'sdtypes': {
    'A': 'city',
    'B': 'state',
    'C': 'country'
  },
  'transformers': {
    ('A', 'B', 'C'): RandomLocationGenerator()
  }
}

# updating one of the columns to AnonymizedFaker instead
>>> ht.update_transformers(column_name_to_transformer={
         'A': AnonymizedFaker(provider_name='address', function_name='city')
})
>>> ht.get_config()
{
  'sdtypes': {
    'A': 'city',
    'B': 'state',
    'C': 'country'
  },
  'transformers': {
    'A': AnonymizedFaker(provider_name='address', function_name='city'),
    ('B', 'C'): RandomLocationGenerator()
  }
}

# Edge case: if there is only 1 transformer, it is entirely replaced
>>> ht.get_config()
{
  'sdtypes': {
    'D': 'city'
  },
  'transformers': {
    ('D'): RandomLocationGenerator()
  }
}
>>> ht.update_transformers(column_name_to_transformer={
         'D': AnonymizedFaker()
})
>>> ht.get_config()
{
  'sdtypes': {
    'D': 'city'
  },
  'transformers': {
    'D': AnonymizedFaker()
  }
}
```
- If you assign a multi-column transformer, the other transformers assigned to those columns should be removed.

```python
# before
>>> ht.get_config()
{
  'sdtypes': {
    'A': 'city',
    'B': 'state',
    'C': 'country'
  },
  'transformers': {
    'A': AnonymizedFaker(),
    'B': AnonymizedFaker(),
    'C': AnonymizedFaker(),
  }
}

# adding a multi-column transformer
>>> ht.update_transformers(column_name_to_transformer={
         ('A', 'B', 'C'): RandomLocationGenerator()
})
>>> ht.get_config()
{
  'sdtypes': {
    'A': 'city',
    'B': 'state',
    'C': 'country'
  },
  'transformers': {
    ('A', 'B', 'C'): RandomLocationGenerator()
  }
}
```

#### update_transformers_by_sdtype
- This works similarly to `update_transformers()` but for all matching columns
```python
# before
>>> ht.get_config()
{
  'sdtypes': {
    'A': 'city',
    'B': 'state',
    'C': 'country'
  },
  'transformers': {
    ('A', 'B', 'C'): RandomLocationGenerator()
  }
}
>>> ht.update_transformers_by_sdtype(
      sdtype='city', transformer='AnonymizedFaker')
>>> ht.get_config()
{
  'sdtypes': {
    'A': 'city',
    'B': 'state',
    'C': 'country'
  },
  'transformers': {
    'A': AnonymizedFaker(),
    ('B', 'C'): RandomLocationGenerator()
  }
}
```

R-Palazzo:
Resolve #685
I’m not using the `BaseMultiColumnTransformer` because it’s crashing on some validation. It’s mainly because this transformer does not have a `supported_sdtype`. So to check what I’m doing, I’m using the option to assign to `None` the transformers. Let me know what we should do about this

Also, about the `update_transformers_by_sdtype` I made the change according to what is written in the issue; there is one case I’m not sure about. For me, when we update the sdtype we should not touch all the columns that have this sdtype but that are used in a `Multi Column transformer`. In this PR I do it, the columns are removed to the tuple and are being assigned to the new transformer.

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -338,7 +338,7 @@ class HyperTransformer:
     def _remove_column_in_multi_column_fields(self, column):
         """Remove a column that is part of a multi-column field.
 
-        Update the tuple to remove the column from it and modify the ``multi_column_fields``
+        Remove the column from the tuple and modify the ``multi_column_fields``
         as well as the ``field_transformers`` dicts accordingly.
 
         Args:


tests/integration/test_hyper_transformer.py
```diff
@@ -1370,3 +1371,636 @@ def test_random_seed(self):
...
+    # Run - make sure resetting randomization works
+    ht.reset_randomization()
+    transformed = ht.transform(data)
+    reverse_transformed3 = ht.reverse_transform(transformed)
+
+    # Assert
+    pd.testing.assert_frame_equal(reverse_transformed1, reverse_transformed3)
+
+
+def test_hypertransformer_pseudo_anonymized_faker():
```
amontanez24, line 1780: I think some of these tests are duplicates. You may have to fix the rebase

rdt/hyper_transformer.py
```diff
@@ -228,7 +246,10 @@ def _validate_config(config):
             )
 
     def _validate_update_columns(self, update_columns):
-        unknown_columns = self._subset(update_columns, self.field_sdtypes.keys(), not_in=True)
+
```
amontanez24, line 249: I don't think we need an extra line

rdt/hyper_transformer.py
```diff
@@ -311,6 +332,25 @@ def _warn_update_transformers_by_sdtype(self, transformer, transformer_name):
                 'parameters instead.', FutureWarning
             )
 
+    def _remove_column_in_multi_column_fields(self, column):
+        """Update the column in tuple dict.
```
amontanez24, line 336: Can we give a more in depth description here. It seems like this method remove on column from a field that is made up of multiple columns by removing it from the tuple. It also updates the transformer

rdt/hyper_transformer.py
```diff
@@ -205,7 +211,19 @@ def _validate_config(config):
 
         sdtypes = config['sdtypes']
         transformers = config['transformers']
-        if set(sdtypes.keys()) != set(transformers.keys()):
+
+        sdtype_keys = flatten_column_list(sdtypes.keys())
```
amontanez24, line 215: I don't think this will ever have tuples as keys. Only the transformers

rdt/hyper_transformer.py
```diff
@@ -335,6 +375,8 @@ def update_transformers_by_sdtype(
         self._warn_update_transformers_by_sdtype(transformer, transformer_name)
 
         transformer_instance = transformer
+        self._multi_column_fields = self._create_multi_column_fields()
```
amontanez24, line 378: we should just remove this from the init and move it into the detect method. Then we don't need to keep calling it

tests/integration/test_hyper_transformer.py
R-Palazzo: Yes you're right, fixed in [e8d956c](https://github.com/sdv-dev/RDT/pull/695/commits/e8d956c4db8162a28e57047d74225fe5fa3b168c)

rdt/hyper_transformer.py
R-Palazzo: Yes done in [f01ceca](https://github.com/sdv-dev/RDT/pull/695/commits/f01ceca786505c287f030d20b7e58710deef127f)

rdt/hyper_transformer.py
```diff
@@ -23,9 +24,14 @@ class Config(dict):
 
     def __repr__(self):
         """Pretty print the dictionary."""
+        transformers_repr = {}
```
fealho, line 27: I honestly prefer the one-liner implementation, I find it more concise and faster to read. It doesn't matter much though, so you can use whichever way you prefer.

rdt/hyper_transformer.py
```diff
@@ -205,7 +211,19 @@ def _validate_config(config):
 
         sdtypes = config['sdtypes']
         transformers = config['transformers']
-        if set(sdtypes.keys()) != set(transformers.keys()):
+
+        sdtype_keys = sdtypes.keys()
```
fealho, line 215: You can't have a dict with repeated keys, so I don't think the `is_sdtype_keys_unique` check if necessary.

tests/integration/test_hyper_transformer.py
```diff
@@ -1370,3 +1391,138 @@ def test_random_seed(self):
 
         # Assert
         pd.testing.assert_frame_equal(reversed1, reversed2)
+
+    def test_update_transformers_single_to_multi_column(self):
+        """Test ``update_transformers`` to go from single to mutli column transformer."""
+
```
fealho, line 1397: extra line

tests/unit/test_hyper_transformer.py
```diff
@@ -2181,6 +2309,71 @@ def test_update_transformers_fitted(self, mock_warnings):
...
+        # Run
+        ht.update_transformers(column_name_to_transformer)
+
+        # Assert
+        expected_field_transformers = {
+            'A': UniformEncoder(),
+            'B': None,
+            'C': FloatFormatter(),
+        }
+        ht.field_transformers == expected_field_transformers
```
fealho, line 2375: assert

tests/unit/test_hyper_transformer.py
```diff
@@ -2181,6 +2309,71 @@ def test_update_transformers_fitted(self, mock_warnings):
...
+        ht.field_transformers = {
+            ('A', 'B'): None,
+            'C': FloatFormatter(),
+        }
+
+        column_name_to_transformer = {
+            'A': UniformEncoder(),
+        }
+        # Run
+        ht.update_transformers(column_name_to_transformer)
```
fealho, line 2367: assert

tests/unit/test_hyper_transformer.py
```diff
@@ -2181,6 +2309,71 @@ def test_update_transformers_fitted(self, mock_warnings):
...
+
+        # Assert
+        expected_field_transformers = {
+            ('A', 'B'): None,
+            'C': None,
+        }
+        ht.field_transformers == expected_field_transformers
```
fealho, line 2341: assert

fealho (commented):
Looks pretty good, just had a couple of comments

rdt/hyper_transformer.py
R-Palazzo: Yes me too but the one-liner implementation was crashing for tuples as key that's why I changed it. I got this error:
`TypeError: keys must be str, int, float, bool or None, not tuple`

fealho (approved):

rdt/hyper_transformer.py
```diff
@@ -541,7 +609,12 @@ def _fit_field_transformer(self, data, field, transformer):
             self._output_columns.append(field)
 
         else:
-            transformer.fit(data, field)
+            if isinstance(field, tuple):
+                columns_to_sdtypes = self._get_columns_to_sdtypes(field)
+                transformer.fit(data, columns_to_sdtypes)
+            else:
+                transformer.fit(data, field)
```
amontanez24, line 616: I think this should be moved to the base PR

amontanez24 (commented):
This looks pretty good! I do think we should move the logic to make the `HyperTransformer` work with a multi-column transformer to the other PR though. It makes more sense with that issue

rdt/hyper_transformer.py
```diff
@@ -311,6 +335,28 @@ def _warn_update_transformers_by_sdtype(self, transformer, transformer_name):
...
+
+        Args:
+            column (str):
+                Column name to be updated.
+        """
+        old_tuple = self._multi_column_fields.pop(column)
+        new_tuple = tuple(item for item in old_tuple if item != column)
+
+        if len(new_tuple) == 1:
+            new_tuple, = new_tuple
```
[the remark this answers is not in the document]
R-Palazzo: It's to not have a single element tuple in the` field_transformer` dict. Without this line I have a test failing with the error:
`AssertionError: assert {('column2',): 'transformer'} == {'column2': 'transformer'}`
The right is the expected dict

PR #702, opened 2023-09-01
Resolve locales warning for providers that end with the locale name

### LINKED ISSUE #701: Resolve locales warning for specific sdtype/locale combos (eg. en_US with postcode)
As described in the issue https://github.com/sdv-dev/SDV/issues/1480

The error occurs because of the given provider name detected by `sdv`. In that case, when using a locale, it tries to get the most precise provider, in the case of it ending with `en_US` which causes the logic in `_check_locales` to raise a warning:
https://github.com/sdv-dev/RDT/blob/5d7f8b7f57c4be044f046a36c28dec243734cb9b/rdt/transformers/pii/anonymizer.py#L78-L94

pvk-developer:
Resolves #701 
Resolves https://github.com/sdv-dev/SDV/issues/1480

### DIFF
--- a/rdt/transformers/pii/anonymizer.py
+++ b/rdt/transformers/pii/anonymizer.py
@@ -81,7 +81,7 @@ class AnonymizedFaker(BaseTransformer):
         missed_locales = []
         for locale in locales:
             spec = importlib.util.find_spec(f'faker.providers.{self.provider_name}.{locale}')
-            if spec is None:
+            if spec is None and not self.provider_name.endswith(locale):
                 missed_locales.append(locale)
 
         if missed_locales:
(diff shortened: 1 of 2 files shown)


frances-h (approved):
Looks good 👍

rdt/transformers/pii/anonymizer.py
```diff
@@ -81,7 +81,7 @@ def _check_locales(self):
         missed_locales = []
         for locale in locales:
             spec = importlib.util.find_spec(f'faker.providers.{self.provider_name}.{locale}')
-            if spec is None:
+            if spec is None and not self.provider_name.endswith(locale):
```
amontanez24, line 84: Is it safe to assume that if the provider name ends with the locale, then it exists? Is it possible it's just a bad provider name that ends with a locale

pvk-developer: I think that you are right, we can improve this by adding `.`. to the locale when looking for.

amontanez24: I meant more that I think we should make sure the provider exists still. Like if the locale is already present, extract it and runs it through `importlib.util.find_spec(f'faker.providers....`

amontanez24 (approved):
Thanks for addressing!

PR #704, opened 2023-09-06
`ClusterBasedNormalizer` code cleanup

### LINKED ISSUE #696: `ClusterBasedNormalizer` code cleanup
### Problem Description

_The current code can be confusing at times and it has a few artifacts which don't have any purpose. We should refactor it to improve readability._

### Expected behavior
The following fixes should be implemented:
- Clarify inputs and outputs in the docstring
- Describe purpose of `STD_MULTIPLIER`
- Remove `n_init` parameter (it's the default in sklearn)
- Either explain what warnings.simplefilter('ignore') does or remove it if it's unnecessary
- Switch the order of operations during the transformation of the normalize values. Select the value first and normalize them after.
- Remove both noqa's
- Change astype to round and clip in the reverse_transform
- Remove else at the end of reverse_transform, just set `data=recovered_data` and update it in the if statement

fealho:
Resolve #696.

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -463,7 +463,6 @@ class ClusterBasedNormalizer(FloatFormatter):
             n_components=self.max_clusters,
             weight_concentration_prior_type='dirichlet_process',
             weight_concentration_prior=0.001,
-            n_init=1,
             random_state=self._get_current_random_seed()
         )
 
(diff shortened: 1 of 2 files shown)


fealho:
Note: the lines ` with warnings.catch_warnings(): warnings.simplefilter('ignore')` don't seem to do anything, ie I didn't manage to raise any warnings after removing those lines. However, since the original PR never gave an example where that warning was raised, it's possible that the examples I ran just happened to not raise the warning. I would have to investigate further to conclude that the warning will indeed never be raised, but that seems out of scope for this PR, so I'm leaving them as is.

rdt/transformers/numerical.py
```diff
@@ -546,8 +548,6 @@ def _reverse_transform(self, data):
 
         recovered_data = self._reverse_transform_helper(data)
         if self.null_transformer and self.null_transformer.models_missing_values():
-            data = np.stack([recovered_data, data[:, -1]], axis=1)  # noqa: PD013
-        else:
-            data = recovered_data
+            recovered_data = np.stack([recovered_data, data[:, -1]], axis=1)  # noqa: PD013
```
amontanez24, line 551: Can we get rid of this noqa?

fealho: Nope. It thinks that is pandas `stack` and tells you to use `melt`, which is a method numpy doesn't have.

pvk-developer: Is `np.concatenate` the same ?

pvk-developer (approved):
LGTM! :shipit:

rdt/transformers/numerical.py
fealho: `stack` concatenates them along a different axis. It's also supposed to be faster. I think we could change it to `np.concatenate([[array1], [array2]], axis=1)`, or something like that, but that just seemed more confusing than the `noqa`.

amontanez24 (approved):
:shipit:

PR #706, opened 2023-09-06
Add `uniform` to `GaussianNormalizer` distributions

### LINKED ISSUE #655: The `GaussianNormalizer` should accept `uniform` distributions
### Problem Description
The [GaussianNormalizer](https://docs.sdv.dev/rdt/transformers-glossary/numerical/gaussiannormalizer) accepts a `distribution` parameter, where you can provide a string name of the univariate distribution. For example:

```python
my_transformer = GaussianNormalizer(distribution='beta')
```

The available distributions are: `'gaussian'`, `'gamma'`, `'beta'`, `'student_t'`, `'truncated_gaussian'` and `'gaussian_kde'`

### Expected behavior
We should also accept `'uniform'` as a possible parameter. When selected, this should use the copulas [UniformUnivariate distribution](https://github.com/sdv-dev/Copulas/blob/master/copulas/univariate/uniform.py#L9).

### Additional context
If you're only using RDT, then instead of supplying the distribution as a string name, you can also directly input a class. This also does not work for uniform distributions.

```python
from copulas import univariate

my_transformer = GaussianNormalizer(distribution=univariate.Uniform)
```

fealho:
Resolve #655.

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -252,9 +252,10 @@ class GaussianNormalizer(FloatFormatter):
                 * ``gamma``: Use a Gamma distribution.
                 * ``beta``: Use a Beta distribution.
                 * ``student_t``: Use a Student T distribution.
-                * ``gussian_kde``: Use a GaussianKDE distribution. This model is non-parametric,
+                * ``gaussian_kde``: Use a GaussianKDE distribution. This model is non-parametric,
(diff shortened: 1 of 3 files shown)


pvk-developer (approved):
👍🏻 LGTM! Also great that you sorted the `__init__`!

amontanez24 (approved):
:shipit:

Issue #180, opened 2021-07-13
Audit CategoricalTransformer


katxiao:
Audit the time and memory performance of `transformers.CategoricalTransformer` 
https://github.com/sdv-dev/RDT/blob/d756679df61e6ab5330bf6132a1c155c2003f1ef/rdt/transformers/categorical.py#L10

katxiao:
For `fit`, about 99% of the runtime is spent on the lines getting and sorting the value counts of the data: 
https://github.com/sdv-dev/RDT/blob/d756679df61e6ab5330bf6132a1c155c2003f1ef/rdt/transformers/categorical.py#L65-L70

In `transform`, ~99% of the runtime is spent on iterating through the data and applying the `_get_value` method: 
https://github.com/sdv-dev/RDT/blob/d756679df61e6ab5330bf6132a1c155c2003f1ef/rdt/transformers/categorical.py#L135

In `reverse_transform`, ~92% of the runtime is spent on this line to set the value back to the original category:
https://github.com/sdv-dev/RDT/blob/d756679df61e6ab5330bf6132a1c155c2003f1ef/rdt/transformers/categorical.py#L169

katxiao:
Proposed changes to `fit`:
In the original implementation, we convert the series to a dataframe and add the index as a column in order to sort by [name, index]. After sorting, we move the index back and drop the index column. Instead, we could use `np.lexsort` to get the desired ordering.
```
frequencies = data.value_counts(dropna=False)
frequencies = frequencies.iloc[np.lexsort([frequencies.values, frequencies.index])[::-1]]
```
After this change, the sorting of value counts takes ~46.5% of the runtime.

Before:
```
CPU times: user 8.71 ms, sys: 910 µs, total: 9.62 ms
Wall time: 10.8 ms
```
After:
```
CPU times: user 4.21 ms, sys: 920 µs, total: 5.13 ms
Wall time: 4.93 ms
```

katxiao:
Proposed changes to `transform`:

In the original implementation, we use the `apply` method which iterates over all the rows in the dataframe to transform the categories. If there are many rows, this could become inefficient. In the case that there are many rows and few categories, we could use the following approach, which iterates over categories instead:
```
        num_rows = len(data)
        num_categories = len(self.means)
        if <appropriate ratio of num_rows/num_categories is met>:
            result = np.empty(shape=(len(data), ), dtype=float)

            # loop over categories
            for category, values in self.intervals.items():
                mean = values[2]
[...]

PR #711, opened 2023-09-14
Remove psutil dependency and deprecate FrequencyEncoder

### LINKED ISSUE #614: Consider deprecating the `FrequencyEncoder`
### Problem Description
The `FrequencyEncoder` takes up more time than `LabelEncoder`. Furthermore, we have evidence that `LabelEncoder` produces data that is better quality for data science purposes.

Considering this, there is not much benefit to keeping `FrequencyEncoder` around in our library. 

### Expected behavior
Issue a `FutureWarning` whenever a `FrequencyEncoder` object is being created. 

```
The 'FrequencyEncoder' transformer will no longer be supported in future versions of the RDT library.
Please use the 'UniformEncoder' transformer instead.
```

Additionally, we should **apply the UniformEncoder** as a default for categorical columns in the `detect_initial_config` method.

### LINKED ISSUE #615: Remove `psutil`
### Description
Currently, the `FrequencyEncoder` uses a library called `psutil` to determine the amount of memory available on the machine and optimize the transform. See [line of code](https://github.com/sdv-dev/RDT/blob/e4fb0e5960feb8e829c8d8f5f64abdade8f8eadf/rdt/transformers/categorical.py#L259).

1. We should remove `psutil` as a dependency, as we want to get rid of unnecessary, hyper-specific dependencies
2. In `FrequencyEncoder`, we should remove the if/then/else statement related to memory management and pick one of the approaches (eg. `_reverse_transform_by_matrix`)

### Additional context
In the future, we are considering deprecating `FrequencyEncoder`. See #614 for more details.

rwedge:
Resolves #614 and resolves #615 

#615 suggests using a single reverse transform code path but based on this [comment](https://github.com/sdv-dev/RDT/issues/180#issuecomment-881198196) switching between row or column based could be useful so I left both in.


tests/unit/transformers/test_categorical.py
```diff
@@ -700,44 +705,6 @@ def test__get_value_add_noise_true(self, norm_mock):
         # Asserts
         assert result == 0.2745
 
-    def test__reverse_transform_array(self):
```
rwedge, line 703: `BaseTransformer.reverse_transform` specifies the input data should be a pandas DataFrame so I removed this test on a numpy array.

setup.py
```diff
@@ -42,6 +41,7 @@
     'jupyter>=1.0.0,<2',
     'rundoc>=0.4.3,<0.5',
     'pytest-subtests>=0.5,<1.0',
+    'psutil>=5.7,<6',
```
fealho, line 44: Why do we need this? If we are removing the only place in the code where `psutil` is used, shouldn't we also be able to remove it from every test where it is mocked?

fealho (changes requested):
Looks good so far. I think you also need to add the `UniformEncoder` to the `detect_initial_config` as specified in #614.

setup.py
rwedge: Yes we can revise those test cases

rwedge:
> Looks good so far. I think you also need to add the `UniformEncoder` to the `detect_initial_config` as specified in #614.

`UniformEncoder` is already the categorical encoder for `rdt.transformers.DEFAULT_TRANSFORMERS` which I believe is where `detect_initial_config` ends up getting its default transformers from

tests/unit/transformers/test_categorical.py
fealho: Maybe you should change the data to be a pd.Series instead? Because this is the only test we have for the `_reverse_transform` in the FrequencyEncoder, so the logic may end up not being tested if you remove it.

fealho (approved):
I just left one comment, after that it looks good to go.

R-Palazzo (approved):
LGTM! ;)

PR #714, opened 2023-09-21
Keep the same column order when fitting the hypertransformer

R-Palazzo:
Ensure that we keep the same order of columns when fitting the hypertransformer compared to before the multi-column transformer

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -691,8 +691,19 @@ class HyperTransformer:
         self._validate_detect_config_called(data)
         self._unfit()
         self._input_columns = list(data.columns)
-        for field_column, field_transformer in self.field_transformers.items():
-            data = self._fit_field_transformer(data, field_column, field_transformer)
+        skipped_columns = []  # skip columns in multi column transformer already fitted
+        for column in self._input_columns:
+            if column in skipped_columns:
+                continue
+
+            if column in self._multi_column_fields:
+                field = self._multi_column_fields[column]
+                field_to_skip = [col for col in field if col != column]
+                skipped_columns.extend(field_to_skip)
+            else:
+                field = column
+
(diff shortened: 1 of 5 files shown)


rdt/transformers/__init__.py
```diff
@@ -100,6 +100,8 @@ def get_transformer_name(transformer):
     'pii': AnonymizedFaker(),
 }
 
+MULTI_COLUMN_TRANSFORMERS = {}
```
amontanez24, line 103: How will this be different than the current dictionary?

amontanez24: Right now this dictionary is meant to map an argument (the anonymization level) to the address transformer to use. I think we should just call it `ADDRESS_TRANSFORMERS`. That's more clear and there's no guarantee other multi column transformers will have defaults set by a parameter anyway. If they do we can update

rdt/hyper_transformer.py
```diff
@@ -691,8 +691,19 @@ def fit(self, data):
         self._validate_detect_config_called(data)
         self._unfit()
         self._input_columns = list(data.columns)
-        for field_column, field_transformer in self.field_transformers.items():
-            data = self._fit_field_transformer(data, field_column, field_transformer)
+        skipped_columns = []  # skip columns in multi column transformer already fitted
+        for column in self._input_columns:
```
amontanez24, line 695: any reason to loop over this instead of the `field_transformers`?

amontanez24 (approved):
Just one questions but looks good!

rdt/hyper_transformer.py
R-Palazzo: Before defining the `multi-column transformers` we were iterating over `self._input_columns`. I found that this order can be important sometimes for the constraint, and that's why some tests were failing. By iterating over field_transformers, some randomness in the order makes some tests fail sometimes.

pvk-developer (approved):

PR #715, opened 2023-09-28
Replace missing values with random values from dataset

### LINKED ISSUE #606: Replace missing values with variable (random) values from the dataset
### Problem Description

Many of the numerical transformers allow you to replace missing values with a static value such as the `'mean'` or `'mode'` (See [FloatFormatter API](https://docs.sdv.dev/rdt/transformers-glossary/browse-transformers/floatformatter)). While this gets rid of missing values, it comes at the cost of _changing the distribution shape_.

Eg. If I have many missing values and I select the `'mean'` strategy, then the resulting distribution has a large peak only at the `'mean'`.

### Expected behavior

To avoid this problem, we can consider adding a `'random'` strategy for `missing_value_replacement`: The missing values will be filled in by choosing other non-missing values completely at random from the uniform `[min, max]` interval.

As a result, the transformed data will 
(a) not contain any missing values and
(b) create a similar overall shape (marginal distribution) as before

**We should make this the default handling strategy**

### Additional context

This can apply to any transformer that has the `missing_value_replacement` parameter: BinaryEncoder, FloatFormatter, OptimizedTimestampEncoder and UnixTimestampEncoder.

pvk-developer:
Resolves #606

### DIFF
--- a/rdt/transformers/null.py
+++ b/rdt/transformers/null.py
@@ -73,7 +73,7 @@ class NullTransformer():
         if self._missing_value_replacement is None:
             return None
 
-        if self._missing_value_replacement in {'mean', 'mode'} and pd.isna(data).all():
+        if self._missing_value_replacement in {'mean', 'mode', 'random'} and pd.isna(data).all():
             msg = (
                 f"'missing_value_replacement' cannot be set to '{self._missing_value_replacement}'"
                 ' when the provided data only contains NaNs. Using 0 instead.'
@@ -114,6 +114,10 @@ class NullTransformer():
             if self._missing_value_generation == 'random':
                 self._null_percentage = null_values.sum() / len(data)
 
+        if self._missing_value_replacement == 'random':
+            self._min_value = data.min()
+            self._max_value = data.max()
+
     def transform(self, data):
         """Replace null values with the indicated ``missing_value_replacement``.
 
@@ -127,7 +131,15 @@ class NullTransformer():
             numpy.ndarray
         """
         isna = data.isna()
-        if isna.any() and self._missing_value_replacement is not None:
+        if self._missing_value_replacement == 'random':
+            data_mask = list(np.random.uniform(
+                low=self._min_value,
+                high=self._max_value,
+                size=len(data)
+            ))
+            data = data.mask(data.isnull(), data_mask)
+
+        elif isna.any() and self._missing_value_replacement is not None:
             data = data.fillna(self._missing_value_replacement)
 
         if self._missing_value_generation == 'from_column':
--- a/rdt/transformers/null.py
+++ b/rdt/transformers/null.py
@@ -44,6 +44,8 @@ class NullTransformer():
             )
 
         self._missing_value_generation = missing_value_generation
+        self._min_value = None
+        self._max_value = None
 
     def models_missing_values(self):
         """Indicate whether this transformer creates a null column on transform.
@@ -137,7 +139,7 @@ class NullTransformer():
                 high=self._max_value,
                 size=len(data)
             ))
-            data = data.mask(data.isnull(), data_mask)
+            data = data.mask(data.isna(), data_mask)
 
         elif isna.any() and self._missing_value_replacement is not None:
             data = data.fillna(self._missing_value_replacement)
--- a/rdt/transformers/null.py
+++ b/rdt/transformers/null.py
@@ -90,10 +90,6 @@ class NullTransformer():
         if self._missing_value_replacement == 'mode':
             return data.mode(dropna=True)[0]
 
-        if self._missing_value_replacement == 'random':
-            self._min_value = data.min()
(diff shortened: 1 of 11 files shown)


tests/integration/test_hyper_transformer.py
```diff
@@ -1353,8 +1379,9 @@ def test_random_seed(self):
 
         # Run
         ht.fit(data)
-        transformed = ht.transform(data)
-        reversed1 = ht.reverse_transform(transformed)
+        ht.reset_randomization()
```
pvk-developer, line 1382: I think that this should be solved before going any further, this should be called after `fit` on the `hypertransformer` as per #716

rdt/transformers/numerical.py
```diff
@@ -299,6 +300,7 @@ def __init__(self, model_missing_values=None, learn_rounding_scheme=False,
         super().__init__(
             model_missing_values=model_missing_values,
             missing_value_generation=missing_value_generation,
+            missing_value_replacement='mean',  # Using this as default instead of random
```
fealho, line 303: Why?

rdt/transformers/null.py
```diff
@@ -73,7 +75,7 @@ def _get_missing_value_replacement(self, data):
         if self._missing_value_replacement is None:
             return None
 
-        if self._missing_value_replacement in {'mean', 'mode'} and pd.isna(data).all():
+        if self._missing_value_replacement in {'mean', 'mode', 'random'} and pd.isna(data).all():
```
R-Palazzo, line 78: Update docstring to include `random` l73, l21

rdt/transformers/numerical.py
```diff
@@ -40,7 +40,8 @@ class FloatFormatter(BaseTransformer):
         missing_value_replacement (object):
             Indicate what to replace the null values with. If an integer or float is given,
             replace them with the given value. If the strings ``'mean'`` or ``'mode'``
-            are given, replace them with the corresponding aggregation. Defaults to ``mean``.
+            are given, replace them with the corresponding aggregation and if ``'random'``
+            replace each null value with a random value from the data. Defaults to ``random``.
```
R-Palazzo, line 44: can be confusing `replace each null value with a random value from the data.` I feel. Maybe something like `replace each null value with a random value in the data range.`

R-Palazzo (commented):
Looks good!

rdt/transformers/numerical.py
pvk-developer: Mainly because before we were using `mean` values as the only option for this transformer and @npatki suggested that this could stay the same as we would need to check how this propagates down the line to the synthesizer that uses this transformer.

fealho: I'd remove the comment (or improve it) if that's the case, it doesn't explain to someone reading the code why you are doing it.

fealho (approved):

rdt/transformers/null.py
```diff
@@ -127,7 +133,15 @@ def transform(self, data):
             numpy.ndarray
         """
         isna = data.isna()
-        if isna.any() and self._missing_value_replacement is not None:
+        if self._missing_value_replacement == 'random':
+            data_mask = list(np.random.uniform(
+                low=self._min_value,
+                high=self._max_value,
+                size=len(data)
```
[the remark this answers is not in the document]
pvk-developer: No, it has to map the full data (1-1).

rdt/transformers/null.py
```diff
@@ -114,6 +117,10 @@ def fit(self, data):
             if self._missing_value_generation == 'random':
                 self._null_percentage = null_values.sum() / len(data)
 
+        if self._missing_value_replacement == 'random':
+            self._min_value = data.min()
+            self._max_value = data.max()
```
amontanez24, line 122: Kind of an edge case, but what happens if all the values are nan? I think the min and max would be set to nan as well.

rdt/transformers/null.py
```diff
@@ -114,6 +117,10 @@ def fit(self, data):
             if self._missing_value_generation == 'random':
                 self._null_percentage = null_values.sum() / len(data)
 
+        if self._missing_value_replacement == 'random':
```
amontanez24, line 120: why don't we put this section in `self._get_missing_value_replacement`?

rdt/transformers/null.py
pvk-developer: If all values are nan we do `0` 
```python
        if self._missing_value_replacement in {'mean', 'mode', 'random'} and pd.isna(data).all():
            msg = (
                f"'missing_value_replacement' cannot be set to '{self._missing_value_replacement}'"
                ' when the provided data only contains NaNs. Using 0 instead.'
            )
            LOGGER.info(msg)
            return 0

```

amontanez24 (approved):
:shipit:

PR #720, opened 2023-10-12
Add include_lowest so that the first interval should be left inclusive

### LINKED ISSUE #719: RDT Uniform Encoder creates nan Value
### Environment Details

* RDT version: 1.4.1.dev0
* Python version: 3.11
* Operating System: Mac

### Error Description

Noticed `pd.cut(data, bins=bins, labels=labels)` returns Nan value if the value is lower than the first interval, which can cause the NaN values to appear in the reverse transform for the UniformEncoder

Code can be found here: https://github.com/sdv-dev/RDT/blob/6d19148e537d3f1faca34fdcd91b2cb6778017c4/rdt/transformers/categorical.py#L194

### Steps to reproduce

```
# Create multitable metadata
synth = HMASynthesizer(metadata)
processed_data = synth.preprocess(data)
synth.fit_processed_data(processed_data)
sample = synth.sample(10)
```

Related to https://github.com/datacebo/SDV-Enterprise/issues/241

lajohn4747:
resolves #719 

This prevents lower values from being converted to Nans which contributed to the issue seen where transformed categorical data had NaNs despite original data not containing and NaNs

### DIFF
--- a/tests/integration/transformers/test_categorical.py
+++ b/tests/integration/transformers/test_categorical.py
@@ -75,7 +75,7 @@ class TestUniformEncoder:
         # Asserts
         # Make sure there is no Nan values due to the negative number
         for col in pd.isna(output).any():
-            assert (col == False)
+            assert (not col)
 
     def test__reverse_transform_nans(self):
         """Test ``reverse_transform`` for data with NaNs."""
--- a/tests/integration/transformers/test_categorical.py
+++ b/tests/integration/transformers/test_categorical.py
@@ -70,10 +70,11 @@ class TestUniformEncoder:
         transformer.fit(data, column)
         transformed = transformer.transform(data)
         transformed.loc[1, 'column_name'] = -0.1
+        transformed.loc[2, 'column_name'] = 100
         output = transformer.reverse_transform(transformed)
 
         # Asserts
-        # Make sure there is no Nan values due to the negative number
+        # Make sure there is no Nan values due to the negative number or large upper bound number
         assert all(~pd.isna(output).any())
 
     def test__reverse_transform_nans(self):
--- a/tests/integration/transformers/test_categorical.py
+++ b/tests/integration/transformers/test_categorical.py
@@ -47,9 +47,8 @@ class TestUniformEncoder:
     def test__reverse_transform(self):
         """Test the ``reverse_transform``."""
         # Setup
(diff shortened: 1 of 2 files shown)


rwedge:
Can we update the `transformed` values in `TestUniformEncoder.test__reverse_transform` to include values less than 0 and greater than 1 to test the clipping?

tests/integration/transformers/test_categorical.py
```diff
@@ -47,7 +47,7 @@ def test_clip_interval(self):
     def test__reverse_transform(self):
         """Test the ``reverse_transform``."""
         # Setup
-        data = pd.DataFrame({'column_name': [1, 2, 3, 2, 2, 1, 3, 3, 2]})
+        data = pd.DataFrame({'column_name': [-1, 0, 1, 2, 3, 2, 2, 1, 3, 3, 2, -11, 0]})
```
amontanez24, line 50: I think since this test runs end to end it won't actually test `reverse_transform` on negative values since it runs against the transformed values on line 57. After being transformed all the negative values are removed. I think we can just add this as a unit test or update the existing one

rwedge, line 50: It'll be easier to test in unit test `test__reverse_transform`.  We'll want to modify the `transformed` data variable, not the original data

lajohn4747:
Added a new test and change the transformed value to a negative value. Confirmed it works only with include_lowest=True

tests/integration/transformers/test_categorical.py
```diff
@@ -60,6 +59,24 @@ def test__reverse_transform(self):
...
+        # Run
+        transformer.fit(data, column)
+        transformed = transformer.transform(data)
+        transformed.loc[1, 'column_name'] = -0.1
+        output = transformer.reverse_transform(transformed)
+
+        # Asserts
+        # Make sure there is no Nan values due to the negative number
+        for col in pd.isna(output).any():
+            assert (not col)
```
amontanez24, line 78: I think you can just do 
```
assert all(~pd.isna(df).any())
```

rwedge:
In this new test can we also test that our `pd.cut` call handles numbers past the upper bound?  Basically additionally set a value in `transformed` to be greater than 1 and confirm there are still no nans.

rwedge:
The CLA check is having a hard time because the emails on the git commits don't match up with any emails on your github account

lajohn4747:
> The CLA check is having a hard time because the emails on the git commits don't match up with any emails on your github account

Hmm I signed it and used the datacebo email in my local git file. Not sure why this comment still appears

PR #722, opened 2023-10-17
Make `GaussianNormalizer` distribution names consistent with `scipy`

### LINKED ISSUE #656: The `GaussianNormalizer` should accept distribution names that are consistent with `scipy`
### Problem Description
The [GaussianNormalizer](https://docs.sdv.dev/rdt/transformers-glossary/numerical/gaussiannormalizer) accepts a distribution parameter, where you can provide a string name of the univariate distribution. 

The acceptable strings are: `'gaussian'`, `'gamma'`, `'beta'`, `'student_t'`, `'truncated_gaussian'` and `'gaussian_kde'`

These names are **inconsistent** with both of the following:
- The underlying scipy library where the univariates come from and
- The [SDV library](https://docs.sdv.dev/sdv/single-table-data/modeling/synthesizers/gaussiancopulasynthesizer) which uses the same names as scipy.

### Expected behavior
Change the mappings to the following:

- `'norm'` should be used for a Gaussian distribution
- `'truncnorm'` should be used for a truncated Gaussian Distribution
- `'t'` should be used for Student T

### Additional context
In order to keep backwards compatibility, we can still accept the older names. If a user inputs them, internally use the newer names and show a Future Warning.

```python
# example: user inputs 'gaussian' instead of 'norm'
my_transformer = GaussianNormalizer(distribution='gaussian')
```
This ultimately uses the correct distribution but it shows a future warning.
```
FutureWarning: Future versions of RDT will not support 'gaussian' as an option. Please use 'norm' instead.
```

fealho:
Resolve #656.

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -251,13 +251,13 @@ class GaussianNormalizer(FloatFormatter):
             Copulas univariate distribution to use. Defaults to ``truncated_gaussian``.
             Options include:
 
-                * ``gaussian``: Use a Gaussian distribution.
+                * ``norm``: Use a Gaussian distribution.
                 * ``gamma``: Use a Gamma distribution.
                 * ``beta``: Use a Beta distribution.
-                * ``student_t``: Use a Student T distribution.
+                * ``t``: Use a Student T distribution.
                 * ``gaussian_kde``: Use a GaussianKDE distribution. This model is non-parametric,
                   so using this will make ``get_parameters`` unusable.
-                * ``truncated_gaussian``: Use a Truncated Gaussian distribution.
+                * ``truncnorm``: Use a Truncated Gaussian distribution.
                 # ``uniform``: Use a UniformUnivariate distribution.
 
         missing_value_generation (str or None):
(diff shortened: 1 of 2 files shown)


rdt/transformers/numerical.py
```diff
@@ -311,6 +315,15 @@ def __init__(self, model_missing_values=None, learn_rounding_scheme=False,
 
         self._distributions = self._get_distributions()
         if isinstance(distribution, str):
+            if distribution in {'gaussian', 'student_t', 'truncated_gaussian'}:
+                deprecated_distributions_mapping = {
```
rwedge, line 319: could we treat this mapping as a private class constant?

rdt/transformers/numerical.py
```diff
@@ -285,13 +285,17 @@ def _get_distributions():
...
-            'student_t': univariate.StudentTUnivariate,
+            't': univariate.StudentTUnivariate,
             'gaussian_kde': univariate.GaussianKDE,
-            'truncated_gaussian': univariate.TruncatedGaussian,
+            'truncnorm': univariate.TruncatedGaussian,
             'uniform': univariate.UniformUnivariate,
+            # the following are deprecated
+            'gaussian': univariate.GaussianUnivariate,
+            'student_t': univariate.StudentTUnivariate,
+            'truncated_gaussian': univariate.TruncatedGaussian,
```
rwedge, line 298: if in the init we replaced the deprecated distribution string with the new one before calling this method we could remove the deprecated mappings from this dictionary

fealho: That's a good suggestion, thanks.

rwedge (approved):
Looks good, one more suggestion

pvk-developer (approved):

PR #728, opened 2023-10-17
Add include_lowest so that the first interval should be left inclusive

lajohn4747:
resolves https://github.com/sdv-dev/RDT/issues/719

This prevents lower values from being converted to Nans which contributed to the issue seen where transformed categorical data had NaNs despite original data not containing and NaNs

### DIFF
--- a/tests/integration/transformers/test_categorical.py
+++ b/tests/integration/transformers/test_categorical.py
@@ -76,7 +76,7 @@ class TestUniformEncoder:
 
         # Asserts
         # Make sure there is no Nan values due to the negative number or large upper bound number
-        assert not any(pd.isna(output).values)
+        assert not any(pd.isna(output).to_numpy())
 
     def test__reverse_transform_nans(self):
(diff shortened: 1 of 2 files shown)


tests/integration/transformers/test_categorical.py
```diff
@@ -60,6 +60,24 @@ def test__reverse_transform(self):
...
+        # Run
+        transformer.fit(data, column)
+        transformed = transformer.transform(data)
+        transformed.loc[1, 'column_name'] = -0.1
+        transformed.loc[2, 'column_name'] = 100
+        output = transformer.reverse_transform(transformed)
+
+        # Asserts
+        # Make sure there is no Nan values due to the negative number or large upper bound number
+        assert all(~pd.isna(output).any())
```
rwedge, line 79:
```suggestion
        assert not any(pd.isna(output).values)
```

rwedge (approved):
Looks good, left one minor suggestion

amontanez24 (approved):

PR #729, opened 2023-10-20
Issue sdv 1439 move anonymization

lajohn4747:
Related to [localization issues](https://github.com/sdv-dev/SDV/issues/1439)

Anonymization module is used for RDT only and should be moved here to avoid bringing the Faker dependency into SDV itself. This prevents us from running into conflicts and makes updates to versioning much simpler. Moving to Faker version 17 as it supports two of the needed localizations found in related [issue](https://github.com/sdv-dev/SDV/issues/1439)


frances-h (approved):
LGTM!

amontanez24 (commented):
LGTM!

PR #735, opened 2023-10-30
1.8.0 Release notes

lajohn4747:


HISTORY.md
```diff
@@ -1,5 +1,30 @@
 # History
 
+## 1.8.0 - 2023-10-30
+
```
amontanez24, line 4: Can you add a summary of the main changes? Here I think that would be the addition of the 'random' missing value replacement strategy, the `UniformEncoder` and `ClusterBasedNormalizer` bugs and the randomization bug. Can we also call out the deprecation?

HISTORY.md
```diff
@@ -1,5 +1,45 @@
...
+This release adds the 'random' missing value replacement strategy, uses random values of the dataset to fill in random values.
```
amontanez24, line 5: can we say: `, which uses`

amontanez24: I think it should be `to fill in missing values.`

HISTORY.md
```diff
@@ -1,5 +1,45 @@
 # History
 
+## 1.8.0 - 2023-10-30
+
+This release adds the 'random' missing value replacement strategy, uses random values of the dataset to fill in random values. 
+Additionally users are now able to use the `UniformUnivariate` distribution within the Gaussian Normalizer with this update.
+
+This release contains fixes for the `ClusterBasedNormalizer` which crashes in the reverse transform caused by values being out of bounds and
```
amontanez24, line 8: minor: `crashed in the reverse transform because of values being out of bounds`

HISTORY.md
```diff
@@ -1,5 +1,45 @@
...
+a patch for the randomization issue dealing with different values after applying `reset_randomization`
```
amontanez24, line 9: Can we end with a period?

HISTORY.md
```diff
@@ -1,5 +1,45 @@
...
+
+Anonymization has been moved into RDT library from SDV as it was found to self contained module for RDT and would reduce dependencies needed in SDV
```
amontanez24, line 11: Here too

amontanez24 (approved):
LGTM!

PR #736, opened 2023-11-02
 Resolve warning when applying the `'en_US'` locale myself

frances-h:
CU-86ayjz2p8
Resolve sdv-dev/SDV#1652

I changed the warning logic to prevent warning for `'en_US'`, but it seems like the actual problem was solved by updating Faker since I can no longer replicate the issue.


pvk-developer (approved):
LGTM!

fealho (approved):

PR #741, opened 2023-11-10
Add enforce_min_max_values to datetime transformers

### LINKED ISSUE #740: Add enforce_min_max_values to datetime transformers
### Problem Description

_As a user, I'd like my data to be within the same bounds as the original data since out of bounds data might not make sense logically (eg. someone is born on a day that hasn't happened yet).

### Expected behavior

- Add a `enforce_min_max_values` parameter to the `UnixTimestampEncoder` and `OptimizedTimestampEncoder`. This can be a boolean the same way as the [FloatFormatter](https://github.com/sdv-dev/RDT/blob/main/rdt/transformers/numerical.py#L31C14-L31C14) has it.
- We can learn the min and max values in the fit.
- We can clip the values back during `reverse_transform`.

R-Palazzo:
CU-86aynzc26
Resolve #740

### DIFF
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -30,9 +30,6 @@ class UnixTimestampEncoder(BaseTransformer):
         datetime_format (str):
             The strftime to use for parsing time. For more information, see
             https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior.
-        enforce_min_max_values (bool):
-            Whether or not to clip the data returned by ``reverse_transform`` to the min and
-            max values seen during ``fit``. Defaults to ``False``.
         missing_value_generation (str or None):
             The way missing values are being handled. There are three strategies:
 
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86aynzc26 RDT - Add enforce_min_max_values to datetime transformers #740](https://app.clickup.com/t/86aynzc26)

rdt/transformers/datetime.py
```diff
@@ -30,6 +30,9 @@ class UnixTimestampEncoder(BaseTransformer):
         datetime_format (str):
             The strftime to use for parsing time. For more information, see
             https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior.
+        enforce_min_max_values (bool):
```
amontanez24, line 33: Maybe we should add this as the last parameter. The main reason is that if someone runs scripts without using the key-word argument names, it might crash after this release since the 4th argument used to be a string and now it is a boolean with a different purpose. This isn't a huge deal though

amontanez24 (approved):
One comment but otherwise looks good!

pvk-developer (approved):
LGTM!

PR #753, opened 2024-01-11
Make `RegexGenerator` generate unlimited regexes even when `enforce_uniqueness=True`

### LINKED ISSUE #749: RegexGenerator should create unlimited regexes, even if unique enforcement is on
### Problem Description
When using a RegexGenerator with `enforce_uniqueness=True`, the transformer correctly ensures that all the generated regexes are unique. However, when it runs out of possible regexes, this transformer crashes and tells the user to choose a different regex instead.

```
TransformerProcessingError: The regex is not able to generate 10 unique values.
Please use a different regex for column ('ID').
```

Unfortunately this happens in the final stage (reverse transform) -- well after the bulk of the computation-heavy operations (eg. fitting a synthesizer). At this point, it is cumbersome to start over again and choose a new Regex. 

### Expected behavior
Instead of crashing with this message, it would be more user-friendly if the RegexGenerator continued generating regexes by appending suffixes/continuing the counting. The exact method would depend on the dtype of the original data.

1. If the original dtype is a string: it should restart the chain with suffix `(1)`, then with suffix `(2)`, etc. ...
2. If the original dtype is an int: It should start incrementing from the final value

It should also show a warning to tell the user this is happening (in place of the current error):

```
Warning: The regex for 'ID' can only generate 6 unique values. Additional values
may not exactly follow the provided regex. 
```

### Additional context
As a full example consider the following:

- Original dtype is string and regex is: `id_[a-c]{1}`. Then the generated values would be strings: `id_a, id_b, id_c, id_a(1), id_b(1), id_c(1), id_a(2), id_b(2), ...`
- Original dtype is a int and the regex is: `[0-3]{1}`. Then the generated values would be ints: `0, 1, 2, 3, 4, 5, 6, ...`

fealho:
CU-86ayz97gr, Resolve #749

### DIFF
--- a/rdt/transformers/text.py
+++ b/rdt/transformers/text.py
@@ -1,5 +1,6 @@
 """Transformers for text data."""
 import logging
+import warnings
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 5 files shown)


sdv-team:
Task linked: [CU-86ayz97gr RDT - RegexGenerator should create unlimited regexes, even if unique enforcement is on #749](https://app.clickup.com/t/86ayz97gr)

frances-h (approved):
Looks good! Just one question about messaging.

rdt/transformers/text.py
```diff
@@ -156,47 +188,44 @@ def _reverse_transform(self, data):
...
+                    int(generated_values[0])
+                    dtype = int
+                    counter = int(generated_values[-1]) + 1
+                except ValueError:
+                    dtype = 'other'
+                    counter = 0
+
+                while len(reverse_transformed) < sample_size:
+                    remaining_samples = sample_size - len(reverse_transformed)
+                    if dtype == int:
```
R-Palazzo, line 217: Do we need the `while` boucle for the `int` case? It seems likely to resolve in one iteration

R-Palazzo (commented):
Looks good! Just let one question

rdt/transformers/text.py
```diff
@@ -156,47 +188,44 @@ def _reverse_transform(self, data):
...
-            if self.enforce_uniqueness:
-                raise TransformerProcessingError(
-                    f'The regex generator is not able to generate {sample_size} new unique '
-                    f'values (only {remaining} unique value left). Please use '
-                    "'reset_randomization' in order to restart the generator."
-                )
-
+        if sample_size > remaining:
             self.reset_randomization()
             remaining = self.generator_size
```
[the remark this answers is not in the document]
fealho: Good catch!

rdt/transformers/text.py
fealho: Thanks, I missed that.

PR #754, opened 2024-01-17
RegexGenerator gives a confusing message: # of possibilities are shown as an imaginary number

### LINKED ISSUE #748: RegexGenerator gives a confusing message: # of possibilities are shown as an imaginary number
### Environment Details
* RDT version: 1.9.0

### Error Description
In some cases, the RegexGenerator may be asked to create more regexes than there are unique possibilities. In this case, it will produce a warning indicating that the IDs may repeat. The message also includes the total number of unique possibilities. This total number is formatted as an imaginary number instead of an integer, which is confusing to users.

In the example below, the RegexGenerator correctly concludes that there 6 possibilities yet it represents the number as `6+0j`.
```
/usr/local/lib/python3.10/dist-packages/rdt/transformers/text.py:164: UserWarning:
The data has 10 rows but the regex for 'ID' can only create (6+0j) unique values.
Some values in 'ID' may be repeated.
```

### Expected Behavior

1. The warning should be an info.log message instead. There is nothing concerning about duplicating Regex values if the unique enforcement flag is off (default).
2. The number should be represented as an integer instead of an imaginary number and it should **not** be inside parenthesis (see below).

```
Info log:
The data has 10 rows but the regex for 'ID' can only create 6 unique values.
Some values in 'ID' may be repeated.
```

### Steps to reproduce

```python
from rdt import get_demo
from rdt.transformers.text import RegexGenerator


data = get_demo()
data['ID'] = ['id_0', 'id_1', 'id_2', 'id_3', 'id_4']

ht = HyperTransformer()
ht.detect_initial_config(data=data)

ht.update_sdtypes({'ID': 'text'})
ht.update_transformers({'ID': RegexGenerator(regex_format='id_[0-5]{1}', enforce_uniqueness=False)})

double_transformed = pd.concat([transformed, transformed], axis=0).reset_index().drop(columns=['index'])

ht.reverse_transform(double_transformed)
```

### Related Issue
This was first observed in [SDV Issue 1729](https://github.com/sdv-dev/SDV/issues/1729).

R-Palazzo:
CU-86ayz967f
Resolve #748

I didn't use f-string formatting because it was raising this lint warning for info message:
`rdt/transformers/text.py:166:12: W1203: Use lazy % formatting in logging functions (logging-fstring-interpolation)`

Also you can see with this screenshot of the code of `step to reproduce` that we now print real values.
Screenshot 2024-01-17 at 10 07 59

### DIFF
--- a/rdt/transformers/text.py
+++ b/rdt/transformers/text.py
@@ -1,5 +1,5 @@
 """Transformers for text data."""
-import warnings
+import logging
 
 import numpy as np
 import pandas as pd
@@ -7,6 +7,8 @@ import pandas as pd
 from rdt.errors import TransformerProcessingError
 from rdt.transformers.base import BaseTransformer
 from rdt.transformers.utils import strings_from_regex
+
+LOGGER = logging.getLogger(__name__)
 
 
 class IDGenerator(BaseTransformer):
@@ -161,10 +163,10 @@ class RegexGenerator(BaseTransformer):
                     f"Please use a different regex for column ('{self.get_input_column()}')."
                 )
 
-            warnings.warn(
-                f"The data has {sample_size} rows but the regex for '{self.get_input_column()}' "
-                f'can only create {self.generator_size} unique values. Some values in '
-                f"'{self.get_input_column()}' may be repeated."
+            LOGGER.info(
+                "The data has %s rows but the regex for '%s' can only create %s unique values."
+                " Some values in '%s' may be repeated.",
+                sample_size, self.get_input_column(), self.generator_size, self.get_input_column()
             )
 
         remaining = self.generator_size - self.generated
--- a/rdt/transformers/utils.py
+++ b/rdt/transformers/utils.py
@@ -149,7 +149,7 @@ def strings_from_regex(regex, max_repeat=16):
             generators.append((generator, option, args))
             sizes.append(size)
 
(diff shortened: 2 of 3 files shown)


sdv-team:
Task linked: [CU-86ayz967f RDT - RegexGenerator gives a confusing message: # of possibilities are shown as an imaginary number #748](https://app.clickup.com/t/86ayz967f)

rdt/transformers/text.py
```diff
@@ -161,10 +163,10 @@ def _reverse_transform(self, data):
...
 
-            warnings.warn(
-                f"The data has {sample_size} rows but the regex for '{self.get_input_column()}' "
-                f'can only create {self.generator_size} unique values. Some values in '
-                f"'{self.get_input_column()}' may be repeated."
+            LOGGER.info(
+                "The data has %s rows but the regex for '%s' can only create %s unique values."
+                " Some values in '%s' may be repeated.",
+                sample_size, self.get_input_column(), self.generator_size, self.get_input_column()
             )
```
pvk-developer, line 170: We could do:
```python
message = (
    f"..."
    f"..."
)
``` 
But it is actually fine to have it as it is right now; The reason why, `f-string` gets 'compiled' always, meanwhile this way of string formatting gets generated only when called, which means that it will be generated only when `logger` is enabled.

pvk-developer (approved):
LGTM!

fealho (approved):
LGTM

PR #755, opened 2024-01-17
Categorical reverse transform may crash with `ValueError` for certain dtypes (int64)

### LINKED ISSUE #747: Categorical reverse transform may crash with `ValueError` for certain dtypes (int64)
### Environment Details
* RDT version: 1.9.0 (latest)
* Python version: Any
* Operating System: Any

### Error Description
Due to certain bugs in SDV (see [SDV issue 1691](https://github.com/sdv-dev/SDV/issues/1691)), the data that is being reverse transformed may erroneously contain null values. That is: There are null values in the transformed data, even though it's supposed to be fully numerical.

If this happens, then the behavior of `reverse_transform` is inconsistent depending on the dtype of the original data

- If the original dtype is a string or a float, then the reverse transform simply allows the null value to pass without any effect
- But if the original dtype is an int, then the reverse transform crashes upon trying to cast the null value to an integer (not allowed by pandas)

### Steps to reproduce

```python
import pandas as pd
import numpy as np

from rdt.transformers.categorical import UniformEncoder

# original data: categorical with dtype int64
    'A': [1, 1, 2, 2, 2, 3, 3, 3]
})

ht = HyperTransformer()
ht.set_config({
    'sdtypes': {
        'A': 'categorical',
    },
    'transformers': {
        'A': UniformEncoder()
    }
})

ht.fit(test_data)

# transformed data contains null values
bad_data = pd.DataFrame(data={
    'A': [0.102693, 0.165187, np.nan, 0.311625, 0.654629, 0.854138, 0.704197]
})

ht.reverse_transform(bad_data)
```

Output:
```
ValueError: Cannot convert float NaN to integer
```
[stack_trace.txt](https://github.com/sdv-dev/RDT/files/13814023/stack_trace.txt)

### Expected Behavior
Applies to categorical transformers: `UniformEncoder`, `OrderedUniformEncoder`, `LabelEncoder`, `OrderedLabelEncoder`

- If the dtype is int64 _and_ the reverse transform is crashing (due to null values), then show the user a warning and cast the column to float64 instead. Since float64 is allowed to contain null values, the behavior will at least be more consistent with other dtypes.

```
Warning: There are null values in the transformed data. The reversed transformed data will contain null values of type 'float'.
```

image

- If the dtype is something else (float, string), then there is no change to the functionality. However, it would be good to show a warning anyways because it is unexpected to have null values.

```
Warning: There are null values in the transformed data. The reversed transformed data will also contain null values.
```

R-Palazzo:
CU-86ayz7xvf
Resolve #747

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -50,6 +50,7 @@ class UniformEncoder(BaseTransformer):
             )
 
         self.order_by = order_by
+        self.is_integer = None
 
     def _order_categories(self, unique_data):
         nans = pd.isna(unique_data)
@@ -122,6 +123,7 @@ class UniformEncoder(BaseTransformer):
                 Data to fit the transformer to.
         """
         self.dtype = data.dtypes
+        self.is_integer = pd.api.types.is_integer_dtype(self.dtype)
         data = fill_nan_with_none(data)
         labels = pd.unique(data)
         labels = self._order_categories(labels)
@@ -181,6 +183,7 @@ class UniformEncoder(BaseTransformer):
         bins = [0]
         labels = []
         nan_name = 'NaN'
+        convert_to_float = False
         while nan_name in self.intervals.keys():
             nan_name += '_'
 
@@ -191,8 +194,26 @@ class UniformEncoder(BaseTransformer):
             else:
                 labels.append(key)
 
+        if pd.isna(data).any():
+            message = (
+                'There are null values in the transformed data. The reversed '
+                'transformed data will contain null values'
+            )
+            if self.is_integer:
+                message += " of type 'float'."
+                convert_to_float = True
+            else:
+                message += '.'
+
+            warnings.warn(message)
+
         result = pd.cut(data, bins=bins, labels=labels, include_lowest=True)
-        return result.replace(nan_name, np.nan).astype(self.dtype)
+        result = result.replace(nan_name, np.nan)
+
+        if convert_to_float:
+            return result.astype(float)
+
+        return result.astype(self.dtype)
 
 
 class OrderedUniformEncoder(UniformEncoder):
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -9,7 +9,7 @@ from scipy.stats import norm
 
 from rdt.errors import TransformerInputError
 from rdt.transformers.base import BaseTransformer
-from rdt.transformers.utils import fill_nan_with_none
+from rdt.transformers.utils import check_nan_in_transform, fill_nan_with_none
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86ayz7xvf RDT - Categorical reverse transform may crash with `ValueError` for certain dtypes (int64) #747](https://app.clickup.com/t/86ayz7xvf)

rdt/transformers/categorical.py
```diff
@@ -50,6 +50,7 @@ def __init__(self, order_by=None):
             )
 
         self.order_by = order_by
+        self._is_integer = False
```
fealho, line 53: Initialize with None.

rdt/transformers/categorical.py
```diff
@@ -333,6 +342,7 @@ def __init__(self, add_noise=False):
         )
         super().__init__()
         self.add_noise = add_noise
+        self._is_integer = False
```
fealho, line 345: Same

rdt/transformers/categorical.py
```diff
@@ -711,6 +723,7 @@ def __init__(self, add_noise=False, order_by=None):
             )
 
         self.order_by = order_by
+        self._is_integer = False
```
fealho, line 726: Same

rdt/transformers/categorical.py
```diff
@@ -516,6 +527,7 @@ def _reverse_transform(self, data):
         Returns:
             pandas.Series
         """
+        check_nan_in_transform(data, self._is_integer)
```
fealho, line 530: Assign it to something.

fealho (commented):
Could you also copy/paste the example from the issue?

rdt/transformers/categorical.py
R-Palazzo: The FrequencyEncoder doesn't need the output of `check_nan_in_transform` because it converts to float somewhere else. The function is just call to raise the warning if we are in the situation of the issue

R-Palazzo:
The example from the issue now:
Screenshot 2024-01-17 at 17 26 42

fealho (approved):

rdt/transformers/categorical.py
```diff
@@ -192,7 +195,12 @@ def _reverse_transform(self, data):
                 labels.append(key)
 
         result = pd.cut(data, bins=bins, labels=labels, include_lowest=True)
-        return result.replace(nan_name, np.nan).astype(self.dtype)
+        result = result.replace(nan_name, np.nan)
+
+        if convert_to_float:
+            return result.astype(float)
+
+        return result.astype(self.dtype)
```
amontanez24, line 203: I think instead of storing if the type needs to be converted to float we can just wrap the conversion in a try. If it fails, then we can see if the dtype is int and convert to float. Otherwise we just error. Then the `check_nan_in_transform` function can just be used to warn

amontanez24 (changes requested):

rdt/transformers/categorical.py
R-Palazzo: Thanks for the advice, I made the change

amontanez24 (approved):
LGTM!

PR #758, opened 2024-02-01
Add a _update_multi_column_transformer method

### LINKED ISSUE #757: Add a _update_multi_column_transformer method
### Problem Description

Currently, when a user calls `update_transformers`, he can modify any column to transformer mapping, including multi-column transformer. In this case, these changes can cause some multi-column transformers not to work (for instance, when they no longer have enough columns in their tuple).

### Expected behavior

A  `_update_multi_column_transformer` should be added to check that the new mappings are valid for the multi-columns.
For each multi-column, this method should:
- Create a `columns_to_sdtype` dict based on the tuple of column
- Validate the transformer by calling its `_validate_sdtype()` in a try/catch
- If didn't crash, this means the config is valid. 
- If it crashes, the multi-column should be removed and each column should be assigned to its default transformer per sdtype. Also, a warning should be raised saying:
```python
Userwarning: "Transformer 'RandomlocationGenerator' is incompatible with the multi-column field 'A'. Assigning default transformer to the columns."
```
- Update the `_multi_column_fields`

Finally, this method should be called at the end of:
- `update_transformers`
-  `remove_transformers`
- `remove_transformers_by_sdtype`
- `update_sdtypes`

R-Palazzo:
CU-86az7g4wc
Resolve #757

### DIFF
--- a/rdt/hyper_transformer.py
+++ b/rdt/hyper_transformer.py
@@ -357,6 +357,30 @@ class HyperTransformer:
 
         self.field_transformers[new_tuple] = self.field_transformers.pop(old_tuple)
 
+    def _update_multi_column_transformer(self):
+        """Check that multi-columns mappings are valid and update them otherwise."""
+        all_fields_multi_column = set(self._multi_column_fields.values())
+        for field in all_fields_multi_column:
+            transformer = self.field_transformers[field]
+            if transformer is None:
+                continue
+
+            columns_to_sdtypes = self._get_columns_to_sdtypes(field)
+            try:
+                transformer._validate_sdtypes(   # pylint: disable=protected-access
+                    columns_to_sdtypes
+                )
+            except TransformerInputError:
+                warnings.warn(
+                    f"Transformer '{transformer.get_name()}' is incompatible with the "
+                    f"multi-column field '{field}'. Assigning default transformer to the columns."
+                )
+                del self.field_transformers[field]
+                for column, sdtype in columns_to_sdtypes.items():
+                    self.field_transformers[column] = deepcopy(get_default_transformer(sdtype))
+
+        self._multi_column_fields = self._create_multi_column_fields()
+
     def update_transformers_by_sdtype(
             self, sdtype, transformer=None, transformer_name=None, transformer_parameters=None):
         """Update the transformers for the specified ``sdtype``.
@@ -397,6 +421,7 @@ class HyperTransformer:
                     self._remove_column_in_multi_column_fields(field)
 
         self._multi_column_fields = self._create_multi_column_fields()
+        self._update_multi_column_transformer()
         self._modified_config = True
 
     def update_sdtypes(self, column_name_to_sdtype):
@@ -445,6 +470,7 @@ class HyperTransformer:
         )
 
         self._multi_column_fields = self._create_multi_column_fields()
+        self._update_multi_column_transformer()
         self._modified_config = True
         if self._fitted:
             warnings.warn(self._REFIT_MESSAGE)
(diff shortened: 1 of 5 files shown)


sdv-team:
Task linked: [CU-86az7g4wc RDT - Add a _update_multi_column_transformer method #757](https://app.clickup.com/t/86az7g4wc)

tests/integration/test_hyper_transformer.py
```diff
@@ -1846,3 +1851,80 @@ def test_with_tuple_returned_by_faker(self):
...
+
+    methods_to_inputs = {
+        'update_transformers': {'column_name_to_transformer': {'C': UniformEncoder()}},
+        'update_transformers_by_sdtype': {'sdtype': 'boolean', 'transformer': UniformEncoder()},
+        'remove_transformers': {'column_names': ['C']},
+        'remove_transformers_by_sdtype': {'sdtype': 'boolean'},
+    }
+
+    @pytest.mark.parametrize('method_name', methods_to_inputs.keys())
+    def test_unvalid_multi_column(self, method_name):
```
amontanez24, line 1863: minor: invalid

tests/integration/test_hyper_transformer.py
```diff
@@ -1846,3 +1851,80 @@ def test_with_tuple_returned_by_faker(self):
...
+        """Test that the Hypertransformer handles invalid multi column transformers."""
```
amontanez24, line 1864: I think we can be more specific here and add a couple sentences after saying that this tests the case where the update/remove call leaves the transformer as invalid

tests/integration/test_hyper_transformer.py
```diff
@@ -1846,3 +1851,80 @@ def test_with_tuple_returned_by_faker(self):
...
+            }
+        }
+        if method_name.startswith('update'):
+            expected_dict_config['transformers'] = {
+                'A': UniformEncoder(),
+                'E': UniformEncoder(),
+                'C': UniformEncoder(),
+                'B': UniformEncoder(),
+                'D': UniformEncoder()
+            }
```
amontanez24, line 1917: you could parametrize this part as well. Basically instead of method_to_inputs being a dict, make it a list of tuples where each tuple has the structure `(method_name, input, expected)`, and then just use those variables here accordingly

amontanez24 (commented):
This looks good, mostly minor comments!

rdt/hyper_transformer.py
```diff
@@ -357,6 +357,30 @@ def _remove_column_in_multi_column_fields(self, column):
...
+            try:
+                transformer._validate_sdtypes(   # pylint: disable=protected-access
+                    columns_to_sdtypes
+                )
+            except TransformerInputError:
+                warnings.warn(
+                    f"Transformer '{transformer.get_name()}' is incompatible with the "
+                    f"multi-column field '{field}'. Assigning default transformer to the columns."
+                )
+                del self.field_transformers[field]
```
frances-h, line 378: If you delete the field transformer here, would it be possible to hit a `KeyError` on line 364 if another field is also mapped to the same transformer?

tests/integration/test_hyper_transformer.py
R-Palazzo: Yes thanks, done in [e6083f7](https://github.com/sdv-dev/RDT/pull/758/commits/e6083f79907c1491785a1063acb9ae4804f15293)

tests/integration/test_hyper_transformer.py
R-Palazzo: That's a good point, done in [e6083f7](https://github.com/sdv-dev/RDT/pull/758/commits/e6083f79907c1491785a1063acb9ae4804f15293)

frances-h (approved):

amontanez24 (approved):
:shipit:

PR #762, opened 2024-02-22
Add `cardinality_rule` parameter to `AnonymizedFaker`

### LINKED ISSUE #756: Allow `AnonymizedFaker` to learn cardinality from the real data
### Problem Description
I have a column with sensitive information such as SSNs or phone numbers. In some cases, I may want to anonymize this data but I want the _cardinality_ of the column to remain the same. That is, if my original data had 20 unique SSNs (that were repeated), then I want the anonymized data to also contain max 20 randomly-generated SSNs that repeat.

### Expected Behavior
Deprecate the `enforce_uniqueness` parameter and instead replace it with a new parameter called `cardinality_rule`. This parameter can take on 3 different values:

- (default) `None`: Do not consider cardinality. This will just allow Faker to create random values, without any regard to cardinality (same as today's `enforce_uniqueness=False`)
- `unique`: Enforce that every created value is unique (the same as today's `enforce_uniqueness=True`)
- `match`: Match the cardinality of the data seen during `fit`. Eg. if the data contains 20 unique values, then enforce that max 20 unique values are created during reverse transform.

### Additional context
For backwards compatibility, we can still accept the `enforce_uniqueness` and provide a deprecation notice when it is used.

```python
transformer = AnonymizedFaker(enforce_uniqueness=True)
FutureWarning: The 'enforce_uniqueness' parameter is no longer supported. Please use the 'cardinality_rule' parameter instead.
```

fealho:
CU-86az4dgdr, Resolve #756.


sdv-team:
Task linked: [CU-86az4dgdr RDT - Allow `AnonymizedFaker` to learn cardinality from the real data #756](https://app.clickup.com/t/86az4dgdr)

tests/integration/transformers/test_categorical.py
```diff
@@ -474,7 +474,6 @@ def test_one_hot_categoricals():
 
     # Run
     transformed_data = transformer.fit_transform(test_data, column='A')
-
```
R-Palazzo, line 477: Should we keep this blank line?

R-Palazzo (commented):
Looks good! Can we add an integration test for the case `cardinality_rule=match` to check that we generate the same cardinality or lower than inside the real data

rdt/transformers/pii/anonymizer.py
```diff
@@ -156,7 +172,11 @@ def reset_randomization(self):
 
     def _function(self):
         """Return the result of calling the ``faker`` function."""
-        faker_attr = self.faker.unique if self.enforce_uniqueness else self.faker
+        if self.cardinality_rule in {'unique', 'match'}:
+            faker_attr = self.faker.unique
+        elif self.cardinality_rule is None:
```
gsheni, line 177:
```suggestion
        elif not self.cardinality_rule:
```

rdt/transformers/pii/anonymizer.py
```diff
@@ -98,10 +106,18 @@ def _check_locales(self):
...
 
     def __init__(self, provider_name=None, function_name=None, function_kwargs=None,
-                 locales=None, enforce_uniqueness=False, missing_value_generation='random'):
+                 locales=None, cardinality_rule=None, enforce_uniqueness=False,
+                 missing_value_generation='random'):
         super().__init__()
+        self._data_cardinality = None
         self.data_length = None
-        self.enforce_uniqueness = enforce_uniqueness
+        self.cardinality_rule = cardinality_rule
```
gsheni, line 114: You can do this (in case the user capitalizes the string) 
```suggestion
        self.cardinality_rule = cardinality_rule.lower()
```

tests/integration/transformers/pii/test_anonymizer.py
```diff
@@ -157,6 +157,97 @@ def test_enforce_uniqueness(self):
...
+        })
+        instance = AnonymizedFaker(cardinality_rule='match')
+
+        # Run
+        transformed = instance.fit(data_fit, 'col')
+        transformed = instance.transform(data_transform)
+        reverse_transform = instance.reverse_transform(transformed)
+
+        # Assert
+        assert len(reverse_transform['col'].unique()) <= 3
```
rwedge, line 248: is the left side length variable per test run?

fealho: @rwedge No because the run is deterministic. I used the <= to make the intent of the test clear, ie as long as that value is less than 4 the function did its job correctly. I can change it if you think it's warranted.

R-Palazzo (approved):

PR #815, opened 2024-04-30
Refactoring code for Enterprise issue #529

amontanez24:
This PR alters the logic of `_reverse_transform` so that it can work naturally with the random regex transformer.

The previous logic was:
1. If the amount of values the generator has left is more than the number of samples asked for, sample them all.
2. If the amount of values the generator has left is less than the number of sample asked for, sample the rest of the generator.
    - If `enforce_uniqueness` is enabled, then add suffixes to old values to make more.
    - If `enforce_uniqueness` is disabled, copy previous values

This doesn't work for the random regex generator for two reasons:
1. It can't always sample all the values because it has collisions
2. You can't easily get all "remaining" values because technically it's unlimited.

The new logic that should work for both cases is
1. Sample as many values from the generator as you can until you either get enough or hit an exception.
    - The exception means the generator either ran out (in the not random case) or we had too many collisions (in the random case)
2. If more values are required, add them by:
    - If `enforce_uniqueness` is enabled, then add suffixes to old values to make more.
    - If `enforce_uniqueness` is disabled, copy previous values

CU-86b04th7c

### DIFF
--- a/rdt/transformers/text.py
+++ b/rdt/transformers/text.py
@@ -216,7 +216,7 @@ class RegexGenerator(BaseTransformer):
             try:
                 generated_values.append(next(self.generator))
                 self.generated += 1
-            except Exception:  # pylint: disable=broad-exception-caught
+            except StopIteration:
                 # Can't generate more rows without collision so breaking out of loop
                 break
 
--- a/rdt/transformers/text.py
+++ b/rdt/transformers/text.py
@@ -189,6 +189,9 @@ class RegexGenerator(BaseTransformer):
                 f'values (only {max(remaining, 0)} unique values left).'
             )
 
+    def _get_remaining_generator_values(self):
+        return list(self.generator)
+
     def _reverse_transform(self, data):
         """Generate new data using the provided ``regex_format``.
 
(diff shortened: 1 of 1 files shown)


sdv-team:
Task linked: [CU-86b04th7c SDV-Enterprise - In `RegexGenerator`, provide an option to generate keys in a random manner #529](https://app.clickup.com/t/86b04th7c)

frances-h (approved):

pvk-developer (approved):

amontanez24:
I'm actually going to have to change this because I found another bug that would require more refactoring

rdt/transformers/text.py
```diff
@@ -211,14 +211,18 @@ def _reverse_transform(self, data):
...
-            reverse_transformed = generated_values[:]
-            self.generated = self.generator_size
+        generated_values = []
+        while len(generated_values) < sample_size:
+            try:
+                generated_values.append(next(self.generator))
+                self.generated += 1
+            except Exception:  # pylint: disable=broad-exception-caught
+                # Can't generate more rows without collision so breaking out of loop
+                break
```
pvk-developer, line 221: This will not raise an error, would that affect the current functionality when attempting to generate `unique` primary keys ?

amontanez24: I'm a little confused, the current functionality doesn't seem to do anything with errors. Is there something in SDV you're referring to?

This idea here is to sample as much as possible until you have enough or a `StopIteration` is raised because the generator is out of values. I can change it to catch specifically that type of error

frances-h (approved):

PR #816, opened 2024-04-30
Resolve column names being integers

lajohn4747:
resolves https://github.com/sdv-dev/SDV/issues/1935


fealho (approved):
Could you add an integration test as well?

PR #832, opened 2024-06-04
Add `set_fitted_parameters` method to `UniformEncoder`

### LINKED ISSUE #826: Add `_set_fitted_parameters` method to `UniformEncoder`
### Problem Description

### Expected behavior
Add the method `_set_fitted_parameters` to the `UnifromEncoder`. When called, the provided arguments should be set on the transformer to get it into a 'fitted' state so that it can be used to reverse transform. After being called, a user should be able to call `reverse_transform` and have it work as expected.

def _set_fitted_parameters(self, column_name, intervals, dtype='object'):
- Args:
  - column_name [str]: The name of the column for this transformer. Should be used to set `self.output_columns` and `self.columns` attributes.
  -  intervals [dict]: dictionary mapping categories to the (min, max) values of the intervals for the category, should be used to set `self.intervals`
  - dtype [str]: the desired pandas dtype to cast the data to after reverse transforming, should be used to set `self._dtype`

This method should not return anything

frances-h:
CU-86b0putec
Resolve #826

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -135,7 +135,7 @@ class UniformEncoder(BaseTransformer):
 
         self.frequencies, self.intervals = self._compute_frequencies_intervals(labels, freq)
 
-    def set_fitted_parameters(self, column_name, intervals, dtype='object'):
+    def _set_fitted_parameters(self, column_name, intervals, dtype='object'):
         """Manually set the parameters on the transformer to get it into a fitted state.
 
         Args:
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b0putec RDT - Add `set_fitted_parameters` method to `UniformEncoder` #826](https://app.clickup.com/t/86b0putec)

pvk-developer (approved):

gsheni (approved):

PR #834, opened 2024-06-06
Add `_set_fitted_parameters` method to `NullTransformer`

### LINKED ISSUE #827: Add `_set_fitted_parameters` method to `NullTransformer`
### Problem Description

### Expected behavior
Add the method `_set_fitted_parameters` to the `NullTransformer`. When called, the provided arguments should be set on the transformer to get it into a 'fitted' state so that it can be used to reverse transform. After being called, a user should be able to call `reverse_transform` and have it work as expected.

def _set_fitted_parameters(self, null_percentage):
- Args:
  - null_percentage [float]: The percentage of values that should be replaced with null values. Should be used to set `self._null_percentage`

This method should not return anything.

frances-h:
CU-86b0pv6p7
Resolve #827

### DIFF
--- a/rdt/transformers/null.py
+++ b/rdt/transformers/null.py
@@ -121,6 +121,20 @@ class NullTransformer:
             if self._missing_value_generation == 'random':
                 self._null_percentage = null_values.sum() / len(data)
 
+    def _set_fitted_parameters(self, null_percentage):
+        """Manually set the parameters on the transformer to get it into a fitted state.
+
+        Args:
+            null_percentage (float):
+                The percentage of values to replace with null values.
+        """
+        if null_percentage < 0 or null_percentage > 1.0:
+            raise ValueError('null_percentage should be a value between 0 and 1.')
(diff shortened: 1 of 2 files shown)


rwedge (approved):

rdt/transformers/null.py
```diff
@@ -121,6 +121,20 @@ def fit(self, data):
             if self._missing_value_generation == 'random':
                 self._null_percentage = null_values.sum() / len(data)
 
+    def _set_fitted_parameters(self, null_percentage):
```
R-Palazzo, line 124: Could we change the parameter name to `null_ratio`? Or adjust the code so we have no confusion between being a percentage or a ratio (range of 0 to 100 vs 0 to 1) @frances-h, @npatki

R-Palazzo (commented):
Looks good! I have just one naming question

R-Palazzo (approved):

PR #835, opened 2024-06-06
Add `_set_fitted_parameters` method to `BinaryEncoder`

### LINKED ISSUE #830: Add `_set_fitted_parameters` method to `BinaryEncoder`
### Problem Description

### Expected behavior
Add the method `_set_fitted_parameters` to the `BinaryEncoder`. When called, the provided arguments should be set on the transformer to get it into a 'fitted' state so that it can be used to reverse transform. After being called, a user should be able to call `reverse_transform` and have it work as expected.

def _set_fitted_parameters(self, column_name, null_transformer):
- Args:
  - column_name [str]: The name of the column to use for the transformer. Used to set self.columns and self._output_columns.
  - null_transformer [a NullTransformer instance]: A fitted null transformer instance that can be used to generate null values for the column.

This method should not return anything

lajohn4747:
resolves #830 
CU-86b0pvkhp

### DIFF
--- a/rdt/transformers/boolean.py
+++ b/rdt/transformers/boolean.py
@@ -110,3 +110,23 @@ class BinaryEncoder(BaseTransformer):
         data[isna] = np.nan
 
         return data
+
+    def _set_fitted_parameters(self, column_names, null_transformer):
+        """Manually set the parameters on the transformer to get it into a fitted state.
+        Args:
+            column_names [str or list]:
+                The name of the column(s) to use for the transformer. Should match the
+                self.columns and self._output_columns attribute on the null_transformer.
+                Used to set self.columns and self.output_columns.
+            null_transformer [a NullTransformer instance]:
+                A fitted null transformer instance that can be used to generate
+                null values for the column.
+        """
+        if isinstance(column_names, list):
+            self.columns = column_names
+            self.output_columns = column_names
+        else:
+            self.columns = [column_names]
+            self.output_columns = [column_names]
+
+        self.null_transformer = null_transformer
--- a/tests/unit/transformers/test_boolean.py
+++ b/tests/unit/transformers/test_boolean.py
@@ -269,14 +269,22 @@ class TestBinaryEncoder(TestCase):
     def test__set_fitted_parameters(self):
         """Test ``_set_fitted_parameters`` sets the required parameters for transformer."""
         # Setup
-        transformer = BinaryEncoder()
+        transformer_list = BinaryEncoder()
         column_names = ['col_1', 'col_2']
-        null_transformer = NullTransformer('mean')
+        null_transformer_list = NullTransformer('mean')
+
+        transformer = BinaryEncoder()
+        column_name = 'single_col'
+        null_transformer = NullTransformer('mode')
 
         # Run
-        transformer._set_fitted_parameters(column_names, null_transformer)
+        transformer_list._set_fitted_parameters(column_names, null_transformer_list)
+        transformer._set_fitted_parameters(column_name, null_transformer)
 
         # Assert
-        assert transformer.columns == column_names
-        assert transformer.output_columns == column_names
+        assert transformer_list.columns == column_names
+        assert transformer_list.output_columns == column_names
+        assert transformer_list.null_transformer == null_transformer_list
+        assert transformer.columns == [column_name]
+        assert transformer.output_columns == [column_name]
         assert transformer.null_transformer == null_transformer
--- a/tests/unit/transformers/test_boolean.py
+++ b/tests/unit/transformers/test_boolean.py
@@ -5,6 +5,7 @@ import numpy as np
 import pandas as pd
 
 from rdt.transformers import BinaryEncoder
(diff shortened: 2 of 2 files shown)


sdv-team:
Task linked: [CU-86b0pvkhp RDT - Add `set_fitted_parameters` method to `BinaryEncoder` #830](https://app.clickup.com/t/86b0pvkhp)

tests/integration/transformers/test_boolean.py
```diff
@@ -85,3 +86,24 @@ def test_boolean_missing_value_generation_none(self):
...
+        })
+        column_name = 'bool'
+        transformer = BinaryEncoder()
+
+        # Run
+        null_transformer = NullTransformer('mode')
+        transformer.reset_randomization()
+        transformer._set_fitted_parameters(column_name, null_transformer)
+        reverse = transformer.reverse_transform(transformed)
+        reverse[column_name] = np.where(reverse['bool.is_null'] == 1.0, None, reverse[column_name])
```
frances-h, line 106: We shouldn't have to do this step here since the BinaryEncoder has a built-in NullTransformer, but we do since we only populate `self.output_columns` with the column name. Instead of modifying it here, can you try appending `'bool.is_null'` to the `output_columns` parameter?

lajohn4747: Moved it but I had to call `_set_fitted_parameters` for the NullTransformer. I also added another test for `from_column` and `random` based NullTransformers

rdt/transformers/boolean.py
```diff
@@ -110,3 +110,19 @@ def _reverse_transform(self, data):
...
+            null_transformer (NullTransformer):
+                A fitted null transformer instance that can be used to generate
+                null values for the column.
+        """
+        self.reset_randomization()
+        self.columns = [column_name]
+        self.output_columns = [column_name]
+        self.null_transformer = null_transformer
+        if self.null_transformer.models_missing_values():
+            self.output_columns.append(column_name + '.is_null')
```
fealho, line 128: Is the BinaryEncoder the only transformer that generates a `is_null` columns?

lajohn4747: No it's not, anything with a `NullTransformer` would also generate it as well.

fealho: So the `FloatFormatter` PR should also have it?

lajohn4747: Yes I am updating that now ^

tests/integration/transformers/test_boolean.py
```diff
@@ -85,3 +86,50 @@ def test_boolean_missing_value_generation_none(self):
...
+        null_transformer._set_fitted_parameters(0.2)
+        transformer._set_fitted_parameters(column_name, null_transformer)
+        reverse = transformer.reverse_transform(transformed)
+
+        # Get indices that are not NaN/None as the transformer used a random ratio
+        nan_indices_data = data[data.isna().any(axis=1)].index
+        nan_indices_reverse = reverse[reverse.isna().any(axis=1)].index
+        nan_indices = nan_indices_data.union(nan_indices_reverse)
+        compare_data = data.drop(index=nan_indices)
+        compare_reverse = data.drop(index=nan_indices)
```
fealho, line 132: `data` is supposed to be `reverse`

tests/integration/transformers/test_boolean.py
```diff
@@ -85,3 +86,50 @@ def test_boolean_missing_value_generation_none(self):
...
+            'bool': [1.0, 1.0, 1.0, 0.0],
+            'bool.is_null': [0.0, 0.0, 1.0, 0.0],
+        })
+        column_name = 'bool'
+        transformer = BinaryEncoder()
+
+        # Run
+        null_transformer = NullTransformer('mode', missing_value_generation='from_column')
+        null_transformer._set_fitted_parameters(0.25)
+        transformer.reset_randomization()
```
frances-h, line 104: We shouldn't need this since we reset when we set the parameters.

tests/integration/transformers/test_boolean.py
```diff
@@ -85,3 +86,50 @@ def test_boolean_missing_value_generation_none(self):
...
+
+        # Get indices that are not NaN/None as the transformer used a random ratio
+        nan_indices_data = data[data.isna().any(axis=1)].index
+        nan_indices_reverse = reverse[reverse.isna().any(axis=1)].index
+        nan_indices = nan_indices_data.union(nan_indices_reverse)
+        compare_data = data.drop(index=nan_indices)
+        compare_reverse = data.drop(index=nan_indices)
+
+        # Assert
+        pd.testing.assert_frame_equal(compare_data, compare_reverse)
```
frances-h, line 135: We control randomness when we reverse transform so you should be able to just do this instead:
```python
expected = pd.DataFrame({'bool': [np.nan, True, np.nan, False, False, True, False, False]})
pd.testing.assert_frame_equal(expected, reverse)
```

frances-h (approved):
Looks good, just 2 small comments 👍

fealho (approved):

PR #836, opened 2024-06-06
Add `_set_fitted_parameters` to `AnonymizedFaker`

### LINKED ISSUE #831: Add `_set_fitted_parameters` method to `AnonymizedFaker`
### Problem Description

### Expected behavior
Add the method `_set_fitted_parameters` to the `AnonymizedFaker`. When called, the provided arguments should be set on the transformer to get it into a 'fitted' state so that it can be used to reverse transform. After being called, a user should be able to call `reverse_transform` and have it work as expected.

def _set_fitted_parameters(self, column_name, cardinality, nan_frequency=0.0):
- Args:
  - column_name [str]: The name of the column to use for the transformer. Should be used to set `self.columns`.
  - cardinality [int]: The number of unique values to generate. Must be set if `self.cardinality_rule` is set to `match`
  - nan_frequency [float]:  The fraction of values that should be replaced with nan values if `self.missing_value_generation` is 'random'. Should be used to set `self._nan_frequency`

This method should not return anything.

lajohn4747:
resolves #831 
CU-86b0q95zh

### DIFF
--- a/rdt/transformers/pii/anonymizer.py
+++ b/rdt/transformers/pii/anonymizer.py
@@ -469,3 +469,16 @@ class PseudoAnonymizedFaker(AnonymizedFaker):
             pandas.Series
         """
         return columns_data
+
+    def _set_fitted_parameters(self, column_name, cardinality):
+        """Manually set the parameters on the transformer to get it into a fitted state.
+        Args:
+            column_name [str]:
+                The name of the column to use for the transformer.
+            cardinality [int]:
+                The number of unique values to generate.
+            nan_frequency [float]:
+                The fraction of values that should be replaced with nan values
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b0q95zh RDT - Add `set_fitted_parameters` method to `AnonymizedFaker` #831](https://app.clickup.com/t/86b0q95zh)

frances-h (approved):

fealho (commented):
Could you add an integration test like [this](https://github.com/sdv-dev/RDT/pull/832/files#diff-333907e2077875da72e38655fe741002f6c8c744b960a6c93b276d49bb93a472R69) one.

fealho (approved):

PR #838, opened 2024-06-06
Add `_set_fitted_parameters` to `FloatFormatter`

### LINKED ISSUE #829: Add `_set_fitted_parameters` method to `FloatFormatter`
### Problem Description

### Expected behavior
Add the method `_set_fitted_parameters` to the `FloatFormatter`. When called, the provided arguments should be set on the transformer to get it into a 'fitted' state so that it can be used to reverse transform. After being called, a user should be able to call `reverse_transform` and have it work as expected.

def _set_fitted_parameters(self, output_columns, min_max_values, null_transformer, rounding_digits, dtype='float64')
- Args:
  - column_names [str]: The name of the column to use for the transformer. Used to set `self.columns` and `self.output_columns`.
  - min_max_values [tuple or None]: `None` or a tuple containing the (min, max) values for the transformer. Should be used to set `self._min_value` and `self._max_value`. If `self.enfore_min_max_values` is `True`, it cannot be `None`.
  - null_transformer [a `NullTransformer` instance]: A _fitted_ null transformer instance that can be used to generate null values for the column.
  - rounding_digits [`None` or int]: The number of digits to round to. Should be used to set `self._rounding_digits`. If `None` then `self.learn_rounding_scheme` must be `False`.
  - dtype [str]: The pandas dtype the reversed data should be converted to. Should be used to set `self._dtype`.

This method should not return anything. It should validate the parameters that are passed in, especially that the null transformer is compatible.

lajohn4747:
resolves #829 
CU-86b0pvkjj

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -227,11 +227,14 @@ class FloatFormatter(BaseTransformer):
         if self.enforce_min_max_values:
             if not min_max_values:
                 raise TransformerInputError('Must provide min and max values for this transformer.')
+
         if min_max_values:
             self._min_value = min(min_max_values)
             self._max_value = max(min_max_values)
+
         if rounding_digits:
             self._rounding_digits = rounding_digits
+
         self._dtype = dtype
 
 
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -221,6 +221,7 @@ class FloatFormatter(BaseTransformer):
             dtype (str):
                 The pandas dtype the reversed data should be converted to
         """
+        self.reset_randomization()
         self.null_transformer = null_transformer
         self.columns = [column_name]
         self.output_columns = [column_name]
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -198,6 +198,38 @@ class FloatFormatter(BaseTransformer):
 
         return data.astype(self._dtype)
 
+    def _set_fitted_parameters(
+        self,
+        column_name,
+        null_transformer,
+        rounding_digits=None,
+        min_max_values=None,
+        dtype='object',
+    ):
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b0pvkjj RDT - Add `set_fitted_parameters` method to `FloatFormatter` #829](https://app.clickup.com/t/86b0pvkjj)

rdt/transformers/numerical.py
```diff
@@ -198,6 +198,38 @@ def _reverse_transform(self, data):
...
+                The number of digits to round to.
+            dtype (str):
+                The pandas dtype the reversed data should be converted to
+        """
+        self.null_transformer = null_transformer
+        self.columns = [column_name]
+        self.output_columns = [column_name]
+        if min_max_values:
+            self._min_value = min(min_max_values)
+            self._max_value = max(min_max_values)
```
frances-h, line 228: If `self.enforce_min_max_values` is `True` we should error if these aren't provided

frances-h (approved):

fealho (commented):
Same as for your other PR, could you add an integration test?

fealho (approved):

rdt/transformers/numerical.py
```diff
@@ -198,6 +199,44 @@ def _reverse_transform(self, data):
...
+                The name of the column to use for the transformer.
+            null_transformer (NullTransformer):
+                A fitted null transformer instance that can be used to generate
+                null values for the column.
+            min_max_values (Tuple(float) or None):
+                None or a tuple containing the (min, max) values for the transformer.
+            rounding_digits (int or None):
+                The number of digits to round to.
+            dtype (str):
+                The pandas dtype the reversed data should be converted to
```
fealho, line 222: Missing full stop.

frances-h (approved):

PR #848, opened 2024-07-02
Set max string length for ruff and add auto-assign

gsheni:


tests/unit/test_hyper_transformer.py
```diff
@@ -2301,7 +2301,7 @@ def test_remove_column_in_multi_column_fields_single_column_left(self):
         assert ht._multi_column_fields == expected_column_in_tuple
 
     def test_update_transformers_by_sdtype_with_multi_column_transformer(self):
-        """Test ``update_transformers_by_sdtype`` with columns use with a multi-column transformer."""
+        """Test ``update_transformers_by_sdtype`` with columns use with a multicolumn transformer"""
```
amontanez24, line 2304: what changed this?

gsheni: We started checking docstring length and this line was too long. I shorted it.

amontanez24: Ok, can you revert for consistency? If you need to shorten, you can get rid of the ``update_transformers_by_sdtype`` part since it's redundant

amontanez24 (approved):
One comment, but after that :shipit:

PR #857, opened 2024-08-01
[dtypes] `FloatFormatter` reverse transform does not support new pandas dtypes

### LINKED ISSUE #855: [dtypes] `FloatFormatter` reverse transform does not support new pandas dtypes
### Error Description

The FloatFormatter crashes due to hard coded logic attempting to cast to a numpy data type (line 191: `is_integer = np.dtype(self._dtype).kind == 'i'`). If the dtype differs from `np.dtype`, this causes a breaking error during reverse transformation. Which is the root cause of FloatFormatter not handling the new `pandas data types`.

### Steps to reproduce

```python
from rdt.transformers import FloatFormatter
import pandas as pd
data = {
    'Int8': pd.Series([1, 2, -3], dtype='Int8'),
    'Int16': pd.Series([1, 2, -3], dtype='Int16'),
    'Int32': pd.Series([1, 2, -3], dtype='Int32'),
    'Int64': pd.Series([1, 2, -3], dtype='Int64'),
    'Float32': pd.Series([1.1, 2.2, 3.3], dtype='Float32'),
    'Float64': pd.Series([1.1, 2.2, 3.3], dtype='Float64'),
}
df = pd.DataFrame(data)

ff = FloatFormatter()
ff.fit(df, 'Int8')
transformed = ff.transform(df)
ff.reverse_transform(transformed)
```

### Expected behavior

* We should support this new dtypes. The best approach for this is to use [is_integer_dtype](https://pandas.pydata.org/docs/reference/api/pandas.api.types.is_integer_dtype.html) from pandas. This also supports `numpy.dtypes` and `uints`.
* Add an integration test to make sure that we support this new dtypes with `Null` values.

### Additional Context

Once this is fixed, we should be able to `fit` and `sample` from `SDV` (it is important to confirm `sampling` since fitting is already confirmed).

R-Palazzo:
CU-86b1gcgpc
Resolve #855

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -188,13 +188,14 @@ class FloatFormatter(BaseTransformer):
             min_bound, max_bound = INTEGER_BOUNDS[self.computer_representation]
             data = data.clip(min_bound, max_bound)
 
-        is_integer = np.dtype(self._dtype).kind == 'i'
+        is_integer = pd.api.types.is_integer_dtype(self._dtype)
+        is_pandas_instance = isinstance(data, (pd.Series, pd.DataFrame))
         if self.learn_rounding_scheme and self._rounding_digits is not None:
             data = data.round(self._rounding_digits)
         elif is_integer:
             data = data.round(0)
 
-        if pd.isna(data).any() and is_integer:
+        if pd.isna(data).any() and is_integer and not is_pandas_instance:
             return data
 
         return data.astype(self._dtype)
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -103,7 +103,7 @@ class FloatFormatter(BaseTransformer):
         )
 
     def _validate_values_within_bounds(self, data):
-        if self.computer_representation != 'Float':
+        if not self.computer_representation.startswith('Float'):
             fractions = data[~data.isna() & data % 1 != 0]
             if not fractions.empty:
                 raise ValueError(
@@ -184,7 +184,7 @@ class FloatFormatter(BaseTransformer):
         data = self.null_transformer.reverse_transform(data)
         if self.enforce_min_max_values:
             data = data.clip(self._min_value, self._max_value)
-        elif self.computer_representation != 'Float':
+        elif not self.computer_representation.startswith('Float'):
             min_bound, max_bound = INTEGER_BOUNDS[self.computer_representation]
             data = data.clip(min_bound, max_bound)
 
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -189,16 +189,13 @@ class FloatFormatter(BaseTransformer):
             data = data.clip(min_bound, max_bound)
 
         is_integer = pd.api.types.is_integer_dtype(self._dtype)
-        is_pandas_instance = isinstance(data, (pd.Series, pd.DataFrame))
+        np_integer_with_nans = isinstance(data, np.ndarray) and is_integer and pd.isna(data).any()
         if self.learn_rounding_scheme and self._rounding_digits is not None:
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b1gcgpc RDT - [dtypes] `FloatFormatter` reverse transform does not support new pandas dtypes #855](https://app.clickup.com/t/86b1gcgpc)

rdt/transformers/numerical.py
```diff
@@ -188,13 +188,14 @@ def _reverse_transform(self, data):
...
-        is_integer = np.dtype(self._dtype).kind == 'i'
+        is_integer = pd.api.types.is_integer_dtype(self._dtype)
+        is_pandas_instance = isinstance(data, (pd.Series, pd.DataFrame))
         if self.learn_rounding_scheme and self._rounding_digits is not None:
             data = data.round(self._rounding_digits)
         elif is_integer:
             data = data.round(0)
 
-        if pd.isna(data).any() and is_integer:
+        if pd.isna(data).any() and is_integer and not is_pandas_instance:
```
R-Palazzo, line 198: The new pandas dtype support NaN integers so we could remove this if statement. However I had to add the check for pandas instance to make this test work:
https://github.com/sdv-dev/RDT/blob/dd654c4911e13f95a6afe70d4d34f4013bd4f237/tests/unit/transformers/test_numerical.py#L461
Let me know if this is fine. All our transformers accept `pandas` and `np.ndarray` object right?

pvk-developer: I think we will have to revisit `RDT` on a performance audit against large inputs. The ideal scenario is that we have a `pd.DataFrame` as input (on the public method), and here we work with `pd.Series` or `np.array`, what it would be ideal is to work only with one input type and avoid the overhead of supporting two types of inputs.

pvk-developer (approved):

rdt/transformers/numerical.py
```diff
@@ -184,17 +184,18 @@ def _reverse_transform(self, data):
...
```
amontanez24, line 198: How can it be an integer type, have nans and not be pandas? Numpy doesn't support nans so if it isn't a pandas object doesn't that mean it has to either be an int type or have no nans?

R-Palazzo: Yes, you're correct. I clarified the statement in [02f0c1a](https://github.com/sdv-dev/RDT/pull/857/commits/02f0c1aa001b046c347323cffedb07cd526deab4).

There is a case where the input that was fitted was a `numpy ndarray` of `dtype int` (therefore without `NaNs`).
However, the input data that we want to `reverse_transform` from is a `np.ndarray of float with NaNs`. In this case we still have to return a `numpy ndarray of float` in order to have `NaNs`, this is what the following test expects (the test is failing with if `integer dtype` is returned):
https://github.com/sdv-dev/RDT/blob/dd654c4911e13f95a6afe70d4d34f4013bd4f237/tests/unit/transformers/test_numerical.py#L461


Nevertheless, I don't know if this case makes sense from the SDV side because it could happen that SDV generated NaNs for an integer column that didn't have any during the fitting, but because it's a  `pandas` instance and not `numpy array`, everything should be fine.

amontanez24: is_integer is calculated on the data though, so if the data is an np.ndarray of floats, is_integer would be False no?

R-Palazzo: `is_integer` is calculated on the dtype of the data that was fitted by the transformer:
https://github.com/sdv-dev/RDT/blob/02f0c1aa001b046c347323cffedb07cd526deab4/rdt/transformers/numerical.py#L191

amontanez24: ah I see. Ok

amontanez24 (approved):

PR #862, opened 2024-08-08
Lazy import BayesianGaussianMixture from sklearn

### LINKED ISSUE #861: Lazy import BayesianGaussianMixture from sklearn
### Problem Description

_As a user, I'd like import times to be quick._
As of now, anytime a user imports anything from RDT, all of the modules are imported. Certain modules scan be slow. The numerical module is especially slow because it takes time to import the `BayesianGaussianMixture` as can be seen in this log file that can be viewed using tuna.

[base_sdv_import_times.log](https://github.com/user-attachments/files/16534623/base_sdv_import_times.log)

### Expected behavior

- Running `import rdt` should be significantly faster

### Additional context

- Check to make sure performance tests aren't way worse. Maybe we can import in the init instead of the fit

amontanez24:
resolves #861 
CU-86b1knznm

TODO:

- [ ] fix unit tests

This is the new time log
Screenshot 2024-08-08 at 3 53 25 PM

Compared to the old one

Screenshot 2024-08-08 at 3 29 18 PM

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -6,7 +6,6 @@ import warnings
 import numpy as np
 import pandas as pd
 import scipy
-from sklearn.mixture import BayesianGaussianMixture
 
 from rdt.errors import TransformerInputError
 from rdt.transformers.base import BaseTransformer
@@ -521,6 +520,8 @@ class ClusterBasedNormalizer(FloatFormatter):
             data (pandas.Series):
                 Data to fit to.
         """
+        from sklearn.mixture import BayesianGaussianMixture
+
         self._bgm_transformer = BayesianGaussianMixture(
             n_components=self.max_clusters,
             weight_concentration_prior_type='dirichlet_process',
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -2,6 +2,7 @@
 
 import copy
 import warnings
+from importlib import import_module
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b1knznm RDT - Lazy import BayesianGaussianMixture from sklearn #861](https://app.clickup.com/t/86b1knznm)

gsheni (approved):

gsheni (approved):

rdt/transformers/numerical.py
```diff
@@ -521,7 +521,9 @@ def _fit(self, data):
             data (pandas.Series):
                 Data to fit to.
         """
-        self._bgm_transformer = BayesianGaussianMixture(
+        sm = import_module('sklearn.mixture')
```
pvk-developer, line 524: do we need to assign it ? If we just did `import_module` would it work ?

pvk-developer (approved):

rdt/transformers/numerical.py
amontanez24: Yeah it returns the module. If I just do `import_module('sklearn.mixture')`, I get `NameError: name 'sklearn' is not defined` when trying to access the class later. Also you can't import the class directly using import_module, only modules. So you can't do `import_module('sklearn.mixture. BayesianGaussianMixture')`

PR #881, opened 2024-09-11
Align test/id sdtypes to match `SDV` 

### LINKED ISSUE #880: Align text/id sdtypes to the SDV library
### Problem Description
Currently, the RDT library uses many of the same sdtypes as the SDV library, however there is a difference for IDs.

- **In the SDV library**, we use the sdtype `id` to refer to an identifier column that may have a structured format. By marking an sdtype as `id`, I am able to supply a regex format so that I can control what the structure looks like. By contrast, the sdtype `text` is a PII type, that is used to create random phrases or sentences.
- **In the RDT library**, there is no such thing as sdtype `id`. Instead, for historical reasons, we use sdtype `text` to refer to ID columns that can get a regex format. (For creating random text, we just use the catch-all sdtype `pii` and assign the AnonymizedFaker to the correct method.).

This difference is especially confusing for those who are using SDV's synthesizers, and are looking to update the transformers assigned to the columns. To assign transformers within an SDV synthesizer, you need to look at the RDT docs which has different sdtype requirements than SDV.

### Expected behavior
The expected behavior would be to add the sdtype `id` into the RDT library and phase out* the sdtype `text`. This means that the path should change to `from rdt.transformers.id import ...`

\* _For backwards compatibility, we should still allow users to submit sdtype `text` as an option, but write a notice that they should be using `id` instead. These two sdtypes should do the same thing in RDT._

pvk-developer:
Resolves #880 
CU-86b2378fg


sdv-team:
Task linked: [CU-86b2378fg RDT - Align text/id sdtypes to the SDV library #880](https://app.clickup.com/t/86b2378fg)

rdt/performance/datasets/text.py
```diff
@@ -1,4 +1,4 @@
-"""Dataset Generators for Text transformers."""
+"""Dataset Generators for 'text' transformers."""
```
amontanez24, line 1: Should we just delete this file now since we have the id one testing the exact same thing?

amontanez24 (commented):
Looks good! Just one suggestion

rdt/transformers/text.py
```diff
@@ -1,251 +1,12 @@
...
-                    remaining_samples = sample_size - len(reverse_transformed)
-                    reverse_transformed.extend(generated_values[:remaining_samples])
-
-        if getattr(self, 'generation_order', 'alphanumeric') == 'scrambled':
-            np.random.shuffle(reverse_transformed)
-
-        return np.array(reverse_transformed, dtype=object)
+warnings.warn(
+    "Importing 'IDGenerator' or 'RegexGenerator' for ID columns from 'rdt.transformers.text' "
+    "is deprecated. Please use 'rdt.transformers.id' instead.",
```
rwedge, line 9: This warning mentions importing from `transformers.id` instead of `transformers.text` but should it also mention switching the sdtype from text to id?

rdt/performance/datasets/text.py
pvk-developer: I left it so we can test the `text` input still works as sdtype.

rdt/transformers/text.py
pvk-developer: Not sure, cc: @npatki

pvk-developer: Addressed here [2a8d014](https://github.com/sdv-dev/RDT/pull/881/commits/2a8d014d84d1422b66fc8fe6b3d97502c67220b0)

rdt/transformers/utils.py
```diff
@@ -274,3 +274,37 @@ def learn_rounding_digits(data):
         name,
     )
     return None
+
+
+class WarnDict(dict):
+    """Custom dictionary to raise a deprecation warning."""
+
+    def get(self, key):
+        """Retrun the value for key if key is in the dictionary, else default.
```
amontanez24, line 283:
```suggestion
        """Return the value for key if key is in the dictionary, else default.
```

rdt/transformers/__init__.py
```diff
@@ -88,15 +90,15 @@ def get_transformer_name(transformer):
...
-
-DEFAULT_TRANSFORMERS = {
-    'numerical': FloatFormatter(),
-    'categorical': UniformEncoder(),
-    'boolean': UniformEncoder(),
-    'datetime': UnixTimestampEncoder(),
-    'text': RegexGenerator(),
-    'pii': AnonymizedFaker(),
-}
+DEFAULT_TRANSFORMERS = WarnDict(
```
amontanez24, line 93: Does this cause the warning to be raised more than once in one workflow? The cases I'm wondering are
1. If the user has multiple columns set to an sdtype of text
2. If the overall process for the HyperTransformer accesses this dict more than once
We probably want to make sure the user doesn't get spammed with this warning

rdt/transformers/utils.py
```diff
@@ -274,3 +274,37 @@ def learn_rounding_digits(data):
...
+
+        If the key is `text` raises a `DeprecationWarning` stating that it will be
+        phased out.
+        """
+        if key == 'text':
```
amontanez24, line 288: should we make a DEPRECATED_SDTYPES tuple and check that. We can just put text in it for now

amontanez24 (changes requested):

amontanez24 (approved):

rdt/transformers/utils.py
```diff
@@ -274,3 +274,37 @@ def learn_rounding_digits(data):
...
+            warnings.warn(
+                "The sdtype 'text' is deprecated and will be phased out. "
+                "Please use 'id' instead.",
+                DeprecationWarning,
+            )
+
+        return super().get(key)
+
+    def __getitem__(self, key):
+        """Retrun the value for key if key is in the dictionary.
```
rwedge:
```suggestion
        """Return the value for key if key is in the dictionary.
```

rwedge (approved):

PR #940, opened 2025-02-11
Set random as default missing value replacement for timestamp encoders

### LINKED ISSUE #937: Update Date Time Transformers to match documentation.
Documentation for both datetime transformers states the default for `missing_value_replacement` is `"random"`.
From the code, this does not seem to be true.
For UnixTimestampEncoder, it seems to be set to 'mean'.
For OptimizedTimestampEncoder, it seems to be set to 'None'.

For example, see this piece of code in class OptimizedTimestampEncoder
```
def __init__(
        self,
        missing_value_replacement=None,
        model_missing_values=None,
        datetime_format=None,
        missing_value_generation='random',
        enforce_min_max_values=False,
    ):
        super().__init__(
            missing_value_replacement=missing_value_replacement,
            missing_value_generation=missing_value_generation,
            enforce_min_max_values=enforce_min_max_values,
            model_missing_values=model_missing_values,
            datetime_format=datetime_format,
        )
```

gsheni:
- Fixes #937

### DIFF
--- a/tests/integration/test_hyper_transformer.py
+++ b/tests/integration/test_hyper_transformer.py
@@ -221,6 +221,8 @@ class TestHyperTransformer:
             '2010-01-01',
             '2010-01-01',
         ])
+        max_datetime = max(datetimes.dropna())
+        min_datetime = min(datetimes.dropna())
         data = pd.DataFrame(
             {
                 'integer': [1, 2, 1, 3, 1, 4, 2, 3],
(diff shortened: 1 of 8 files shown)


tests/integration/test_hyper_transformer.py
```diff
@@ -293,7 +293,7 @@ def test_default_inputs(self):
                     0.3024284729840169,
                 ],
                 'datetime': [
-                    1.2630692571428572e18,
+                    1.2631629169758298e18,
```
gsheni, line 296: I believe this changed because the `NaN` replacement for datetime is now `random`, which also changed the below `reversed_datetimes`

gsheni:
Closing due to: https://github.com/sdv-dev/RDT/issues/937#issuecomment-2654181689

PR #942, opened 2025-02-12
Allow me to opt out of replacing missing values (`missing_value_replacement=None`)

### LINKED ISSUE #938: Allow me to opt out of replacing missing values (`missing_value_replacement=None`)
### Problem Description
A number of the RDT transformers have a `missing_value_replacement` parameter that describes how to replace missing values in the forward transform. There are a variety of strategies that include replacement with a random value (`'random'`), replacing with the mean of the data (`'mean'`), etc. This was done under the assumption that the downstream software does not accept missing values.

However, I may have some use cases where my downstream software _does_ handle missing values. In this case, I do not want the RDTs to do anything to the missing values. In such cases, we should offer `missing_value_replacement=None` as an option.

_Note that we used to have this option. Just that it is listed as deprecated for all the transformers. We should un-deprecated it because there are valid use cases that need this._

### Expected behavior
Reinstate the `None` option for the `missing_value_replacement` parameter, wherever it exists. If this is passed in, then during the **forward transform**: do not replace the missing values with anything. Just pass the missing values along. 

- Note that the default value for `missing_value_replacement` should not change; in most cases, it is `'random'`. The user would need to explicitly pass in `missing_value_replacement=None` to access this new functionality.

For the **reverse transform** things can be a bit trickier. Ideally the `missing_value_generation` parameter is supposed to tell us how to recreate missing values. But it the passed-in data already contains missing values, then it can mess up generation. Here's what we can do:

If the `missing_value_generation` is not `None` _and_ the passed in data contains some null values, then don't do anything to the missing values. Instead, show the user a warning.
```
Warning: The 'missing_value_generation' parameter is set to 'from_column' but the data already contains 
missing values. Missing value generation will be skipped.
```

### Additional context
This change should probably be done to a base class. It would ultimately affect the following transformers:

- numerical: `ClusterBasedNormalizer`, `FloatFormatter`, `GaussianNormalizer`, `XGaussianNormalizer`
- categorical: `BinaryEncoder`
- datetime: `OptimizedTimestampEncoder`, `UnixTimestampEncoder`

R-Palazzo:
CU-86b3v4nm3
Resolve #938

### DIFF
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -472,7 +472,8 @@ class BaseTransformer:
             warnings.warn(
                 "The 'missing_value_generation' parameter is set to '"
                 f"{self.missing_value_generation}' but the data already contains missing values."
-                ' Missing value generation will be skipped.'
+                ' Missing value generation will be skipped.',
+                UserWarning,
             )
             self.missing_value_generation = None
 
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -142,17 +142,6 @@ class BaseTransformer:
         elif model_missing_values is False:
             self._set_missing_value_generation('random')
 
-    def _set_missing_value_replacement(self, default, missing_value_replacement):
-        if missing_value_replacement is None:
-            warnings.warn(
-                "Setting 'missing_value_replacement' to 'None' is no longer supported. "
-                f"Imputing with the '{default}' instead.",
-                FutureWarning,
-            )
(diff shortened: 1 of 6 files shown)


sdv-team:
Task linked: [CU-86b3v4nm3 RDT - Allow me to opt out of replacing missing values (`missing_value_replacement=None`) #938](https://app.clickup.com/t/86b3v4nm3)

frances-h (approved):

PR #956, opened 2025-03-13
Allow sdtype id columns to use some of the categorical transformers

### LINKED ISSUE #953: Allow sdtype `id` columns to use some of the categorical transformers
### Problem Description
For `id` columns, many of the transformers in RDT are assuming that the column is a primary or foreign key. However, I may have a general ID column that is not a key. And for such columns, I may want my final synthetic data to include the same exact values as the real data had.

An example is shown below. Both `transaction_id` and `product_id` are ID columns. The `transaction_id` is a primary key but the `product_id` is neither a primary nor foreign key. In this case, I may want my synthetic data to repeat the same product id values as my real data.

| transaction_id | date        | amount | product_id |
|----------------|-------------|--------|------------|
| t_k39jcd       | Mar 9, 2025 | 23.55  | p_39cl1    |
| t_0391ls       | Mar 8, 2025 | 100.90 | p_0ckql    |
| t_93mcp        | Mar 8, 2025 | 101.35 | p_0ckql    |
| ...        | ... | ... | ...   |

I would be able to achieve this if I was allowed to apply some of the categorical transformers for sdtype `id` columns.

### Expected behavior
Allow the following categorical transformers to be applied for `id` columns:
- LabelEncoder
- OrderedLabelEncoder
- UniformEncoder
- OrderedUniformEncoder

If possible, it would be nice to _disallow_ the following transformers from being applied to `id` columsn:
- FrequencyEncoder – it is deprecated
- OneHotEncoder – it will create too many columns to be useful, especially for high cardinality data like 'id'
- BinaryEncoder – it is only for data that has 0/1 values, which doesn't align with what an 'id' column is

### Additional context
Ideally, allowing this within RDT should also make it allowable in an SDV synthesizer. If that's not the case, then we'd need to file a new issue in SDV for it.

frances-h:
Resolve #953 
CU-86b47rdu4

Tested locally and confirmed this change allows `UniformEncoder`, `OrderedUnifromEncoder`, `LabelEncoder`, and `OrderedLabelEncoder` to be set for columns with `id` sdtype.

### DIFF
--- a/rdt/transformers/categorical.py
+++ b/rdt/transformers/categorical.py
@@ -40,7 +40,7 @@ class UniformEncoder(BaseTransformer):
     """
 
     INPUT_SDTYPE = 'categorical'
-    SUPPORTED_SDTYPES = ['categorical', 'boolean']
+    SUPPORTED_SDTYPES = ['categorical', 'boolean', 'id', 'text']
     frequencies = None
     intervals = None
     dtype = None
@@ -733,7 +733,7 @@ class LabelEncoder(BaseTransformer):
     """
 
     INPUT_SDTYPE = 'categorical'
-    SUPPORTED_SDTYPES = ['categorical', 'boolean']
+    SUPPORTED_SDTYPES = ['categorical', 'boolean', 'id', 'text']
     values_to_categories = None
     categories_to_values = None
     dtype = 'O'
--- a/tests/integration/transformers/test_categorical.py
+++ b/tests/integration/transformers/test_categorical.py
@@ -1,4 +1,5 @@
 import pickle
+import re
 import warnings
 from io import BytesIO
 
@@ -6,6 +7,7 @@ import numpy as np
 import pandas as pd
 import pytest
 
+from rdt.errors import InvalidConfigError
 from rdt.hyper_transformer import HyperTransformer
 from rdt.transformers import (
     FrequencyEncoder,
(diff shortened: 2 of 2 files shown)


sdv-team:
Task linked: [CU-86b47rdu4 RDT - Allow sdtype `id` columns to use some of the categorical transformers #953](https://app.clickup.com/t/86b47rdu4)

rdt/transformers/categorical.py
```diff
@@ -40,7 +40,7 @@ class UniformEncoder(BaseTransformer):
     """
 
     INPUT_SDTYPE = 'categorical'
-    SUPPORTED_SDTYPES = ['categorical', 'boolean']
+    SUPPORTED_SDTYPES = ['categorical', 'boolean', 'id', 'text']
```
frances-h, line 43: SDV currently assigns columns with the sdtype `id` to `text`, so I included both as supported.

R-Palazzo (commented):
Looks good!
Could you add an integration test with an HyperTransformer and try to update id columns with categorical transformers that work (`UniformEncoder`) and also one where it's expected to crash (`OneHotEncoder`), for instance?
This is a sanity check that we have all the expected behavior described in the issue.

R-Palazzo (approved):
LGTM!
Thanks for adding the tests 👍

amontanez24 (commented):
Can we merge this to a feature branch?

rdt/transformers/categorical.py
amontanez24: should we update sdv?

frances-h: We can, I can file an issue for it. Should we remove text here in that case?

amontanez24: Maybe we can ask @npatki ? I think it's probably fine to leave text

npatki: I remember `text` was the old name of the sdtype that eventually became `id`. Since we updated RDT library first to 1.0, we had called it `text`. But ultimately in SDV 1.0, we decided to call it `id`. 

I thought we went back to RDT and renamed `text` to `id` everywhere. Is that not the case?

frances-h: We did rename `'text'` to `'id'`, but left `'text'` as a supported sdtype for all the ID transformers. The SDV data processor [still assigns the `'text'` sdtype to ID columns](https://github.com/sdv-dev/SDV/blob/391fd824fad082a97207705c522a61b7ae0ca68a/sdv/data_processing/data_processor.py#L574) when it creates the config for the hyper transformer.

npatki: The journey we want to support it:
1. Metadata has listed a column's sdtype as `id` without any `regex_format`; this column is not a primary or foreign key
2. SDV's data processor should assign this column to `UniformEncoder`

I think to support this journey, RDT needs to allow `text` then right?

(We should have some separate issues to officially remove `text` from RDT though. It should all be updated to `id`.)

frances-h: Right, I think the question is do we want to include updating SDV to use `'id'` instead of `'text'` when it creates the hyper transformer with this set of changes. That would negate the need for the categorical transformers to support `'text'` (since we will only ever use `'id'`). 

If we remove `'text'` here, then we would need to make the update to the data processor with the changes to SDV to support using the categorical transformers with ID sdtypes. Or, we can leave it as is for now and put up issues in SDV/RDT to align on using `'id'` everywhere.

frances-h: Issue for updating SDV `DataProcessor` is [here](https://github.com/sdv-dev/SDV/issues/2424)

amontanez24 (approved):

Issue #950, opened 2025-03-11
For `RegexGenerator`, add add an option for `cardinality_rule='match'` to match the exact cardinality


npatki:
### Problem Description
In RDT #949, we will add a `cardinality_rule` parameter to the `RegexGenerator`. We are doing this to allow multiple ways of controlling the cardinality of the generated column.

This issue is for adding an option called `'match'` for the cardinality rule. When selected, this option will generate synthetic data with the exact same cardinality (# of unique values) that the real data had.

### Expected behavior
Add an option called `'match'` for the `cardinality_rule` parameter. When selected, the transformer should do the following:

- During **fit**: It should learn the number of unique values that were in the real data and save this info.
    - _It can also go ahead and generate Regexes of that same size, to be used every time during `reverse_transform`._
- During **transform**: No changes. It should drop the column.
- During **reverse_transform**: It should generated Regexes from scratch. Make sure that the # of unique Regexes match the value learned during `fit`.
    - **This should be global.** So if I apply `reverse_transform` multiple times and concatenate all the data together, I will only ever get the same # of unique Regexes as were present in the fit data. _Note that AnonymizedFaker currently has a bug when it comes to this functionality. See [RDT #951](https://github.com/sdv-dev/RDT/issues/951)._

After `fit`, I should be able to access the number of unique values learned using the `_data_cardinality` attribute. If `cardinality_rule` was something else other than `'match'`, then this attribute will be `None`.

```python
>>> transformer = RegexGenerator(
         regex_format='[A-Za-z1-9]{5}',
         cardinality_rule='match')
>>> transformer.fit(data)
>>> transformer._data_cardinality
10

# if you use a different cardinality rule, this attribute is None
>>> transformer = RegexGenerator(
         regex_format='[A-Za-z1-9]{5}',
         cardinality_rule=None)
>>> transformer.fit(data)
>>> transformer._data_cardinality
None
```

### Additional context
This issue is blocked on RDT #949.

PR #957, opened 2025-03-13
For `RegexGenerator`, update the `enforce_uniqueness` parameter to `cardinality_rule`

### LINKED ISSUE #949: For `RegexGenerator`, update the `enforce_uniqueness` parameter to `cardinality_rule`
### Problem Description
Currently, the [`RegexGenerator` transformer](https://docs.sdv.dev/rdt/transformers-glossary/id/regexgenerator#parameters) has a parameter called `enforce_uniqueness` that is used to (partially) control the uniqueness of the generated data. Enforcing uniqueness falls more broadly under the  concept of _controlling the cardinality_ of the column. And there are multiple ways to control the cardinality besides just uniqueness.

We should update the `enforce_uniqueness` parameter to `cardinality_rule` instead. The `cardinality_rule` parameter would accept strings, so it would allow different options for controlling the cardinality. _We eventually aim to make the RegexGenerator consistent with the [`AnonymizedFaker` transformer](https://docs.sdv.dev/rdt/transformers-glossary/id/anonymizedfaker)_.

### Expected behavior
Deprecate the `enforce_uniqueness` parameter in favor of `cardinality_rule`. Each of the options for enforcing uniqueness today will map to different options in the new `cardinality_rule` parameter:

- Today's `enforce_uniqueness=True` will map to `cardinality_rule='unique'`
- (default) Today's `enforce_uniqueness=False` will map to `cardinality_rule=None`

**Backwards compatibility**: If a user tries to apply the old `enforce_uniqueness` parameter, we should throw a `FutureWarning`.  We should always default to using the `cardinality_rule` value.

```python
         regex_format='[a-z]{3}',
         enforce_uniqueness=True)
FutureWarning: The 'enforce_uniqueness' parameter has been deprecated. 
Please use 'cardinality_rule' instead and set it to 'unique'.

         regex_format='[a-z]{3}',
         enforce_uniqueness=False)
FutureWarning: The 'enforce_uniqueness' parameter has been deprecated. 
Please use 'cardinality_rule' instead and set it to None.
```

If there is a `cardinality_rule` provided in addition to `enforce_uniqueness`, then defer to the `cardinality_rule` and ignore `enforce_uniqueness`.
```python
         regex_format='[a-z]{3}',
         enforce_uniqueness=False,
         cardinality_rule='unique')
FutureWarning: The 'enforce_uniqueness' parameter has been deprecated. 
```

R-Palazzo:
Resolve #949
CU-86b47pqbf

### DIFF
--- a/rdt/transformers/id.py
+++ b/rdt/transformers/id.py
@@ -7,11 +7,12 @@ import numpy as np
 import pandas as pd
 
 from rdt.transformers.base import BaseTransformer
-from rdt.transformers.utils import strings_from_regex
+from rdt.transformers.utils import (
+    _handle_enforce_uniqueness_and_cardinality_rule,
+    strings_from_regex,
+)
 
 LOGGER = logging.getLogger(__name__)
-_CARDINALITY_RULE_SENTINEL = object()
-
 
 class IDGenerator(BaseTransformer):
     """Generate an ID column.
@@ -124,41 +125,19 @@ class RegexGenerator(BaseTransformer):
         state['generator'] = generator
         self.__dict__ = state
 
-    def _set_cardinality_rule(
-        self, cardinality_rule, enforce_uniqueness, cardinality_rule_provided
-    ):
-        """Set the ``cardinality_rule`` parameter."""
-        if enforce_uniqueness is not None:
-            warn_message = "The 'enforce_uniqueness' parameter has been deprecated.\n"
-            if not cardinality_rule_provided:
-                cardinality_rule = 'unique' if enforce_uniqueness else None
-                warn_message += (
-                    "Please use 'cardinality_rule' instead and set it to '"
-                    f"{str(cardinality_rule)}'."
-                )
-
-            warnings.warn(warn_message, FutureWarning)
-
-        if cardinality_rule not in {None, 'unique'}:
-            raise ValueError("`cardinality_rule` must be one of 'unique' or None.")
-
-        self.cardinality_rule = cardinality_rule
-
     def __init__(
         self,
         regex_format='[A-Za-z]{5}',
-        cardinality_rule=_CARDINALITY_RULE_SENTINEL,
+        cardinality_rule=None,
         generation_order='alphanumeric',
         enforce_uniqueness=None,
     ):
         super().__init__()
         self.output_properties = {None: {'next_transformer': None}}
         self.regex_format = regex_format
-        cardinality_rule_provided = cardinality_rule is not _CARDINALITY_RULE_SENTINEL
-        if not cardinality_rule_provided:
-            cardinality_rule = None
-
-        self._set_cardinality_rule(cardinality_rule, enforce_uniqueness, cardinality_rule_provided)
+        self.cardinality_rule = _handle_enforce_uniqueness_and_cardinality_rule(
+            enforce_uniqueness, cardinality_rule
+        )
         self.data_length = None
         self.generator = None
         self.generator_size = None
--- a/rdt/transformers/id.py
+++ b/rdt/transformers/id.py
@@ -81,10 +81,14 @@ class RegexGenerator(BaseTransformer):
     Args:
         regex (str):
             String representing the regex function.
-        enforce_uniqueness (bool):
+        enforce_uniqueness (bool): Deprecated.
             Whether or not to ensure that the new generated data is all unique. If it isn't
             possible to create the requested number of rows, then an error will be raised.
             Defaults to ``False``.
+        cardinality_rule (str):
+            Rule that the generated data must follow. If set to ``unique``, the generated
+            data must be unique. If set to ``None``, then the generated data may contain
+            duplicates. Defaults to ``None``.
         generation_order (str):
             String defining how to generate the output. If set to ``alphanumeric``, it will
             generate the output in alphanumeric order (ie. 'aaa', 'aab' or '1', '2'...). If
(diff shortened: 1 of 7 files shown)


sdv-team:
Task linked: [CU-86b47pqbf RDT - For `RegexGenerator`, update the `enforce_uniqueness` parameter to `cardinality_rule` #949](https://app.clickup.com/t/86b47pqbf)

rdt/transformers/id.py
```diff
@@ -119,16 +123,48 @@ def __setstate__(self, state):
...
+        warn_message += f"Please use 'cardinality_rule' instead and set it to '{to_print}'."
+        warnings.warn(warn_message, FutureWarning)
+        self._set_cardinality_rule(cardinality_rule)
+
     def __init__(
         self,
         regex_format='[A-Za-z]{5}',
-        enforce_uniqueness=False,
+        enforce_uniqueness=None,
+        cardinality_rule=None,
```
pvk-developer, line 158: Move `enforce_uniqueness` as last argument.

rdt/transformers/id.py
```diff
@@ -119,16 +123,48 @@ def __setstate__(self, state):
...
         generation_order='alphanumeric',
     ):
         super().__init__()
         self.output_properties = {None: {'next_transformer': None}}
-        self.enforce_uniqueness = enforce_uniqueness
         self.regex_format = regex_format
+        self._set_cardinality_rule(cardinality_rule)
```
pvk-developer, line 164: make `_set_cardinality_rule` to accept `cardinality_rule, enforce_uniqueness`

rdt/transformers/id.py
```diff
@@ -119,16 +123,48 @@ def __setstate__(self, state):
...
+        if enforce_uniqueness is not None:
+            self._set_enforce_uniqueness(enforce_uniqueness)
+
```
pvk-developer, line 167: This can be removed

rdt/transformers/id.py
```diff
@@ -119,16 +123,48 @@ def __setstate__(self, state):
         state['generator'] = generator
         self.__dict__ = state
 
+    def _set_cardinality_rule(self, cardinality_rule):
```
pvk-developer, line 126:
```python
    def _set_cardinality_rule(self, cardinality_rule, enforce_uniqueness):
```

rdt/transformers/id.py
```diff
@@ -119,16 +123,48 @@ def __setstate__(self, state):
...
+                Rule that the generated data must follow. If set to ``unique``, the generated
+                data must be unique. If set to ``None``, then the generated data may contain
+                duplicates.
+        """
+        if cardinality_rule not in {None, 'unique'}:
+            raise ValueError("`cardinality_rule must` be one of 'unique' or None.")
+
+        self.cardinality_rule = cardinality_rule
+
+    def _set_enforce_uniqueness(self, enforce_uniqueness):
```
pvk-developer, line 140: Rename this to `_warn_if_enforce_uniqueness`

rdt/transformers/id.py
```diff
@@ -119,16 +123,48 @@ def __setstate__(self, state):
...
+        Args:
+            enforce_uniqueness (bool):
+                Whether or not to ensure that the new generated data is all unique.
+        """
+        warn_message = "The 'enforce_uniqueness' parameter has been deprecated.\n"
+        cardinality_rule = 'unique' if enforce_uniqueness else None
+        to_print = 'unique' if enforce_uniqueness else 'None'
+        warn_message += f"Please use 'cardinality_rule' instead and set it to '{to_print}'."
+        warnings.warn(warn_message, FutureWarning)
+        self._set_cardinality_rule(cardinality_rule)
```
pvk-developer, line 152: Remove this

rdt/transformers/id.py
```diff
@@ -119,16 +123,48 @@ def __setstate__(self, state):
...
 
+    def _set_cardinality_rule(self, cardinality_rule):
+        """Set the ``cardinality_rule`` parameter.
+
+        Args:
+            cardinality_rule (str):
+                Rule that the generated data must follow. If set to ``unique``, the generated
+                data must be unique. If set to ``None``, then the generated data may contain
+                duplicates.
+        """
```
pvk-developer, line 134:
```python
        if enforce_uniqueness is not None:
            cardinality_rule = 'unique' if enforce_uniqueness else None
            _warn_if_enforce_uniqueness(enforce_uniqueness)
```

rdt/transformers/id.py
```diff
@@ -167,7 +203,7 @@ def _warn_not_enough_unique_values(self, sample_size):
         """
         warned = False
         if sample_size > self.generator_size:
-            if self.enforce_uniqueness:
+            if self.cardinality_rule == 'unique':
```
pvk-developer, line 206: Since this can only be `self.cardinality_rule == 'unique'` or `self.cardinality_rule -> None` this can be skipped and just use `self.cardinality_rule`.
```
>>> bool('unique')
True
>>> bool(None)
False
>>> 
```

rdt/transformers/id.py
```diff
@@ -185,7 +221,7 @@ def _warn_not_enough_unique_values(self, sample_size):
                 )
 
         remaining = self.generator_size - self.generated
-        if sample_size > remaining and self.enforce_uniqueness and not warned:
+        if sample_size > remaining and self.cardinality_rule == 'unique' and not warned:
```
pvk-developer, line 224: same as above

rdt/transformers/id.py
```diff
@@ -225,7 +261,7 @@ def _reverse_transform(self, data):
         reverse_transformed = generated_values[:]
 
         if len(reverse_transformed) < sample_size:
-            if self.enforce_uniqueness:
+            if self.cardinality_rule == 'unique':
```
pvk-developer, line 264: same as above

pvk-developer (changes requested):

amontanez24 (commented):
can we point this to a feature branch?

rdt/transformers/id.py
R-Palazzo: I think we can keep it the way it is since we are going to add other options after https://github.com/sdv-dev/RDT/issues/950

rdt/transformers/id.py
```diff
@@ -10,6 +10,7 @@
 from rdt.transformers.utils import strings_from_regex
 
 LOGGER = logging.getLogger(__name__)
+_CARDINALITY_RULE_SENTINEL = object()
```
amontanez24, line 13: I spoke with @npatki and I think we can skip this part. Instead, we can always default to whatever the cardinality rule is set to. That is similar to what we do here
https://github.com/sdv-dev/RDT/blob/5d396e01ee67d6cd3e0ad1f0638dd75983f6ca4e/rdt/transformers/pii/anonymizer.py#L124-L131

amontanez24 (commented):
Can we update the code here to set the default for `enforce_uniqueness` to None, similarly to what you're doing in this PR? That way, we always raise the warning when a user passes it
https://github.com/sdv-dev/RDT/blob/5d396e01ee67d6cd3e0ad1f0638dd75983f6ca4e/rdt/transformers/pii/anonymizer.py#L124-L131

tests/unit/transformers/test_id.py
```diff
@@ -591,6 +591,8 @@ def test__reverse_transform_info_message(self, mock_logger):
         """
         # Setup
         instance = RegexGenerator('[A-Z]', cardinality_rule=None)
+        del instance.cardinality_rule  # coverage and backwards compatibility
```
amontanez24, line 594: why do we have to delete this?

amontanez24: Ah, can we just add that to the test description?

R-Palazzo: Done in [6cf6dae](https://github.com/sdv-dev/RDT/pull/957/commits/6cf6daead164fb4d4cf570f9035467b6a6f1e8f9)

amontanez24 (approved):

PR #959, opened 2025-03-14
Fix AnonymizedFaker Cardinality Issue Across Multiple Reverse Transforms

### LINKED ISSUE #951: The `AnonymizedFaker` does not match cardinality across multiple applications of `reverse_transform`
### Environment Details
* RDT version: 1.14.0
* Tested with SDV version: 1.18.0

### Error Description
The [`AnonymizedFaker` transformer](https://docs.sdv.dev/rdt/transformers-glossary/id/anonymizedfaker#parameters) has a parameter called `cardinality_rule` that's meant to control the # of unique values generated (from reverse_transform). When I apply this transformer with `cardinality_rule='match'`, I expect that the transformer should (a) learn the # of unique values in the my original data, and (b) only ever generate the same # of unique values during reverse transform.

The key part to (b) is that the **# of generated unique values should be global across all/any applications of reverse transform**. So if I reverse transform data twice and concatenate the results, it shouldn't be an issue.

In practice, this is an issue. The cardinality is match locally during each application of `reverse_transform`, but it is not match globally across multiple applications.

### Steps to reproduce
You can replicate this via an SDV synthesizer. Apply the `AnonymizedFaker` transformer with this cardinality rule. 

```python
import pandas as pd
from sdv.metadata import Metadata

# the 'name' column has 3 unique values
data = pd.DataFrame(data={'name': ['Amy']*10 + ['Bob']*20 + ['Carla']*50})

metadata = Metadata.load_from_dict({
    'tables': {
        'table': {
            'columns': {
                'name': { 'sdtype': 'id' }}}}})

synthesizer = GaussianCopulaSynthesizer(metadata)
synthesizer.auto_assign_transformers(data)
synthesizer.update_transformers({
    'name': AnonymizedFaker(
        provider_name='person',
        function_name='name', 
        cardinality_rule='match')
})
synthesizer.fit(data)

# sample synthetic data multiple times
# this essentially calls the transformer's reverse_transform multiple times
synthetic_data1 = synthesizer.sample(80)
synthetic_data2 = synthesizer.sample(80)

# stitch the synthetic data togther
all_synthetic_data = pd.concat([synthetic_data, synthetic_data2])
all_synthetic_data['name'].value_counts()
```

You can observe that each sample of synthetic data (`synthetic_data1` and `synthetic_data2`) has the correct cardinality (3 unique values). But when combined, there are actually 6 unique values. 

Image

This should not be the case. No matter how many times I sample data (reverse transform the transformer), there should only ever be 3 unique values created.

### Additional Context
A possible fix: Randomly generate new values in `fit`, after learning the # of unique values. That way, every time we `reverse_transform`, we can pick from that same list.

pvk-developer:
Resolve #951 
CU-86b47q992

A reproducible code snippet for this issue can be found [here](https://drive.google.com/drive/u/1/folders/1-33lmHW2H-eiq_mFZ7-A4VkyvFkeAr9T) including the before and after the fix.

### DIFF
--- a/tests/integration/transformers/pii/test_anonymizer.py
+++ b/tests/integration/transformers/pii/test_anonymizer.py
@@ -297,6 +297,30 @@ class TestAnonymizedFaker:
         # Assert
         assert missing_values / output.size == freq
 
+    def test_anonymized_faker_produces_only_n_values_for_each_reverse_transform_cardinality_match(
+        self,
+    ):
+        """Test `AnonymizedFaker` when `cardinality_rule` is set to `match`.
+
+        Ensure that the AnonymizedFaker transformer with `cardinality_rule='match'`
+        maintains the correct number of unique values across multiple `reverse_transform` calls.
+        """
+        # Setup
+        data = pd.DataFrame(data={'name': ['Amy'] * 10 + ['Bob'] * 20 + ['Carla'] * 50})
+        transformer = AnonymizedFaker(
+            provider_name='person', function_name='name', cardinality_rule='match'
+        )
+
+        # Run
+        transformed_data = transformer.fit_transform(data, 'name')
+        first_reverse_transformed = transformer.reverse_transform(transformed_data)
+
+        transformed_again = transformer.transform(first_reverse_transformed)
+        second_reverse_transformed = transformer.reverse_transform(transformed_again)
+
+        # Assert
+        assert set(first_reverse_transformed['name']) == set(second_reverse_transformed['name'])
+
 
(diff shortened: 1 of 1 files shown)


frances-h (commented):
Looks good, can we add the example from the issue as an integration test as well?

pvk-developer:
> Looks good, can we add the example from the issue as an integration test as well?

Added in https://github.com/sdv-dev/RDT/pull/959/commits/f78f40ace7f99614fa601e69da47cd47bd9c24b0

PR #960, opened 2025-03-20
Rename IDGenerator to IndexGenerator

### LINKED ISSUE #947: Rename `IDGenerator` transformer to `IndexGenerator`
### Problem Description
Currently the [IDGenerator transformer](https://docs.sdv.dev/rdt/transformers-glossary/id/idgenerator) is used to create an ID column from scratch when it represents an ordered index (`0, 1, 2, 3, ...`) with optional prefixes and suffixes (eg. `id_0, id_1, id_2, ...`.

While this transformer is useful for certain types of datasets, the name **IDGenerator** is not very descriptive of what it's doing. There can be many different ways to generate IDs (which we eventually plan to add); creating ordered indices is just one such way.

### Expected behavior
Rename the `IDGenerator` to `IndexGenerator` to make it clear that this transformer is specifically for ID columns that are indices.

Note that RDT transformers _should_ be backwards compatible. So if someone is still using `IDGenerator` in their code, we should throw a `FutureWarning` explaining that it has been deprecated. However, the code can continue to work, using the `IndexGenerator` under-the-hood instead.

```python
>>> from rdt.transformers.id import IDGenerator
>>> transformer = IDGenerator(prefix='id_', starting_value=0)
FutureWarning: The 'IDGenerator' has been renamed to 'IndexGenerator'.
```

gsheni:
- Fixes #947

### DIFF
--- a/rdt/transformers/id.py
+++ b/rdt/transformers/id.py
@@ -12,7 +12,7 @@ from rdt.transformers.utils import strings_from_regex
 LOGGER = logging.getLogger(__name__)
 
 
-class IDGenerator(BaseTransformer):
+class IndexGenerator(BaseTransformer):
     """Generate an ID column.
 
     This transformer generates an ID column based on a given prefix, starting value and suffix.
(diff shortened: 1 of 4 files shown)


rdt/transformers/id.py
```diff
@@ -12,7 +12,7 @@
 LOGGER = logging.getLogger(__name__)
 
 
-class IDGenerator(BaseTransformer):
+class IndexGenerator(BaseTransformer):
```
gsheni, line 15: One thing I was unsure on was, if we wanted to also rename the `id.py` file too.

gsheni: Verified with Neha the filename should stay `id.py`

R-Palazzo (approved):

frances-h (approved):

PR #963, opened 2025-03-24
Fix `BaseTransformer.__repr__` when transformer has required arguments

### LINKED ISSUE #961: `BaseTransformer.__repr__` errors if transformer has positional arguments
### Environment Details

* RDT version:
* Python version:
* Operating System:

### Error Description
The `__repr__` method on the base transformer tries to pretty print each transformer with the arguments it was called with. However, this method expects all arguments to be keyword arguments, so transformers that have positional arguments will cause an error.

We should update the `__repr__` method so it works as expected with required and keyword arguments.

### Steps to reproduce
```python
from rdt.transformers.categorical import DPResponseRandomizer

transformer = DPResponseRandomizer(1)
print(transformer)
```

fealho:
CU-86b4bz2r6, Resolve #961.

### DIFF
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -327,19 +327,19 @@ class BaseTransformer:
         custom_args = []
         args = inspect.getfullargspec(self.__init__)
         keys = args.args[1:]
-        defaults = args.defaults or []
-        defaults = dict(zip(keys, defaults))
         instanced = {
             key: getattr(self, key)
             for key in keys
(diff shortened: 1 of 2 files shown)


pvk-developer (commented):
Can we add a simple integration or unit test that showcases this, eg:
```python
class MockedTransformer(BaseTransformer):
    def __init__(mock_value):
        self.mock_value = mock_value

mocked = MockTransformer(mock_value='my_name')
assert mocked.__repr__ == 'MockedTransformer(mock_value="my_name")'
```

frances-h (approved):

PR #964, opened 2025-03-27
Improve `learn_rounding_digits` precision

fealho:
CU-86b455jpu, Resolve https://github.com/datacebo/SDV-Enterprise/issues/1107.

A float64 has 15-17 digits of precision. Anything after that should be disconsidered. For example:
```
f"{np.float64(0.1):.15f}" -> '0.100000000000000'
f"{np.float64(0.1):.20f}" -> '0.10000000000000000555'
```

These 15-17 digits include the integer part of the number, for example:
```
f"{np.float64(100.1):.12f}" -> '100.100000000000'
f"{np.float64(100.1):.15f}" -> '100.099999999999994'
```

The previous implementation always attempted to round to `MAX_DECIMALS - 1` digits, ie 14 decimal places, which could fail for cases like above where there's an integer component to the number:
```
np.float64(944343.1221).round(13) -> 944343.1220999999
np.float64(944343.1221).round(12) -> 944343.1221
```

This PR fixes this issue by rounding to `MAX_DECIMALS - num_integer_digits` instead, ensuring the situation above never happens.

### DIFF
--- a/rdt/transformers/utils.py
+++ b/rdt/transformers/utils.py
@@ -15,7 +15,7 @@ import sre_parse  # isort:skip
 
 LOGGER = logging.getLogger(__name__)
 
-MAX_DECIMALS = sys.float_info.dig - 1
+MAX_DECIMALS = sys.float_info.dig
 DEPRECATED_SDTYPES_MAPPING = {'text': 'id'}
 
 
@@ -279,8 +279,11 @@ def learn_rounding_digits(data):
         return 0
 
     # Try to round to fewer digits
-    if (roundable_data == roundable_data.round(MAX_DECIMALS)).all():
-        for decimal in range(MAX_DECIMALS + 1):
+    num_digits = len(str(int(np.max(np.abs(roundable_data)))))
+    max_rounding = MAX_DECIMALS - num_digits
+    rounded_data = roundable_data.round(max_rounding)
+    if np.isclose(roundable_data, rounded_data, atol=10**(-max_rounding - 1)).all():
+        for decimal in range(1, max_rounding + 1):
             if (roundable_data == roundable_data.round(decimal)).all():
                 return decimal
 
--- a/rdt/transformers/utils.py
+++ b/rdt/transformers/utils.py
@@ -278,9 +278,10 @@ def learn_rounding_digits(data):
     highest_int = int(np.max(np.abs(roundable_data)))
     most_digits = len(str(highest_int)) if highest_int != 0 else 0
     max_decimals = max(0, MAX_DECIMALS - most_digits)
-    for decimal in range(max_decimals + 1):
-        if (roundable_data == roundable_data.round(decimal)).all():
-            return decimal
+    if (roundable_data == roundable_data.round(max_decimals)).all():
+        for decimal in range(max_decimals + 1):
+            if (roundable_data == roundable_data.round(decimal)).all():
+                return decimal
 
     # Can't round, not equal after MAX_DECIMALS digits of precision
     LOGGER.info(
--- a/rdt/transformers/utils.py
+++ b/rdt/transformers/utils.py
@@ -270,22 +270,17 @@ def learn_rounding_digits(data):
         data = data.to_numpy()
     roundable_data = data[~(np.isinf(data.astype(float)) | pd.isna(data))]
 
-    # Doesn't contain numbers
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b455jpu SDV-Enterprise - HSA Synthesizer fails to properly learn rounding precision for some numerical columns #1107](https://app.clickup.com/t/86b455jpu)

tests/unit/transformers/test_utils.py
```diff
@@ -222,43 +222,155 @@ def test_try_convert_to_dtype():
...
+    """Test it for high digit counts."""
+    data = pd.Series(np.array([0.1234567890123456]))
+
+    output = learn_rounding_digits(data)
+
+    assert output is None
+
+
+def test_learn_rounding_digits_many_decimlas():
+    """Test it for high digit counts."""
```
R-Palazzo, line 312: Can we have a different description for each test here

rdt/transformers/utils.py
```diff
@@ -270,19 +270,17 @@ def learn_rounding_digits(data):
...
-    # Doesn't contain decimal digits
-    if (roundable_data == roundable_data.astype(int)).all():
-        return 0
-
     # Try to round to fewer digits
-    if (roundable_data == roundable_data.round(MAX_DECIMALS)).all():
-        for decimal in range(MAX_DECIMALS + 1):
-            if (roundable_data == roundable_data.round(decimal)).all():
-                return decimal
+    highest_int = int(np.max(np.abs(roundable_data)))
```
R-Palazzo, line 278: Why are we considering the integer part now?

R-Palazzo (commented):
Have you tried the changes on the colab notebook of the issue?

R-Palazzo (approved):
LGTM!
Great test to understand and double check everything 👍

fealho:
> Have you tried the changes on the colab notebook of the issue?

Added a notebook showing it fixes the original problem [here](https://drive.google.com/drive/u/5/folders/1HkDmlIsufVpcl-w9pbBCUgb-NpUOk_Sx).

tests/unit/transformers/test_utils.py
```diff
@@ -222,41 +222,66 @@ def test_try_convert_to_dtype():
...
+            1,
+        ),
+        # data_4-8: Various precision tests
+        (pd.Series([123456789012345.6789]), None),
+        (pd.Series([12345678901.234]), 3),
+        (pd.Series([12345678901.2345]), 4),
+        (pd.Series([0.1234567890123456]), None),
+        (pd.Series([0.123456789012345]), 15),
+        # data_9-10: Integer tests
+        (pd.Series([123456789012345]), 0),
```
pvk-developer, line 267: I think that you are trying to set up multiple tests cases in here, just use `@pytest.mark.parametrize(`, it is better since it will run one test per case and not having to 'wait' for all them. The positive part of it it is that will run all of them even if one fails, meanwhile yours will stop on the first `assertion` error.

tests/unit/transformers/test_utils.py
```diff
@@ -289,54 +314,6 @@ def test_learn_rounding_digits_pyarrow_float():
...
-    Input:
-    - An array of NaN.
-    Output:
-    - None
-    """
-    data = pd.Series(np.array([np.nan, np.nan, np.nan, np.nan]))
-
-    output = learn_rounding_digits(data)
-
-    assert output is None
```
pvk-developer, line 337: Can we keep this test?

fealho: It's been moved to `test_learn_rounding_digits`, I didn't delete it.

rdt/transformers/utils.py
```diff
@@ -270,19 +270,17 @@ def learn_rounding_digits(data):
...
+    # Empty dataset
     if len(roundable_data) == 0:
         return None
 
-    # Doesn't contain decimal digits
-    if (roundable_data == roundable_data.astype(int)).all():
-        return 0
-
     # Try to round to fewer digits
-    if (roundable_data == roundable_data.round(MAX_DECIMALS)).all():
```
frances-h, line 282: Can we keep this early catch if `max_decimals` won't be enough to let us round correctly so we don't enter the for loop unnecessarily?

frances-h (approved):

PR #970, opened 2025-04-01
Make `learn_rounding_digits` work for data of `object` dtype

### LINKED ISSUE #972: `learn_rounding_digits` should work for numerical data of `object` dtype
### Problem Description
The `learn_rounding_digits` method may raise the following error when running on numerical data of dtype `object`: ```TypeError: loop of ufunc does not support argument 0 of type float which has no callable rint method```.

This happens because pandas.Series doesn't support rounding when `dtype=object` for versions 2.1 or lower of pandas.

### Expected behavior
The `learn_rounding_digits` method should accept as input pandas.Series of numerical data with dtype `object`.

fealho:
Resolve #972

### DIFF
--- a/rdt/transformers/utils.py
+++ b/rdt/transformers/utils.py
@@ -274,6 +274,9 @@ def learn_rounding_digits(data):
     if len(roundable_data) == 0:
         return None
 
+    if roundable_data.dtype == 'object':
+        roundable_data = roundable_data.astype(float)
+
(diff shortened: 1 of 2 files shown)


amontanez24 (commented):
Can we add the background for this change to the description of the PR? Also file a bug for it

amontanez24 (approved):

frances-h (approved):

PR #971, opened 2025-04-01
Add `cardinality_rule='match'` option to `RegexGenerator`

### LINKED ISSUE #950: For `RegexGenerator`, add add an option for `cardinality_rule='match'` to match the exact cardinality

fealho:
CU-86b47q01w, Resolve #950.

Notes:

This PR doesn't change the behavior of passing `regex_format=''`, ie it will still generate an array of empty strings, regardless of the `cardinality_rule`.

If `cardinality_rule='match'`, every `reverse_transform` call generates values from the beginning of the `regex_format`, while if `cardinality_rule!='match'`, every call attempts to start from where the previous one left off.

### DIFF
--- a/rdt/transformers/id.py
+++ b/rdt/transformers/id.py
@@ -276,13 +276,16 @@ class RegexGenerator(BaseTransformer):
         return generated_values
 
     def _generate_num_samples(self, num_samples, template_samples):
-        """Generate num_samples values from template_samples."""
-        samples = []
-        while num_samples > 0:
-            samples.extend(template_samples[:num_samples])
-            num_samples -= len(template_samples)
+        """Generate num_samples values from template_samples.
 
-        return samples
+        Eg: num_samples = 5, template_samples = ['a', 'b']
+        The output will be ['a', 'b', 'a', 'b', 'a']
+        """
+        if num_samples <= 0:
+            return []
+
+        repeats = num_samples // len(template_samples) + 1
+        return np.tile(template_samples, repeats)[:num_samples].tolist()
 
     def _generate_match_cardinality(self, num_samples):
         """Generate values until the sample size is reached, while respecting the cardinality."""
--- a/rdt/transformers/id.py
+++ b/rdt/transformers/id.py
@@ -159,24 +159,39 @@ class RegexGenerator(BaseTransformer):
         self.generator, self.generator_size = strings_from_regex(self.regex_format)
         self.generated = 0
 
-    def _generate_fallback_samples(self, sample_size, generated_values, samples):
-        """Generate the remaining values such that they are all unique, disregarding the regex."""
+    def _generate_fallback_samples(self, num_samples, template_samples):
+        """Generate values such that they are all unique, disregarding the regex."""
         try:
-            # Integer-based fallback: attempts to convert the last generated value to an integer
-            # and then generate the remaining values in a sequential manner.
-            remaining_samples = sample_size - len(samples)
-            start = int(generated_values[-1]) + 1
-            samples.extend([str(i) for i in range(start, start + remaining_samples)])
+            # Integer-based fallback: attempt to convert the last template sample to an integer
+            # and then generate values in a sequential manner.
+            start = int(template_samples[-1]) + 1
+            return [str(i) for i in range(start, start + num_samples)]
+
         except ValueError:
-            # String-based fallback: if the integer conversion fails, it uses the last generated
-            # value as a base and appends a counter to make each value unique.
+            # String-based fallback: if the integer conversion fails, it uses the template
+            # samples as a base and appends a counter to make each value unique.
             counter = 0
-            while len(samples) < sample_size:
-                remaining_samples = sample_size - len(samples)
-                samples.extend([f'{i}({counter})' for i in generated_values[:remaining_samples]])
+            samples = []
+            while num_samples > 0:
+                samples.extend([f'{i}({counter})' for i in template_samples[:num_samples]])
+                num_samples -= len(template_samples)
                 counter += 1
 
-        return samples
+            return samples
+
+    def _generate_unique_regex_values(self):
+        regex_values = []
+        try:
+            while len(regex_values) < self._data_cardinality:
+                regex_values.append(next(self.generator))
+        except (RuntimeError, StopIteration):
+            fallback_samples = self._generate_fallback_samples(
+                num_samples=self._data_cardinality - len(regex_values),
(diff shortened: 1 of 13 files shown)


sdv-team:
Task linked: [CU-86b47q01w RDT - For `RegexGenerator`, add add an option for `cardinality_rule='match'` to match the exact cardinality #950](https://app.clickup.com/t/86b47q01w)

rdt/transformers/id.py
```diff
@@ -164,27 +203,35 @@ def _fit(self, data):
         self.reset_randomization()
         self.data_length = len(data)
 
+        if self.cardinality_rule == 'match':
+            is_nan = int(pd.isna(data).any())  # nans count as a unique value
+            self._data_cardinality = data.nunique() + is_nan
```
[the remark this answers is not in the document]
fealho: It doesn't combine `None` with `nan`:

```python3
pd.Series([np.nan, None, '1']).unique()  # Output: array([nan, None, '1'], dtype=object)
```

There's a similar problem if you use `nunique`, which is why I used this approach.

rdt/transformers/id.py
```diff
@@ -164,27 +203,35 @@ def _fit(self, data):
...
+            self._unique_regex_values = self._generate_unique_regex_values()
```
fealho, line 209: I believe this list will always have at least 1 value. It comes from the `strings_from_regex` method, which always produces a non-empty list, it either crashes for weird regexes like `(?!)` or `[1-0]{1}`, or it produces an empty string generator.

Throughout this file we make use of this assumption that this list doesn't have length 0.

rdt/transformers/id.py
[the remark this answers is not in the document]
pvk-developer: I see 🤔 , cool let's keep it that way then!

pvk-developer (approved):

rdt/transformers/id.py
amontanez24, line 208: minor: I think you can just do 
```python
self._data_cardinality = data.nunique(dropna=False)
```

rdt/transformers/id.py
```diff
@@ -203,6 +250,70 @@ def _warn_not_enough_unique_values(self, sample_size, unique_condition):
...
+                    f'Only {sample_size} values can be generated. Cannot match the cardinality '
+                    f'of the data, it requires {self._data_cardinality} values.'
+                )
+            if sample_size > self.generator_size and self._data_cardinality > self.generator_size:
+                warnings.warn(warn_msg)
+
+    def _generate_as_many_as_possible(self, num_samples):
+        """Generate samples.
+
+        Generate values following the regex until either the sample size is reached or
```
amontanez24, line 265:
```suggestion
        Generate values following the regex until either the ``num_samples`` is reached or
```

rdt/transformers/id.py
```diff
@@ -203,6 +250,70 @@ def _warn_not_enough_unique_values(self, sample_size, unique_condition):
...
+        try:
+            while len(generated_values) < num_samples:
+                generated_values.append(next(self.generator))
+                self.generated += 1
+        except (RuntimeError, StopIteration):
+            pass
+
+        return generated_values
+
+    def _generate_num_samples(self, num_samples, template_samples):
```
amontanez24, line 278: can we call this _generate_from_samples or _generate_from_template?

rdt/transformers/id.py
amontanez24, line 209: why do this in the fit instead of the beginning of reverse_transform? We can just do it there and store the value as an attribute for future calls to reverse transform

amontanez24 (commented):
The refactoring looks good! I left some minor comments. The main thing is I think we should just sample in reverse_transform only. This might also let you coalesce the code that generates regex matching values into one function (ie. all lines that say `next(self.generator)`)

rdt/transformers/id.py
fealho: @amontanez24 It unfortunately doesn't work, @pvk-developer mentioned a similar [idea](https://github.com/sdv-dev/RDT/pull/971#discussion_r2024614878).

```python3
>>> pd.Series(['1', np.nan, None]).nunique(dropna=False)
3
```

rdt/transformers/id.py
amontanez24: Is there any reason it would need to generate during fit? I think it makes more sense to just do the actual sampling during sampling.

fealho: Do you want to just move the `_generate_unique_regex_values` call to the first time `reverse_transform` is called? Or do you want to optimize this, e.g. if the `reverse_transform` data has fewer values than the fit data, only a subset of the regexes gets generated?

amontanez24: I'm fine with either. Whichever works out to be cleaner and easier

amontanez24 (approved):

PR #973, opened 2025-04-02
Add cardinality_rule='scale' option to RegexGenerator

### LINKED ISSUE #952: For `RegexGenerator`, add an option for `cardinality_rule='scale'` to scale up the cardinality based on the real data
### Problem Description
In RDT #949, we will add a `cardinality_rule` parameter to the `RegexGenerator`. We are doing this to allow multiple ways of controlling the cardinality of the generated column. Then, in RDT #950, we are adding a new cardinality rule called `'match'` to match the exact # of unique values that were present in the real data.

This issue is for adding an option called `'scale'` for the cardinality rule. When selected, this option will match the # of repetitions that each value is allowed to have. As a result, the cardinality will scale (increase) the more you generate synthetic data -- but it will increase at realistic pace. Essentially, it will "scale up" the cardinality.

### Expected behavior
Add a option for `cardinality_rule='scale'`. When selected, the `'scale'` option should work in the following way:

**fit**: Learn the # of repetitions (# of rows) that occur for each unique value in the real data. Save the distribution of this. This can be exposed after fit under the `_data_cardinality_scale` attribute. In the example below, 90% of the values repeated only 1 time, 6% repeated 2 times, and the remaining 4% repeated 10 times.
```python
>>> transformer = RegexGenerator(regex_format='[a-z]{3}', cardinality_rule='scale')
>>> transformer._data_cardinality_scale
{
  'num_repetitions': [1, 2, 10],
  'frequency': [0.9, 0.06, 0.04]
}

         regex_format='[a-z]{3}',
         cardinality_rule='match')
>>> transformer.fit(data)
>>> transformer._data_cardinality_scale
None
```

**transform**: Drop the column

**reverse transform**:  Create new values from scratch using the order specified in the `generation_order` parameter. Make sure that the # of repetitions for the value matches what we learned from the real data.
1. Create a new string that was never used before using the `generation_order` parameter.
2. Decide on the # of times it should repeat. To do this, we should do a weighted sample from the learned `_data_cardinality_scale`
3. The outputted data should have the Regex repeated that exact # of times

_Important edge case:_ It may be the case that there are too few rows to do the repetitions (or it doesn't divide up evenly). In this case, save the number of repetitions remaining so that we can continue it any future time there is a `reverse_trasform`.
- For example: Let's say we sample num_repetitions=10 for some newly created string (`'xyz'`) but there are only 6 (remaining) rows to create
- In this case, we will repeat `'xyz'` 6 times, but we will _save the fact that there are 4 more repetitions remaining for this Regex_
- The next time someone calls `reverse_transform`, the first 4 rows will have that string `'xyz'` before we create a new one

### Additional context
This issue is blocked on RDT https://github.com/sdv-dev/RDT/issues/949.

fealho:
CU-86b47r468, Resolve #952

### DIFF
--- a/rdt/transformers/id.py
+++ b/rdt/transformers/id.py
@@ -9,6 +9,7 @@ import pandas as pd
 from rdt.transformers.base import BaseTransformer
 from rdt.transformers.utils import (
     _handle_enforce_uniqueness_and_cardinality_rule,
+    fill_nan_with_none,
     strings_from_regex,
 )
 
@@ -106,8 +107,10 @@ class RegexGenerator(BaseTransformer):
         cardinality_rule (str):
             Rule that the generated data must follow.
             - If set to 'unique', the generated data must be unique.
-            - If set to 'match', the generated data must have the exact same cardinality
-              (# of unique values) as the real data.
+            - If set to 'match', the generated data will have the exact same cardinality
+              (number of unique values) as the real data.
+            - If set to 'scale', the generated data will match the number of repetitions that
+              each value is allowed to have.
             - If set to ``None``, then the generated data may contain duplicates.
             Defaults to ``None``.
         generation_order (str):
@@ -161,12 +164,22 @@ class RegexGenerator(BaseTransformer):
         self._unique_regex_values = None
         self.data_length = None
         self.generator = None
-        self.generator_size = None
-        self.generated = None
         if generation_order not in ['alphanumeric', 'scrambled']:
             raise ValueError("generation_order must be one of 'alphanumeric' or 'scrambled'.")
 
         self.generation_order = generation_order
+
+        # Used when cardinality_rule is 'scale'
+        self._data_cardinality_scale = None
+        self._remaining_samples = {'value': None, 'repetitions': 0}
+
+        # Used when cardinality_rule is 'match'
+        self._data_cardinality = None
+        self._unique_regex_values = None
+
+        # Used otherwise
+        self.generator_size = None
+        self.generated = None
 
     def reset_randomization(self):
         """Create a new generator and reset the generated values counter."""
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b47r468 RDT - For `RegexGenerator`, add an option for `cardinality_rule='scale'` to scale up the cardinality based on the real data #952](https://app.clickup.com/t/86b47r468)

rdt/transformers/id.py
```diff
@@ -172,27 +183,43 @@ def _generate_fallback_samples(self, num_samples, template_samples):
...
+
+        repetitions_counter = Counter(repetitions)
+        total_repetitions = sum(repetitions_counter.values())
+        normalized_frequency = {
+            repetition: freq / total_repetitions for repetition, freq in repetitions_counter.items()
+        }
+        sorted_repetitions = sorted(normalized_frequency.keys())
+        sorted_frequencies = [normalized_frequency[repetition] for repetition in sorted_repetitions]
+
+        return sorted_repetitions, sorted_frequencies
```
pvk-developer, line 221: I think that this could be done with pandas in a more easy way. I'm not 100% sure about the 'nan' and 'None's but I'm thinking of something like this:
```
    value_counts = data.value_counts(dropna=False)

    repetition_counts = value_counts.value_counts().sort_index()
    total = repetition_counts.sum()

    # Since this is pandas we can apply this in one go
    frequencies = (repetition_counts / total).to_numpy()
    repetitions = repetition_counts.index.to_numpy()
```

fealho: Thank you for the suggestion. The nans can be handled with the line `data = fill_nan_with_none(data)`.

pvk-developer (approved):

rdt/transformers/id.py
```diff
@@ -273,61 +287,146 @@ def _warn_not_enough_unique_values(self, sample_size, unique_condition, match_ca
...
                 self.generated += 1
         except (RuntimeError, StopIteration):
             pass
 
-        return generated_values
+        return samples
 
-    def _generate_num_samples(self, num_samples, template_samples):
-        """Generate num_samples values from template_samples.
+    def _sample_cyclic(self, num_samples, template_samples):
```
amontanez24, line 306: aalso minor: I think sample from template is more clear

amontanez24 (approved):
LGTM! Some minor comments then :shipit:

PR #979, opened 2025-04-08
Add `cardinality_rule='scale'` option for `AnonymizedFaker`

### LINKED ISSUE #954: For `AnonymizedFaker`, add an option for `cardinality_rule='scale'` to scale up the cardinality based on the real data
### Problem Description
The [`AnonymizedFaker` transformer]() has a parameter called `cardinality_rule` that allows you to control the cardinality of the generated data (from reverse transform). Currently, this transformer offers the options of (a) creating all different values (`'unique'`), useful for primary keys or (b) matching the exact # of unique values from the fit data (`'match'`). 

However, it would be nice to have third option for scaling up the cardinality of the data in a realistic way, especially if the fit data is not always unique but allows for a certain degree of repetition.

### Expected behavior
Add a option for `cardinality_rule='scale'`. When selected, the 'scale' option should work exactly the same as what's described for `RegexGenerator` in #952 -- except using Faker instead of Regexes to generate new values.

**fit**: Learn the # of repetitions (# of rows) that occur for each unique value in the real data. Save the distribution of this. This can be exposed after fit under the `_data_cardinality_scale` attribute. In the example below, 90% of the values repeated only 1 time, 6% repeated 2 times, and the remaining 4% repeated 10 times.
```python
>>> transformer = AnonymizedFaker(provider_name='person', function_name='name',
          cardinality_rule='scale')
>>> transformer._data_cardinality_scale
{
  'num_repetitions': [1, 2, 10],
  'frequency': [0.9, 0.06, 0.04]
}

>>> transformer = AnonymizedFaker(provider_name='person', function_name='name',
          cardinality_rule='match')
>>> transformer.fit(data)
>>> transformer._data_cardinality_scale
None
```

**transform**: Drop the column

**reverse transform**:  Create new values from scratch. Make sure that the # of repetitions matches what we learned from the real data.
1. Create a new Faker value that was never used before
2. Decide on the # of times it should repeat. To do this, we should do a weighted sample from the learned `_data_cardinality_scale`
3. The outputted data should have the Regex repeated that exact # of times

_Important edge case:_ It may be the case that there are too few rows to do the repetitions (or it doesn't divide up evenly). In this case, save the number of repetitions remaining so that we can continue it any future time there is a `reverse_trasform`.
- For example: Let's say we sample num_repetitions=10 for some newly created Faker string (`'John Doe'`) but there are only 6 (remaining) rows to create
- In this case, we will repeat `'John Doe'` 6 times, but we will _save the fact that there are 4 more repetitions remaining for this value
- The next time someone calls `reverse_transform`, the first 4 rows will have that value `'John Doe'` before we create a new one

fealho:
CU-86b47t377, Resolve #954.

### DIFF
--- a/tests/integration/transformers/test_id.py
+++ b/tests/integration/transformers/test_id.py
@@ -589,7 +589,7 @@ class TestRegexGenerator:
         assert np.isclose(one_count, two_count * 2, atol=samples * 0.2)
         assert np.isclose(one_count, three_count * 3, atol=samples * 0.2)
         assert len(out) == samples
-        assert more_count == 0
+        assert more_count <= 1
 
     def test_cardinality_rule_scale_not_enough_regex_numerical(self):
         """Test when cardinality rule is 'scale'."""
--- a/rdt/transformers/id.py
+++ b/rdt/transformers/id.py
@@ -8,7 +8,9 @@ import pandas as pd
 
 from rdt.transformers.base import BaseTransformer
 from rdt.transformers.utils import (
+    _get_cardinality_frequency,
     _handle_enforce_uniqueness_and_cardinality_rule,
+    _sample_repetitions,
     fill_nan_with_none,
     strings_from_regex,
 )
@@ -185,6 +187,10 @@ class RegexGenerator(BaseTransformer):
         self.generator, self.generator_size = strings_from_regex(self.regex_format)
         self.generated = 0
 
+        if hasattr(self, 'cardinality_rule') and self.cardinality_rule == 'scale':
+            self._remaining_samples['repetitions'] = 0
+            np.random.seed(self.random_seed)
+
     def _sample_fallback(self, num_samples, template_samples):
         """Sample num_samples values such that they are all unique, disregarding the regex."""
         try:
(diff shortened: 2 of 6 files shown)


sdv-team:
Task linked: [CU-86b47t377 RDT - For `AnonymizedFaker`, add an option for `cardinality_rule='scale'` to scale up the cardinality based on the real data #954](https://app.clickup.com/t/86b47t377)

pvk-developer (approved):

tests/integration/transformers/test_id.py
```diff
@@ -589,7 +589,7 @@ def assert_proportions(self, out, samples):
         assert np.isclose(one_count, two_count * 2, atol=samples * 0.2)
         assert np.isclose(one_count, three_count * 3, atol=samples * 0.2)
         assert len(out) == samples
-        assert more_count == 0
+        assert more_count <= 1
```
amontanez24, line 592: why did this change?

amontanez24 (commented):
LGTM! One small question

tests/integration/transformers/test_id.py
fealho: It's an edge case if the following happens:

 * The `regex_format` can't generate enough regexes
 * During the first reverse_transform call, not all copies of an element are generated, ie these remaining samples get stored in the `self._remaining_samples` attribute
 * Reverse gets called a second time with the same or more num_samples than the first call

In that case the second call will end up generating the same value as what had been stored in `self._remaining_samples`, so it may end up repeated more than expected.

amontanez24: What is done in the cases where cardinality_rule is not scale?

fealho: * `cardinality_rule = None`: simply loop through the generator, there are no fallback samples. `out2` would be `[a,b,a,b,a]` in this case.
* `cardinality_rule = 'unique'`: if there are not enough values remaining in the generator, reset the generator. If even that's not enough, create fallback samples based on the samples generated so far.
* `cardinality_rule = 'match'`: the first time reverse transform is called, generate all the regex values necessary to match the data cardinality seen during fit, and store it in the `_unique_regex_values` attribute. Any extra/future samples will simply be copies of the values stored in this attribute, repeated until we have enough samples.

amontanez24 (approved):

PR #986, opened 2025-04-23
Add Fallback to `AnonymizedFaker` and raise warning instead of an error.

### LINKED ISSUE #981: AnonymizedFaker should have a fallback in the case that it's unable to generate enough unique values
### Problem Description
I'm filing this issue based on a discussion in [SDV #2427](https://github.com/sdv-dev/SDV/issues/2427): Faker functions generally have a max number of unique values that they are able to generate. In the case where you request a larger # of unique values, the transformer will crash.

```python
import pandas as pd

ht = HyperTransformer()
ht.set_config({
    'sdtypes': {
        'name': 'pii'
    },
    'transformers': {
        'name': AnonymizedFaker(provider_name='person', function_name='first_name',
                                cardinality_rule='unique')
    }
})

    'name': ['Alice', 'Bob', 'Carol']
})

ht.fit(test_data)
ht.create_anonymized_columns(num_rows=1000, column_names=['name'])
```
```
TransformerProcessingError: The Faker function you specified is not able to generate 1000 unique values. Please use a different Faker function for column ('name').
```

### Expected behavior
The [RegexGenerator RDT](https://docs.sdv.dev/rdt/transformers-glossary/id/regexgenerator) runs into a very similar problem where uniqueness cannot be enforced if the Regex format string doesn't allow for a large enough # of values. In this case, we do not crash; we _continue to generate new strings_ while warning the user that they will not follow the format. 

We can take a similar approach for the AnonymizedFaker:
```python
ht.create_anonymized_columns(num_rows=1000, column_names=['name'])
```
```
UserWarning: Unable to generate enough unique values for column 'name' in a human-readable format. Additional values may be created randomly.
```

Suggested approach: Similar to `RegexGenerater`, we can repeat the values but with parenthesis after them to indicate that they are copies. For example, `Alice(1)`, `Bob(1)`, etc.

pvk-developer:
Resolves #981 
CU-86b4ktxkr

### DIFF
--- a/rdt/transformers/pii/anonymizer.py
+++ b/rdt/transformers/pii/anonymizer.py
@@ -201,7 +201,7 @@ class AnonymizedFaker(BaseTransformer):
 
     def _fallback_function(self):
         """Return the result of calling a fallback ``faker`` function."""
-        return getattr(self.faker.unique, 'bothify')(text='??????')
+        return self.faker.unique.bothify(text='??????')
 
     def _set_faker_seed(self, data):
         hash_value = self.get_input_column()
--- a/rdt/transformers/pii/anonymizer.py
+++ b/rdt/transformers/pii/anonymizer.py
@@ -199,6 +199,10 @@ class AnonymizedFaker(BaseTransformer):
 
         return result
 
+    def _fallback_function(self):
+        """Return the result of calling a fallback ``faker`` function."""
+        return getattr(self.faker.unique, 'bothify')(text='??????')
+
     def _set_faker_seed(self, data):
         hash_value = self.get_input_column()
         for value in data.head(5):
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b4ktxkr RDT - AnonymizedFaker should have a fallback in the case that it's unable to generate enough unique values #981](https://app.clickup.com/t/86b4ktxkr)

R-Palazzo (approved):
LGTM!

rdt/transformers/pii/anonymizer.py
```diff
@@ -199,6 +199,10 @@ def _function(self):
 
         return result
 
+    def _fallback_function(self):
+        """Return the result of calling a fallback ``faker`` function."""
+        return getattr(self.faker.unique, 'bothify')(text='??????')
```
amontanez24, line 204: why do we need to use getattr here?

amontanez24 (commented):
Looks good, just one question

frances-h (approved):

rdt/transformers/pii/anonymizer.py
pvk-developer: I just copied over what we had on the other method in order to be the same but you are right there is no need to have the getattr if we are not even treating it.

I removed it here: https://github.com/sdv-dev/RDT/pull/986/commits/34d54931ef9c1bfa9fc3a0589882d8e3287a70a5

PR #989, opened 2025-04-30
UnixTimestampEncoder: Fix reverse transform missing timezone information for UTC datetime data

### LINKED ISSUE #987: For datetimes: Reverse transformed data no longer includes my timezone
### Environment Details
* RDT version: 1.16.0
* Python version: 3.11
* Operating System: Linux (Colab notebook)

### Error Description
My original datetime column is in a format that contains timezone information -- usually denoted by `%Z` (indicating a timezone name such as `UTC`) or `%z` (indicating a timezone offset such as `+00:00`). All of the timezones in my data are the same (eg. they are all `UTC`).

In this case, the datetime transformers (eg. `UnixTimestampEncoder`) will be able to forward and reverse transform the data without crashing. However, the reverse transformed data will no longer include the timezone component (`%Z` or `%z`). 

### Steps to reproduce
```python
from rdt.transformers import UnixTimestampEncoder

# create some data that has a datetime format (eg. 'UTC')
data = pd.DataFrame(data={
    'my_datetime_col': ['2025-02-04 08:32:21 UTC', '2025-01-13 08:32:21 UTC']
})

ht = HyperTransformer()
ht.set_config({
    'sdtypes': { 'my_datetime_col': 'datetime' },
    'transformers': { 'my_datetime_col': UnixTimestampEncoder(datetime_format='%Y-%m-%d %H:%M:%S %Z')}
})

reversed
```
Observe that the reverse transformed data includes all other datetime components _except_ for the datetime format:

Image

### Root Cause
The reverse transform happens in 2 steps:

1. We convert the Unix timestamp data (numerical) into a `pd.datetime` representation
2. We then convert the `pd.datetime` representation back into a `string` of the desired format

The problem happens in step (1). We are just using `pd.to_datetime` without specifying any parameter `utc=True`. This step then gets rid of any timezone information.
https://github.com/sdv-dev/RDT/blob/de8884c9d6636eb4ab0b80d62157825b1bac08f8/rdt/transformers/datetime.py#L217

Adding `utc=True` to this step will fix this problem.

**What if the original datetime value did not have a timezone?**:
- If there is a datetime format provided, then we are convert it back into a string anyways and the timezone info will be dropped. So this solution will still work.
- If there is no datetime format provided, then step (2) will be skipped. In this case, we would need to convert it back into the original, timezone-naive dtype.

pvk-developer:
Resolves #987 
CU-86b4tkdmn

This pull request also modularizes the code base for the `UnixTimestampEncoder` in order to help us overwrite methods in future if required.

### Additional changes:
* Fallback to `UTC` if the timezone used is dubious.
    * Warnings are being displayed if the transformer has to fallback to this default.
    * This only happens when your data looks like: `2025-05-01 22:03:40 EST` and the dateutil library can't determine which EST zone you are referring to.

* Added `python-dateutil` library as requirement in order to easily parse the input datetime strings to timezone aware objects.

### DIFF
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -86,7 +86,6 @@ class UnixTimestampEncoder(BaseTransformer):
             try:
                 self._learn_has_multiple_timezones(data)
                 parsed_data = self._to_datetime(data)
-                self._warn_if_mixed_timezones()
                 return parsed_data
 
             except ValueError as error:
@@ -109,7 +108,9 @@ class UnixTimestampEncoder(BaseTransformer):
 
     def _warn_if_mixed_timezones(self):
         if self._has_multiple_timezones:
-            warnings.warn('Mixed timezones are supported only in SDV Enterprise.')
+            warnings.warn(
+                'Mixed timezones are not supported in SDV Community. Data will be converted to UTC.'
+            )
 
     def _raise_appropiate_conversion_error(self, error):
         message = str(error)
@@ -165,6 +166,8 @@ class UnixTimestampEncoder(BaseTransformer):
         transformed = self._transform_helper(data)
 
         self._learn_timezone_offest(data)
+        self._warn_if_mixed_timezones()
+
         if self.enforce_min_max_values:
             self._min_value = transformed.min()
             self._max_value = transformed.max()
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -1,13 +1,17 @@
 """Transformer for datetime data."""
 
+import warnings
+
 import numpy as np
 import pandas as pd
+from dateutil import parser
 from pandas.api.types import is_datetime64_dtype, is_numeric_dtype
 from pandas.core.tools.datetimes import _guess_datetime_format_for_array
 
 from rdt.errors import TransformerInputError
 from rdt.transformers.base import BaseTransformer
 from rdt.transformers.null import NullTransformer
+from rdt.transformers.utils import data_has_multiple_timezones
 
 
 class UnixTimestampEncoder(BaseTransformer):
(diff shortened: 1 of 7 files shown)


rdt/transformers/datetime.py
```diff
@@ -200,6 +200,28 @@ def _transform(self, data):
         data = self._transform_helper(data)
         return self.null_transformer.transform(data)
 
+    def _convert_data_to_pandas_datetime(self, data):
+        use_utc = '%z' in self.datetime_format.lower() if self.datetime_format else False
+        datetime_data = pd.to_datetime(data, utc=use_utc)
```
gsheni, line 205: Similar to `pd.to_numeric` do you need to set `errors='coerce'` for `pd.to_datetime`?

rdt/transformers/utils.py
```diff
@@ -383,3 +385,39 @@ def _handle_enforce_uniqueness_and_cardinality_rule(enforce_uniqueness, cardinal
...
 
     return result
+
+
+def _safe_parse_datetime(value):
+    if isinstance(value, (pd.Timestamp, datetime.datetime)):
+        return value
+    try:
+        return parser.parse(value)
+    except Exception:
```
gsheni, line 395: I don't see any [other blanket Exception catches](https://github.com/search?q=repo%3Asdv-dev%2FRDT+except+Exception&type=code) in RDT. Should we be more specific?

rdt/transformers/utils.py
```diff
@@ -383,3 +385,39 @@ def _handle_enforce_uniqueness_and_cardinality_rule(enforce_uniqueness, cardinal
...
+        return None
+
+
+def _get_utc_offset(dt):
+    try:
+        return dt.utcoffset()
+    except Exception:
```
gsheni, line 402: Same comment as above

rdt/transformers/utils.py
```diff
@@ -383,3 +385,39 @@ def _handle_enforce_uniqueness_and_cardinality_rule(enforce_uniqueness, cardinal
...
+    Returns:
+        bool:
+            True if multiple timezones are detected, False otherwise.
+    """
+    try:
+        parsed_datetimes = data.apply(_safe_parse_datetime).dropna()
+        offsets = parsed_datetimes.apply(_get_utc_offset).dropna()
+        return offsets.nunique() > 1
+
+    except Exception:
```
gsheni, line 422: Same comment as above

rdt/transformers/datetime.py
```diff
@@ -68,43 +72,52 @@ def __init__(
...
-                if self.datetime_format:
-                    pandas_datetime_format = self.datetime_format.replace('%-', '%')
+    def _to_datetime(self, data):
+        datetime_format = self._get_pandas_datetime_format()
+        return pd.to_datetime(data, format=datetime_format, utc=self._has_multiple_timezones)
 
-                data = pd.to_datetime(data, format=pandas_datetime_format)
+    def _get_pandas_datetime_format(self):
+        if self.datetime_format:
+            return self.datetime_format.replace('%-', '%')
```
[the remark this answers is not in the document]
pvk-developer: Thanks I have addressed and added some tests for it!

frances-h (approved):
LGTM, just one question about the error message

gsheni (approved):

rdt/transformers/datetime.py
pvk-developer: Sorry I haven't answered this earlier but there is a reason why not to use `'coerce'` . Mainly we want to make sure that the data provided in your column can be casted to datetime, that way we ensure that you have assigned properly the transformer to your column.

If we used `errors='coerce'` we can have a scenario were you have assigned this transformer to a cetegorical column, let's say this:
```python
df = pd.DataFrame({
    'cat_column': [
        'cat',
        'tiger',
        'monster',
        'randomstring'
    ]
})

pd.to_datetime(df['cat_column'], errors='coerce')
```

-Output:
```bash
0   NaT
1   NaT
2   NaT
3   NaT
Name: cat_column, dtype: datetime64[ns]
```

When you run this code, you end up getting `NaT`'s all around, which would continue working fine in `SDV`. Then, when you sample you go from `cat's` to `NaT's` which is just not the input data (not even close). Since we want to avoid this scenarios, we just want to be able to cast the entire column to `datetime`, if it is not convertible you probably should use another transformer (eg, categorical).

rdt/transformers/datetime.py
```diff
@@ -68,43 +71,56 @@ def __init__(
...
+                'Mixed timezones are not supported in SDV Community. Data will be converted to UTC.'
+            )
 
-        return data
+    def _raise_appropiate_conversion_error(self, error):
+        message = str(error)
+        if 'Unknown string' in message or 'Unknown datetime string' in message:
+            raise TypeError('Data must be of dtype datetime, or castable to datetime.') from None
+
+        raise ValueError('Data does not match specified datetime format.') from None
```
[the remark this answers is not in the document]
pvk-developer: This is a great question for @npatki 

I’ve just refactored the existing codebase using the current behavior as-is — I haven’t added or changed any error messages except those related to timezone awareness.

PR #992, opened 2025-05-07
Make `_safe_parse_datetime` work for for large numbers

### LINKED ISSUE #991: `_safe_parse_datetime` causes `OverflowError` for large numbers
### Error Description

The `_safe_parse_datetime` method calls `dateutil.parser.parse` on the input value. When the value is a large integer, it will try to interpret it as a Unix timestamp but will fail to convert this large number to a C long (which is used internally for timestamp calculations), causing the overflow.

### Steps to Reproduce

Running `_safe_parse_datetime('20220902110443000000')` causes `OverflowError: Python int too large to convert to C long`.

### Additional Context

This error was first observed on [this](https://github.com/sdv-dev/SDV/actions/runs/14865377207/job/41740757122#step:5:497) test case in SDV.

fealho:
Resolves #991.

### DIFF
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -105,6 +105,9 @@ class UnixTimestampEncoder(BaseTransformer):
 
     def _learn_has_multiple_timezones(self, data):
         """Determines if the data contains multiple timezones and stores the result."""
+        if self.datetime_format and '%z' not in self.datetime_format.lower():
+            return
+
         if not isinstance(data, pd.Series):
             data = data.to_series()
 
--- a/rdt/transformers/datetime.py
+++ b/rdt/transformers/datetime.py
@@ -74,43 +74,6 @@ class UnixTimestampEncoder(BaseTransformer):
         self._has_multiple_timezones = False
         self._timezone_offset = None
 
-    def _learn_has_multiple_timezones(self, data):
-        """Determines if the data contains multiple timezones and stores the result."""
-        if not isinstance(data, pd.Series):
-            data = data.to_series()
-
-        self._has_multiple_timezones = data_has_multiple_timezones(
-            data, datetime_format=self.datetime_format
-        )
-
-    def _convert_to_datetime(self, data):
-        if self._needs_datetime_conversion(data):
-            try:
-                self._learn_has_multiple_timezones(data)
-                parsed_data = self._to_datetime(data)
-                return parsed_data
-
-            except ValueError as error:
-                self._raise_appropiate_conversion_error(error)
-
-        return data
-
-    def _needs_datetime_conversion(self, data):
-        """Determines if the data requires datetime conversion."""
-        return self.datetime_format is not None or not is_numeric_dtype(data)
-
-    def _to_datetime(self, data):
-        """Converts data to datetime using the instance's datetime format."""
-        datetime_format = self._get_pandas_datetime_format()
-        return pd.to_datetime(data, format=datetime_format, utc=self._has_multiple_timezones)
-
-    def _get_pandas_datetime_format(self):
-        """Converts the instance's datetime format to a pandas-compatible format."""
-        if self.datetime_format:
-            return self.datetime_format.replace('%#', '%').replace('%-', '%')
-
-        return None
-
     def _warn_if_mixed_timezones(self):
         if self._has_multiple_timezones:
             warnings.warn(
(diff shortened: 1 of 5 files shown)


tests/unit/transformers/test_utils.py
```diff
@@ -550,17 +550,20 @@ def test__safe_parse_datetime():
     """
     # Setup
     str_input = '2023-01-01 12:00:00+0200'
+    str_input_no_format = '20220902110443000000'
```
amontanez24, line 553: what is the case where an input string like this would come up? It doesn't seem to be a valid datetime since I think pandas only accepts up to microseconds and this goes 10^-8

amontanez24 (commented):
Why is this test failing now?

rdt/transformers/utils.py
```diff
@@ -422,6 +425,10 @@ def _safe_parse_datetime(value, warn=False):
         return value
 
     try:
+        # Strings of large numbers cause parser.parse to overflow
+        if isinstance(value, str) and value.isdigit() and datetime_format:
+            return pd.to_datetime(value, format=datetime_format)
```
pvk-developer, line 434:
```suggestion
        if isinstance(value, str) or value.isdigit() and datetime_format:
            value = pd.to_datetime(value, format=datetime_format)
```

We need the `value` below in order to detect the `timezone` for it. (We are either returning the `parser.parse` or `None`)

fealho: parser.parse doesn't support datetime objects as input. If we want to handle timezones for this case, I would have to implement special logic for it.

pvk-developer (changes requested):

rdt/transformers/utils.py
fealho: Can you confirm that for point 1 we want to set `self._has_multiple_timezones = False` when `'%z'` is in the `datetime_format`?

fealho: @pvk-developer A test case using mixed timezone offsets fails if I add this change. The reason is that `_has_multiple_timezones` never gets set to True, since we are skipping that line. Could you clarify why this change is needed?

I implemented all your other recommendations, they are all good additions, this is the only one I'm unsure about.

rdt/transformers/datetime.py
```diff
@@ -74,53 +74,71 @@ def __init__(
...
+            for val in data:
+                if pd.notna(val):
+                    try:
+                        dt = _safe_parse_datetime(
+                            str(val), warn=True, datetime_format=self.datetime_format
+                        )
+                        self._timezone_offset = dt.tzinfo
+                        break
+
+                    except (ValueError, TypeError, AttributeError):
```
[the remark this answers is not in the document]
pvk-developer: I added this to make sure that if date util parser returns a `nonetype` or an unexpected value that has no `tzinfo`.

rdt/transformers/utils.py
```diff
@@ -427,7 +430,15 @@ def _safe_parse_datetime(value, warn=False):
 
     try:
         with warnings.catch_warnings(record=True) as captured_warnings:
-            dt = parser.parse(value)
+            try:
+                dt = parser.parse(value)
+
+            # Strings of large numbers cause parser.parse to overflow
+            except OverflowError as error:
```
pvk-developer, line 437: This will fix the `utc` being dropped. For some reason when there is `UTC` in the large number of values (as string) this will produce `ParserError`.
PS: you need to import it from dateutil:
```python
from dateutil.parser import UnknownTimezoneWarning, ParserError
```

```suggestion
            except (OverflowError, ParserError) as error:
```

tests/integration/transformers/test_datetime.py
```diff
@@ -316,6 +316,68 @@ def test_datetime_strings_with_timezone_name(self):
...
+
+        # Run
+        transformer.fit(data, column='my_datetime_col')
+        transformer.set_random_state(np.random.RandomState(42), 'reverse_transform')
+        transformed = transformer.transform(data)
+        reverse_transformed = transformer.reverse_transform(transformed)
+
+        # Assert
+        expected_data = data = pd.DataFrame({
+            'my_datetime_col': ['20220902110443000001 ', '20220902110443000000 ']
```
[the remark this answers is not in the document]
pvk-developer: With the suggested change on the previous comment this will work:
image

pvk-developer (commented):
Looks great, thank you for going the extra mile to address this! I just left what to me is the final comment (in order to fix the tests) and then I will approve 👍🏻

rdt/transformers/utils.py
fealho: That's a great find!

rdt/transformers/datetime.py
```diff
@@ -74,53 +74,74 @@ def __init__(
         self._has_multiple_timezones = False
         self._timezone_offset = None
 
+    def _warn_if_mixed_timezones(self):
+        if self._has_multiple_timezones:
+            warnings.warn(
+                'Mixed timezones are not supported in SDV Community. Data will be converted to UTC.'
+            )
+
+    def _raise_appropiate_conversion_error(self, error):
```
amontanez24, line 83:
```suggestion
    def _raise_appropriate_conversion_error(self, error):
```

rdt/transformers/utils.py
```diff
@@ -427,7 +430,15 @@ def _safe_parse_datetime(value, warn=False):
...
-            dt = parser.parse(value)
+            try:
+                dt = parser.parse(value)
+
+            # Strings of large numbers cause parser.parse to overflow
+            except (OverflowError, ParserError) as error:
+                if not datetime_format:
+                    raise error
+                value = str(pd.to_datetime(value, format=datetime_format))
+                dt = parser.parse(str(value))
```
amontanez24, line 441: isn't the str here redundant?

amontanez24 (commented):
Just some minor comments

amontanez24 (approved):

PR #1002, opened 2025-06-10
Change default distribution for `GaussianNormalizer` to `truncnorm`

### LINKED ISSUE #997: Change default distribution for `GaussianNormalizer` to `truncnorm`
### Environment Details

* RDT version: 1.17.0
* Python version: 3.12

### Error Description

When using `GaussianNormalizer` with default values a `FutureWarning` is being raised as follows:

```
FutureWarning: Future versions of RDT will not support 'truncated_gaussian' as an option. Please use 'truncnorm' instead.
```

We should change the default distribution to `truncnorm` instead and avoid raising our own warning by default.

### Steps to reproduce

```python
from rdt.transformers import GaussianNormalizer

transformer = GaussianNormalizer()
```

fealho:
CU-86b51x0g5, Resolve #997.

### DIFF
--- a/rdt/transformers/numerical.py
+++ b/rdt/transformers/numerical.py
@@ -276,7 +276,7 @@ class GaussianNormalizer(FloatFormatter):
             Whether or not to clip the data returned by ``reverse_transform`` to the min and
             max values seen during ``fit``. Defaults to ``False``.
         distribution (copulas.univariate.Univariate or str):
-            Copulas univariate distribution to use. Defaults to ``truncated_gaussian``.
+            Copulas univariate distribution to use. Defaults to ``truncnorm``.
             Options include:
 
                 * ``norm``: Use a Gaussian distribution.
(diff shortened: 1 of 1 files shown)


rdt/transformers/numerical.py
```diff
@@ -334,7 +334,7 @@ def __init__(
         model_missing_values=None,
         learn_rounding_scheme=False,
         enforce_min_max_values=False,
-        distribution='truncated_gaussian',
+        distribution='truncnorm',
```
gsheni, line 337: Add unit test check?
https://github.com/sdv-dev/RDT/blob/main/tests/unit/transformers/test_numerical.py#L796

fealho: Done

gsheni (approved):

pvk-developer (approved):

PR #1031, opened 2025-09-03
Add workflow to create and upload package to track downloads

### LINKED ISSUE #1030: Add workflow to create and upload package to track downloads
### Problem Description

We would like to be able to track who is downloading RDT. In order to do this, we will be adding a dependency that is hosted on a server that can track the information from the download request.

### Expected behavior

- When RDT is released, a package called `rdt-download-tracker` should be uploaded to S3
- The contents of the package should be 
```
rdt-download-tracker/
|-- rdt_download_tracker/
  |-- [version.py](http://version.py/)
  |-- __init__.py
|-- [README.md](http://readme.md/)
|-- pyproject.toml
```
- The version should be set to match the version of the RDT release

### Additional context

- Rather than maintaining a separate repo for this tracking package, we can just build the package structure and wheels on the fly in a Github action
- Once this is setup, we can add this package as a dependency

amontanez24:
resolves #1030

This PR adds a script and github workflow that:
- On release, creates a python package called `rdt-download-tracker`
- Matches the version of this package to the version of the rdt release
- uploads the package to S3
- Updates the index.html page in the S3 bucket to have the new files (this is required for pip to be able to install it)

### DIFF
--- a/scripts/rdt-download-tracker/README.md
+++ b/scripts/rdt-download-tracker/README.md
@@ -1,1 +1,2 @@
 # TODO: fill this in
+Update this with whatever we add in [issue 1034](https://github.com/sdv-dev/RDT/issues/1034)


.github/workflows/publish_download_tracker.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  publish-download-tracker:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+        # with:
+        #   ref: 'stable'
```
amontanez24, line 16: I will uncomment this before merging because we want it to install the rdt release version

.github/workflows/publish_download_tracker.yml
```diff
@@ -0,0 +1,36 @@
+name: Publish Download Tracker
+on:
+  release:
+    types: [published]
+  pull_request:
```
amontanez24, line 5: I will remove this before merging

scripts/create_download_tracker.py
```diff
@@ -0,0 +1,108 @@
...
+    if dryrun:
+        print('New index file:')  # noqa: T201 `print` found
+        print(new_index)  # noqa: T201 `print` found
+    else:
+        s3_client.put_object(
+            Bucket=BUCKET,
+            Key=index_file_path,
+            Body=new_index,
+            ContentType='text/html',
+            CacheControl='no-cache',
```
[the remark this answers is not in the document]
amontanez24: does the index file need a checksum? Wouldn't it just be needed for the actual wheels and tar.gz files?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,87 @@
+name: Publish
```
gsheni, line 1:
```suggestion
name: Publish RDT
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,87 @@
...
+        description: 'Release candidate.'
+        required: true
+        type: boolean
+        default: true
+      test_pypi:
+        description: 'Test PyPI.'
+        type: boolean
+        default: false
+jobs:
+  release:
```
gsheni, line 27:
```suggestion
  publish-rdt::
```

PR #1090, opened 2026-07-09
Allow categorical transformers to ignore missing values for forward and reverse

### LINKED ISSUE #1089: Allow categorical transformers to ignore missing values for forward and reverse
### Problem Description
For some types of usage, it is desirable for the transformers to ignore missing values for both the forward and reverse transform. For exmaple, maybe the downstream model or analysis actually is OK with missing values so we want to keep them.

Currently, all transformers have settings for missing value handling _except_ for the categorical ones.

### Expected behavior
The following applies to: UniformEncoder, OrderedUniformEncoder, LabelEncoder, OrderedLabelEncoder, ~BinaryEncoder~ _(Edit: BinaryEncoder does not need this behavior because it already has methods of ignoring missing values.)_

These transformers should have a parameter called:
- `missing_value_encoding`: This parameter controls how missing values are encoded when transforming and reversing the data.
    - (default) `"new_category"`: Missing values are encoded as a separate category. When reversing, the transformer reverses the encoding to regenerate the missing values. (Note that this is the status quo for all of these transformers.)
    - `None`: Missing values are not encoded. In this case, missing values are kept missing. When reversing, the transformer passes any missing values through as-is.

Note that for `OrderedUniformEncoder` and `OrderedLableEncoder`: It is ok if the order parameter does not include missing values if `missing_value_encoding=None`.

fealho:
CU-86bang6q2, Resolve #1089

### DIFF
--- a/rdt/transformers/base.py
+++ b/rdt/transformers/base.py
@@ -113,7 +113,7 @@ class BaseTransformer:
 
     @property
     def model_missing_values(self):
-        """Return whether or not a new column is being used to model missing values."""
+        """Whether or not a new column is being used to model missing values."""
         warnings.warn(
             "Future versions of RDT will not support the 'model_missing_values' parameter. "
             "Please switch to using the 'missing_value_generation' parameter instead.",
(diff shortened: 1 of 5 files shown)


sdv-team:
Task linked: [CU-86bang6q2 RDT - Allow categorical transformers to ignore missing values for forward and reverse #1089](https://app.clickup.com/t/86bang6q2)

frances-h (approved):

amontanez24 (approved):

PR #1096, opened 2026-07-29
Make `RegexGenerator` generate unique values on consecutive `reverse_transform` calls 

### LINKED ISSUE #1091: `RegexGenerator` not generating unique values on consecutive `reverse_transform` calls
### Environment Details

* RDT version:
* Python version:
* Operating System:

### Error Description
For `RegexGenerator`, consecutive `reverse_transform` calls can return already generated values even if `cardinality_rule` is set to `unique`. As long as randomization has not been reset, `reverse_transform` should always generate a new unique value, using appropriate fallbacks if the original generator has been exhausted. 

### Steps to reproduce
```python
transformer = RegexGenerator(regex_format=r'\d', cardinality_rule='unique')
transformer.fit(pd.DataFrame(columns=['id']), 'id')
samples = pd.concat([transformer.reverse_transform(pd.DataFrame(index=[0])) for i in range(50)])['id']
assert samples.is_unique
```

fealho:
CU-86bav75gd, Resolve #1091

### DIFF
--- a/rdt/transformers/id.py
+++ b/rdt/transformers/id.py
@@ -136,6 +136,7 @@ class RegexGenerator(BaseTransformer):
         """Set the generator when pickling."""
         generator_size = state.get('generator_size')
         generated = state.get('generated')
+        last_generated_value = state.get('_last_generated_value')
         generator, size = strings_from_regex(state.get('regex_format'))
         if generator_size is None:
             state['generator_size'] = size
@@ -144,7 +145,13 @@ class RegexGenerator(BaseTransformer):
 
         if generated:
             for _ in range(generated):
-                next(generator)
+                regex_value = next(generator)
+
+            if last_generated_value is None:
+                state['_last_generated_value'] = regex_value
+
+        elif '_last_generated_value' not in state:
+            state['_last_generated_value'] = None
 
         state['generator'] = generator
         self.__dict__ = state
@@ -180,12 +187,14 @@ class RegexGenerator(BaseTransformer):
         # Used otherwise
         self.generator_size = None
         self.generated = None
+        self._last_generated_value = None
 
     def reset_randomization(self):
         """Create a new generator and reset the generated values counter."""
         super().reset_randomization()
         self.generator, self.generator_size = strings_from_regex(self.regex_format)
         self.generated = 0
+        self._last_generated_value = None
 
         if hasattr(self, 'cardinality_rule') and self.cardinality_rule == 'scale':
             self._remaining_samples['repetitions'] = 0
@@ -297,6 +306,9 @@ class RegexGenerator(BaseTransformer):
         except (RuntimeError, StopIteration):
             pass
 
+        if samples:
+            self._last_generated_value = samples[-1]
+
         return samples
 
     def _sample_from_template(self, num_samples, template_samples):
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86bav75gd RDT - `RegexGenerator` not generating unique values on consecutive `reverse_transform` calls #1091](https://app.clickup.com/t/86bav75gd)

pvk-developer (changes requested):
I think that this introduced a bug with categorical values:
```python

transformer = RegexGenerator(regex_format='[A-Da-d]', cardinality_rule='unique')
transformer.fit(pd.DataFrame(columns=['id']), 'id')
samples = pd.concat([transformer.reverse_transform(pd.DataFrame(index=[0])) for i in range(50)])['id']

samples
# output 
0                                                    A
0                                                    B
0                                                    C
0                                                    D
0                                                    a
0                                                    b
0                                                    c
0                                                    d
0                                                 d(0)
0                                              d(0)(0)
0                                           d(0)(0)(0)
........
```

rdt/transformers/id.py
```diff
@@ -180,12 +212,14 @@ def __init__(
         # Used otherwise
         self.generator_size = None
         self.generated = None
+        self._fallback_counter = 0
```
frances-h: Is this just the number of samples that have been generated? Can we change the variable name to indicate that?

rdt/transformers/id.py
```diff
@@ -389,16 +423,28 @@ def _sample(self, num_samples, unique_condition):
...
             self.reset_randomization()
 
         samples = self._sample_from_generator(num_samples)
+
+        # Need more samples than the generator can produce
         if num_samples > len(samples):
             if unique_condition:
-                new_samples = self._sample_fallback(num_samples - len(samples), samples)
+                fallback_size = num_samples - len(samples)
+                self.generator = self._create_fallback_generator()
```
frances-h: Why do we need to reset the generator here? Shouldn't it already be set to the fallback generator?

frances-h (commented):
Looks good, just a couple of questions.

Also, after the generator has been exhausted, every `reverse_transform` call I'm seeing: ```UserWarning: The regex for 'id' can only generate 100 unique values. Additional values may not exactly follow the provided regex.```. Can we update that so it only displays once (the first time the generator has to fall back)?

rdt/transformers/id.py
fealho: it's the number of fallback values that have been generated. I'll rename it.

rdt/transformers/id.py
fealho: This runs when the original generator runs out of regexes halfway through, and we still need more. For this case we haven't set `self.generator` yet.

frances-h: How is `self.generator` not set yet if we called `_sample_from_generator` above? Shouldn't we just set the generator to the fallback generator in the unique case? If the fallback generator is an infinite generator (which I think it is?), would we ever even hit this in the unique case? Shouldn't we always be able to generate enough values?

pvk-developer (approved):
LGTM!

frances-h (approved):
It looks like some other cases also use `_sample_fallback` to create unique values when the regex is too small/runs out of values. Would it make sense to replace the generator with the fallback generator entirely (adding a yield in the first loop so it first returns the expected regex values first)?

If the generator itself handles running out of values, I imagine we could simplify some of the pathways below. If it'll require a significantly large refactor though, then maybe it should be its own issue.

fealho:
> It looks like some other cases also use `_sample_fallback` to create unique values when the regex is too small/runs out of values. Would it make sense to replace the generator with the fallback generator entirely (adding a yield in the first loop so it first returns the expected regex values first)?
> 
> If the generator itself handles running out of values, I imagine we could simplify some of the pathways below. If it'll require a significantly large refactor though, then maybe it should be its own issue.

This could be a good abstraction, but it would require some refactoring because because non-unique modes like `scale` and `None` would need special handling, it would need logic to pickle correctly, and Enterprise would need to be updated as well.

Issue #1102, opened 2026-08-17
RDT 2.0: Remove all deprecated functionality


npatki:
### Problem Description
Throughout the lifetime of RDT 1.X, we added new parameters and classes that replaced old ones. For backwards compatibility, we continued to support the deprecated classes and parameters, raising a FutureWarning to let users know of upcoming removal.

For RDT 2.0, all of that functionality can be officially dropped.

### Expected behavior
- Remove all deprecated classes, modules, functions and parameters
- Remove all FutureWarnings

### Additional Context
Merge into a feature branch for RDT 2.0, as there will be other issues that are also needed for this change.

PR #1104, opened 2026-08-20
RDT 2.0: Remove all deprecated functionality

R-Palazzo:
Resolve #1102
86bbfm8g1

- Added this test [file](https://github.com/sdv-dev/RDT/blob/issue-1102-remove-deprecated-functionality/tests/unit/test__rdt_2_0_deprecations.py) to keep track of everything that is removed.
- Did one commit per functionality removed


rdt/transformers/pii/anonymizer.py
```diff
@@ -209,7 +198,9 @@ def _function(self):
                 faker_attr = self.faker
 
         except AttributeError:
-            faker_attr = self.faker.unique if self.enforce_uniqueness else self.faker
+            faker_attr = (
+                self.faker.unique if getattr(self, 'enforce_uniqueness', False) else self.faker
```
amontanez24: I don't think we have to maintain the backwards compatibility

rdt/transformers/utils.py
```diff
@@ -367,57 +365,6 @@ def sigmoid(data, low, high):
...
-        if sdtype in DEPRECATED_SDTYPES_MAPPING and not self._warned.get(sdtype):
-            new_sdtype = DEPRECATED_SDTYPES_MAPPING.get(sdtype)
-            warnings.warn(
-                f"The sdtype '{sdtype}' is deprecated and will be phased out. "
-                f"Please use '{new_sdtype}' instead.",
-                DeprecationWarning,
-            )
-            self._warned[sdtype] = True
-
-        return super().get(sdtype)
```
amontanez24, line 392: this might still be useful in the future no?

rdt/transformers/pii/anonymizer.py
R-Palazzo: Ok I wanted to ask about this. I wasn't sure because there is this comment in an integration test: 
```
Checks that transformers without the ``cardinality_rule`` attribute still function as
        expected (can happen when previous transformer version is loaded from a pkl file).
```

https://github.com/sdv-dev/RDT/blob/c50ebd6dd11981956b14dd3522783686c17633ff/tests/integration/transformers/pii/test_anonymizer.py#L248

With RDT 2.0 should we stop maintaining this backwards compatibility?
And in general what we kept for backwards compatibility so we can still sample data

amontanez24: @frances-h will we need this for 2.0?

amontanez24 (approved):

frances-h (approved):

